PR #35, opened 2025-04-07
feat(tracing): Use `rich` for console exporter.

daavoo:
Didn't iterate much on the style but it should be easy to extend/modify:

| Before | After |
|--------|--------|
| Screenshot 2025-04-07 115958 | Screenshot 2025-04-07 120134 |


njbrake:
Looks great. As you see, there's more work to do to better extract nested json/lists/etc (left over from my initial console exporting PR), but the coloring and conversion to RICH is a nice improvement!

PR #38, opened 2025-04-07
bugfix for span extraction

njbrake:
extract_interaction returns a tuple now, update typing as well as the telemetry extraction

### DIFF
--- a/src/any_agent/telemetry/openai_telemetry.py
+++ b/src/any_agent/telemetry/openai_telemetry.py
@@ -83,16 +83,48 @@ class OpenAITelemetryProcessor(TelemetryProcessor):
 
         return span_info
 
+    def _extract_chain_interaction(self, span: Dict[str, Any]) -> Dict[str, Any]:
+        """Extract information from a CHAIN span."""
+        attributes = span.get("attributes", {})
+
+        span_info = {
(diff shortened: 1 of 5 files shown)


src/any_agent/telemetry/openai_telemetry.py
```diff
@@ -83,16 +83,48 @@ def _extract_agent_interaction(self, span: Dict[str, Any]) -> Dict[str, Any]:
...
+        try:
+            span_info["output"] = json.loads(output_value)
+        except (json.JSONDecodeError, TypeError):
+            span_info["output"] = output_value
+
+        # Add service name if available
+        if "service.name" in span.get("resource", {}).get("attributes", {}):
+            span_info["service"] = span["resource"]["attributes"]["service.name"]
+
+        return span_info
```
njbrake, line 116: OpenAI was missing chain extraction so adding it here

daavoo (commented):
👍 I guess we will be catching this in the tests once we put the eval?

njbrake:
> 👍 I guess we will be catching this in the tests once we put the eval?

I think we would also have caught it if we had something like mypy doing more advance pre-compile checking 🤷

PR #46, opened 2025-04-08
More direct calls for contribution

njbrake:
I will wait for approval before merging this one, I could go either way but it seemed like it might be nice to directly say "we know we have more frameworks to support and we welcome help making it happen"


daavoo:
Define `soon` 🤔 

😆

njbrake:
> Define `soon` 🤔
> 
> 😆

Yeah you're right lol. Updated the phrasing to not promise that it's coming soon, but to indicate that we know about them and would love for any-agent to support.

PR #67, opened 2025-04-11
Tests are sync

njbrake:
I didn't realize that async tests get skipped unless you put a plugin into place. This re-arranges the load_agent code to be a part of the create function instead of being dynamically loaded at agent run time

### DIFF
--- a/src/any_agent/tools/mcp.py
+++ b/src/any_agent/tools/mcp.py
@@ -1,11 +1,11 @@
 """Tools for managing MCP (Model Context Protocol) connections and resources."""
 
-import os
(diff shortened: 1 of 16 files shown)


src/any_agent/tools/mcp.py
```diff
@@ -1,11 +1,11 @@
 """Tools for managing MCP (Model Context Protocol) connections and resources."""
 
+from abc import ABC, abstractmethod
 import os
+from loguru import logger
```
daavoo: loguru was dropped in #66

PR #75, opened 2025-04-14
74 allow to configure tracing console output

daavoo:
Decided to disable `CHAIN` and `AGENT` by default. They were too noisy in some frameworks (`langchain`, `llama_index`)

### DIFF
--- a/src/any_agent/telemetry/langchain_telemetry.py
+++ b/src/any_agent/telemetry/langchain_telemetry.py
@@ -221,7 +221,7 @@ class LangchainTelemetryProcessor(TelemetryProcessor):
         attributes = span.get("attributes", {})
         span_kind = attributes.get("openinference.span.kind", "")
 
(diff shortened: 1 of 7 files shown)


src/any_agent/telemetry/langchain_telemetry.py
```diff
@@ -221,7 +221,7 @@ def extract_interaction(self, span: Dict[str, Any]) -> tuple[str, dict[str, Any]
         attributes = span.get("attributes", {})
         span_kind = attributes.get("openinference.span.kind", "")
 
-        if span_kind == "LLM" and "llm.output_messages.0.message.content" in attributes:
+        if span_kind == "LLM":
```
daavoo, line 224: @njbrake I think this was not the expected so I updated if (it was causing the initial `LLM` span to be considered unkown.

I added some minimal tests for the `TracingConfig` but we should add more tests for the `telemetry` module

PR #76, opened 2025-04-14
70 support managed agents with langchain

daavoo:


src/any_agent/frameworks/langchain.py
```diff
@@ -65,28 +66,69 @@ async def _load_agent(self) -> None:
...
-            prompt=self.config.instructions,
-            **self.config.agent_args or {},
-        )
-        # Langgraph doesn't let you easily access what tools are loaded from the CompiledGraph, so we'll store a list of them in this class
-        self._tools = imported_tools
+            main_agent = create_react_agent(
+                name=self.config.name,
+                model=self._get_model(self.config),
+                tools=imported_tools,
+                prompt=get_instructions(self.config.instructions),
```
njbrake, line 110: The implementation in google AnyAgent is 
`get_instructions(managed_agent.instructions) or "",`

not sure if one style should be used everywhere

njbrake (approved):
LGTM

src/any_agent/frameworks/langchain.py
daavoo: nah, is just a quirk in the google API where they use an empty string as default argument but don't support `None` 🤷

PR #77, opened 2025-04-14
Support for Agno

njbrake:


src/any_agent/frameworks/agno.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def _get_model(self, agent_config: AgentConfig):
+        """Get the model configuration for an Agno agent."""
+        return LiteLLM(
+            id=agent_config.model_id,
+            **agent_config.model_args or {},
+        )
+
+    async def _load_agent(self) -> None:
+        if not self.config.tools:
```
daavoo, line 42: I fell into this trap in #70 .

If you don't put that then the main agent ignores the managed_agents and tries to solve it with the default tools.

```suggestion
        if not managed_agents and not self.config.tools:
```

daavoo (commented):
LGTM. 

Could you also add some simple test(s) in `tests/unit/frameworks`?

src/any_agent/frameworks/agno.py
njbrake: Currently that isn't necessary because of https://github.com/mozilla-ai/any-agent/pull/77/files#diff-8e36bbb2ae7abf556304d4ba4434155ef1ee52df9fe76e0ae9b25b5d96c9cbf8R26-R29 I think?

Basically, if you try to set managed_agents, I throw an exception in the constructor because Agno doesn't yet support multi_agent.

However, doesn't hurt to add it to make it easier for multi agent when that comes, so I'm fine with adding it in

daavoo: Correct, it is not needed, I just commented because it doesn't hurt and prevents making the same mistake I made

PR #85, opened 2025-04-15
tests: Add autouse fixture `disable_rich_console`.

daavoo:
https://github.com/mozilla-ai/any-agent/actions/runs/14469560059/job/40579792875?pr=85

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 import sys
 
 import pytest
+import rich.console
 
 
 @pytest.fixture(scope="function")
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -19,3 +20,14 @@ def refresh_tools():
...
+
+@pytest.fixture(autouse=True)
+def disable_rich_console(monkeypatch):
+    original_init = rich.console.Console.__init__
+
+    def quiet_init(self, *args, **kwargs):
+        kwargs["quiet"] = True
+        original_init(self, *args, **kwargs)
+
+    monkeypatch.setattr(rich.console.Console, "__init__", quiet_init)
```
njbrake, line 33: If a test fails, do we still get the logs, or does this completly silence the logs? If it completely silences them, maybe we should have some sort of --verbose type option so that we can turn the logs back on if we want them for debugging something

njbrake (approved):

tests/conftest.py
daavoo: good point. What do you think if we don't set `quiet=True` if you pass `-s` to pytest (which is the usual argument you would pass to see the logs)?

njbrake: Sounds good to me 👍

njbrake: Usually when I'm debugging tests I run 

```
python -m debugpy --listen 5678 --wait-for-client -m pytest
```

so I generally don't need the logs, but sometimes it comes in handy if the issue is simple to fix and easy to find by a quick skim of a log

PR #88, opened 2025-04-15
Update tracing.md

njbrake:
I think the doc example was broken?

### DIFF
--- a/docs/tracing.md
+++ b/docs/tracing.md
@@ -17,7 +17,7 @@ setup_tracing(framework)
 
 agent = AnyAgent.create(
         framework,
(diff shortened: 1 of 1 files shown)


docs/tracing.md
```diff
@@ -16,11 +16,12 @@ framework = AgentFramework("openai")
 setup_tracing(framework)
 
 agent = AnyAgent.create(
+        framework,
         main_agent=AgentConfig(
```
daavoo, line 20: This is also wrong:

```suggestion
        ¡AgentConfig(
```

Now i want to find a way to run this codeblocks in CI 😅

PR #94, opened 2025-04-15
Simplify and document options for adding sys prompt

njbrake:
The thought here is that it's easy enough to import and extend a string outside of the library, so removing the option for any_agent to handle the importing of a variable makes the library a little more straightforward.

I also added some words to try to explain that libraries are interchanging sys_prompt vs instructions.


daavoo (approved):
Nothing like the look of a PR deleting code 💘

daavoo:
> The thought here is that it's easy enough to import and extend a string outside of the library

unless... someone decides to put the strings in a yaml that gets dynamically imported in the code in a variable that can't be accessed 👼 (not that this was possible with the deleted code)

PR #95, opened 2025-04-16
feat(lint): add strict mypy checks

ELC:

PR #96, opened 2025-04-16
Add exhaustive checks for enums

ELC:
The idea of this PR is to help identify which places will need to get modified whenever a new backend is added. 

For example, by adding a new member to the enum:

```diff
 class AgentFramework(str, Enum):
     GOOGLE = "google"
     LANGCHAIN = "langchain"
     LLAMAINDEX = "llama_index"
     OPENAI = "openai"
     AGNO = "agno"
     SMOLAGENTS = "smolagents"
+    TEST = "TEST"
```

When mypy is run with the `strict` configuration, we can get this type of warnings:
image

Alternatively using Pylance:
image

This is just a peek on what is possible from this type of exhaustive checks, some other parts of the codebase could be also refactor to leverage this mechanism more effectively.

### DIFF
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -1,7 +1,10 @@
+from __future__ import annotations
 from typing import Any, Optional, List
 from abc import ABC, abstractmethod
 import asyncio
 from any_agent.config import AgentFramework, AgentConfig
+
+from typing_extensions import assert_never
 
 
 class AnyAgent(ABC):
@@ -10,7 +13,40 @@ class AnyAgent(ABC):
     This provides a unified interface for different agent frameworks.
     """
 
-    # factory method
+    @staticmethod
+    def _get_agent_type_by_framework(framework: AgentFramework) -> type[AnyAgent]:
+        if framework is AgentFramework.SMOLAGENTS:
+            from any_agent.frameworks.smolagents import SmolagentsAgent
+
+            return SmolagentsAgent
+
+        if framework is AgentFramework.LANGCHAIN:
+            from any_agent.frameworks.langchain import LangchainAgent
+
+            return LangchainAgent
+
+        if framework is AgentFramework.OPENAI:
+            from any_agent.frameworks.openai import OpenAIAgent
+
+            return OpenAIAgent
+
+        if framework is AgentFramework.LLAMAINDEX:
+            from any_agent.frameworks.llama_index import LlamaIndexAgent
+
+            return LlamaIndexAgent
+
+        if framework is AgentFramework.GOOGLE:
+            from any_agent.frameworks.google import GoogleAgent
+
+            return GoogleAgent
+
+        if framework is AgentFramework.AGNO:
+            from any_agent.frameworks.agno import AgnoAgent
+
+            return AgnoAgent
+
+        assert_never(framework)
+
     @classmethod
     def create(
         cls,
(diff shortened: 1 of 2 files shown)


src/any_agent/frameworks/any_agent.py
```diff
@@ -1,37 +1,60 @@
...
+            from any_agent.frameworks.google import GoogleAgent
+
+            return GoogleAgent
+
+        if framework is AgentFramework.AGNO:
+            from any_agent.frameworks.agno import AgnoAgent
+
+            return AgnoAgent
+
+        assert_never(framework)
```
daavoo, line 48: To double-check my understanding (I have never used this before), from the high-level user perspective this is equivalent to before but:

- Raises an AssertionError instead of ValueError (old debate on conventions but no strong opinion on this)
- "Helps" mypy catch missing pieces like the screenshot you shared

Aside from that, I like the method extraction anyhow

src/any_agent/tracing.py
```diff
@@ -123,7 +127,7 @@ def setup_tracing(
     agent_framework: AgentFramework,
     output_dir: str = "traces",
     tracing_config: TracingConfig | None = None,
-) -> str:
+) -> str | NoneType:
```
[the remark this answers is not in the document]
ELC: @daavoo I haven't done any typing checks on this, we can revisit after merging #95

src/any_agent/frameworks/any_agent.py
ELC: The idea is that it should never raise AssertionError, when adding mypy to pre-commit (part of #95) adding a new backend will become more straightforward, just add the enum member and follow the linting errors. This makes "oh, forgot the enum here" not only harder, it makes them impossible to happen

That way we can ensure that when adding a backend, all the necessary constructs are present for that specific backend or else explicitly excluding them by raising a NotImplementedError or similar (for instance, telemetry for Agno)

ELC:
@daavoo this is a good read to understand the benefit:
https://adamj.eu/tech/2022/10/14/python-type-hints-exhuastiveness-checking/

src/any_agent/frameworks/any_agent.py
daavoo: Right, I didn't mean while developing but rather at runtime.

Here is current main behavior:

```py
>>> from any_agent import AnyAgent, AgentConfig
>>> AnyAgent.create("llama_index", AgentConfig(model_id="gpt-4o"))
<any_agent.frameworks.llama_index.LlamaIndexAgent object at 0x7fcd6ff93350>
>>> AnyAgent.create("foo", AgentConfig(model_id="gpt-4o"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/daavoo/any-agent/src/any_agent/frameworks/any_agent.py", line 48, in create
    raise ValueError(msg)
ValueError: Unsupported agent framework: foo
```

And here is this PR:

```py
>>> from any_agent import AnyAgent, AgentConfig
>>> AnyAgent.create("llama_index", AgentConfig(model_id="gpt-4o"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/daavoo/any-agent/src/any_agent/frameworks/any_agent.py", line 57, in create
    agent_cls = cls._get_agent_type_by_framework(agent_framework)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/daavoo/any-agent/src/any_agent/frameworks/any_agent.py", line 48, in _get_agent_type_by_framework
    assert_never(framework)
  File "/usr/lib/python3.11/typing.py", line 2464, in assert_never
    raise AssertionError(f"Expected code to be unreachable, but got: {value}")
AssertionError: Expected code to be unreachable, but got: 'llama_index'
>>> AnyAgent.create("foo", AgentConfig(model_id="gpt-4o"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/daavoo/any-agent/src/any_agent/frameworks/any_agent.py", line 57, in create
    agent_cls = cls._get_agent_type_by_framework(agent_framework)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/daavoo/any-agent/src/any_agent/frameworks/any_agent.py", line 48, in _get_agent_type_by_framework
    assert_never(framework)
  File "/usr/lib/python3.11/typing.py", line 2464, in assert_never
    raise AssertionError(f"Expected code to be unreachable, but got: {value}")
AssertionError: Expected code to be unreachable, but got: 'foo'
```

So, as you can see there is a user-facing change in behavior.

btw @njbrake I am not sure if the first behavior in `main` is intended or not (works by passing string instead of AgentFramework )?

njbrake: @daavoo  yes I think I'd like to let the user create an agent by either passing in an AgentFramework enum or passing in a string. It seems like a nice flexibility for a user so that they don't have to import AgentFramework types into their code if they would rather specify via a string.

ELC: Got it, let me do the necessary change so that user facing behaviour stays the same

ELC:
@daavoo All conflicts have been addressed and I added unit tests for the ValueError you mentioned above, let me know if there is anything you would like to change

njbrake:
> @daavoo All conflicts have been addressed and I added unit tests for the ValueError you mentioned above, let me know if there is anything you would like to change

@ELC thanks! Looks like there may still be an issue with these changes, since the tests are failing.

ELC:
@njbrake I adapted all the cases in the docs to use strings directly and modified the tracing and telemetry public entrypoints to accept both strings and AgentFramework enums.

To keep things separated I worked solely with the Enum's name rather than the value and use `auto` to remove the hardcoded strings.

I also checked we are not using strings anywhere internally and that is just a convenience for the final user. I believe doing something like `AgentFramework(string)` may be confusing so I think we should allow for string values directly or enum values directly, the conversion from string to enum happens transparently for the final user

CC @daavoo

njbrake:
Thanks for all the work on this @ELC, this makes the code cleaner and I am really liking the use of MyPy.

PR #98, opened 2025-04-16
Only pass in a callable or MCPTool

njbrake:


tests/integration/test_import_and_wrap_tools.py
daavoo, line 1: minor: better rename the file to remove import

src/any_agent/tools/wrappers.py
```diff
@@ -36,7 +36,12 @@ def wrap_tool_smolagents(tool):
     from smolagents import tool as smolagents_tool
 
     if not isinstance(tool, Tool):
-        return smolagents_tool(tool)
+        # this wrapping needed until https://github.com/huggingface/smolagents/pull/1203 is merged and released
```
daavoo, line 39: nice. I think we can drop the fixture `refresh_tools`

src/any_agent/tools/wrappers.py
```diff
@@ -100,8 +105,33 @@ async def wrap_mcp_server(
 }
 
 
-async def import_and_wrap_tools(
-    tools: list[str | dict], agent_framework: AgentFramework
+def verify_callable(tool: Callable) -> None:
```
daavoo, line 108: do we need to do this ourselves? Doesn't each framework already do this check (and might have different constraints)?

njbrake: Smolagents does this check immediately, but other tools like llama index didn't throw an error right away if they were missing. So, I went with adding this ourselves so that we can give immediate feedback about the issue. I didn't make it exhaustive (e.g. I don't check that the docstring has correct content like a description of each arg, I don't check that the docstring is in the google docstring format, etc).

PR #101, opened 2025-04-16
refactor(telemetry): Move shared methods to base class.

daavoo:
Add tests/unit/test_unit_telemetry.

### DIFF
--- a/src/any_agent/evaluation/evaluate.py
+++ b/src/any_agent/evaluation/evaluate.py
@@ -10,7 +10,7 @@ from any_agent.evaluation.evaluators import (
 from any_agent.evaluation.results_saver import save_evaluation_results
 from any_agent.evaluation.test_case import TestCase
 from any_agent.logging import logger
-from any_agent.telemetry.telemetry import TelemetryProcessor
+from any_agent.telemetry.base import TelemetryProcessor
 
 
 def evaluate_telemetry(test_case: TestCase, telemetry_path: str) -> bool:
--- a/src/any_agent/telemetry/base.py
+++ b/src/any_agent/telemetry/base.py
@@ -42,6 +42,30 @@ class TelemetryProcessor(ABC):
         msg = f"Unsupported agent type {agent_framework}"
         raise ValueError(msg)
 
+    @abstractmethod
+    def extract_hypothesis_answer(self, trace: list[dict[str, Any]]) -> str:
+        """Extract the hypothesis agent final answer from the trace."""
+
+    @abstractmethod
+    def _get_agent_framework(self) -> AgentFramework:
+        """Get the agent type associated with this processor."""
+
+    @abstractmethod
+    def _extract_llm_interaction(self, span: dict[str, Any]) -> dict[str, Any]:
+        """Extract interaction details of a span of type LLM"""
+
+    @abstractmethod
+    def _extract_tool_interaction(self, span: dict[str, Any]) -> dict[str, Any]:
+        """Extract interaction details of a span of type TOOL"""
+
+    @abstractmethod
+    def _extract_chain_interaction(self, span: dict[str, Any]) -> dict[str, Any]:
+        """Extract interaction details of a span of type CHAIN"""
+
+    @abstractmethod
+    def _extract_agent_interaction(self, span: dict[str, Any]) -> dict[str, Any]:
+        """Extract interaction details of a span of type AGENT"""
+
     @staticmethod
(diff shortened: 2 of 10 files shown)


tests/unit/test_unit_telemetry.py
```diff
@@ -0,0 +1,18 @@
+import json
+
+import pytest
+
+from any_agent import AgentFramework
+from any_agent.telemetry import TelemetryProcessor
+
+
+@pytest.mark.parametrize("framework", list(AgentFramework))
+def test_telemetry_extract_interaction(framework, llm_span):
```
daavoo, line 10: TODO: Add more fixture spans

src/any_agent/evaluation/evaluate.py
```diff
@@ -10,7 +10,7 @@
 from any_agent.evaluation.results_saver import save_evaluation_results
 from any_agent.evaluation.test_case import TestCase
 from any_agent.logging import logger
-from any_agent.telemetry.telemetry import TelemetryProcessor
+from any_agent.telemetry.base import TelemetryProcessor
```
njbrake, line 13: How do you feel about editing the __init__ so that it can become 

```suggestion
from any_agent.telemetry import TelemetryProcessor
```

src/any_agent/telemetry/base.py
```diff
@@ -130,3 +138,28 @@ def parse_generic_key_value_string(text: str) -> dict[str, str]:
...
+            calls.append(self.extract_interaction(span)[1])
+
+        return calls
+
+    def extract_interaction(self, span: dict[str, Any]) -> tuple[str, dict[str, Any]]:
+        """Extract interaction details from a span."""
+        attributes = span.get("attributes", {})
+        span_kind = attributes.get("openinference.span.kind", "")
+
+        if span_kind == "LLM" or "LiteLLMModel.__call__" in span.get("name", ""):
```
ELC, line 156: Could we use an Enum for the `span_kind`?

src/any_agent/telemetry/base.py
```diff
@@ -42,6 +42,30 @@ def create(cls, agent_framework: AgentFramework) -> "TelemetryProcessor":
...
+    def _extract_tool_interaction(self, span: dict[str, Any]) -> dict[str, Any]:
+        """Extract interaction details of a span of type TOOL"""
+
+    @abstractmethod
+    def _extract_chain_interaction(self, span: dict[str, Any]) -> dict[str, Any]:
+        """Extract interaction details of a span of type CHAIN"""
+
+    @abstractmethod
+    def _extract_agent_interaction(self, span: dict[str, Any]) -> dict[str, Any]:
+        """Extract interaction details of a span of type AGENT"""
```
ELC, line 67: It is not common to add abstract method with leading `_`, should we make them public instead?

Since they are going to be mandatory for all concrete classes, I believe it would be a good idea to type the return as a specific `TypedDict` instead of a plain `dict[str, Any]`, this will help consumer understand which fields are present in the different dicts regardless of the framework

tests/unit/test_unit_telemetry.py
```diff
@@ -0,0 +1,18 @@
+import json
+
+import pytest
+
+from any_agent import AgentFramework
+from any_agent.telemetry import TelemetryProcessor
+
+
+@pytest.mark.parametrize("framework", list(AgentFramework))
```
ELC, line 9: There is a fixture already for the `AgentFramework` called `agent_framework: AgentFramework`. I suggest using that instead of the parametrize in this test

src/any_agent/telemetry/base.py
daavoo: > Could we use an Enum for the span_kind?

Yes, I think it is a good idea. We will be revisiting the telemtetry code soon so will probably add something like that

src/any_agent/telemetry/base.py
daavoo: > It is not common to add abstract method with leading _, should we make them public instead?

I don't know if it is common but I think is the right convention here. The methods need to be abstract (any subclass need to implement them) and are not expected to be used outside the class (`_` is a common naming convention for that)

> Since they are going to be mandatory for all concrete classes, I believe it would be a good idea to type the return as a specific TypedDict instead of a plain dict[str, Any], this will help consumer understand which fields are present in the different dicts regardless of the framework

I agree with you, however I feel like we still need to iterate a little bit on the telemetry code (and there are also pending bugs on openinference) before we reach the point where we can define a proper expected type

PR #102, opened 2025-04-16
OpenAI and Smolagents SSE support

njbrake:
I also cleaned up the MCP implementations by using ExitStacks so that the aexit's should get called when the stack exits.


ELC:
@njbrake Have you run the CICD tests against these changes? Seems like they are throwing TimeOut errors for the Smolagents SSE I would like to double check with you before assuming it was my changes

Issue #92, opened 2025-04-15
Standardized output from agent.run()


njbrake:
We use open inference as a way to standardize the tracing that comes from each agent. However, we don't currently have any standardized return and value from agent run, so although you can use any framework to run an agent easily, we don't have an easy way to reliably access the output values from running the agent.

We should probably look through how each framework returns the answer and then make a pydantic base model that would work reasonably well for all of them

njbrake:
Oooh I like that idea. So basically we can return an object that has the final answer but also has the entire trace that they can access?

PR #103, opened 2025-04-16
refactor(frameworks): Implement `_load_tools` in base class.

daavoo:


src/any_agent/frameworks/agno.py
```diff
@@ -40,18 +34,19 @@ def _get_model(self, agent_config: AgentConfig):
         )
 
     async def _load_agent(self) -> None:
+        if not agno_available:
+            msg = "You need to `pip install 'any-agent[agno]'` to use this agent"
+            raise ImportError(msg)
+        if self.managed_agents:
+            msg = "Managed agents are not yet supported in Agno agent."
+            raise NotImplementedError(msg)
+
```
njbrake, line 43: Why do this here instead of in the init?  I think I shifted it out of _load_agent and into the __init__ last week 😅

src/any_agent/frameworks/agno.py
```diff
@@ -40,18 +34,19 @@ def _get_model(self, agent_config: AgentConfig):
...
                 visit_webpage,
             ]
-        tools, mcp_servers = await wrap_tools(
-            self.config.tools, agent_framework=AgentFramework.AGNO
-        )
-        # Add to agent so that it doesn't get garbage collected
-        self._mcp_servers = mcp_servers
-        mcp_tools = [mcp_server.tools for mcp_server in mcp_servers]
-        tools.extend(mcp_tools)
+        tools, _ = await self._load_tools(self.config.tools)
```
njbrake, line 49: Is it necessary to pass self.config.tools as a parameter, or can _load_tools access it via self.config.tools since it's already a member function?

src/any_agent/frameworks/agno.py
daavoo: Right, I forgot to make one update. 

I am removing the __init__ from the subclasses and implementing it in the parent class.

The __init__ with exception that you added is never reached because Python already raises a TypeError when you try to instantiate `AnyAgent()`: `TypeError: Can't instantiate abstract class AnyAgent with abstract methods ...`

src/any_agent/frameworks/agno.py
daavoo: I leave tools as argument because it is also used in the managed_agents like `self._load_tools(managed_agent.tools)`

ELC: Then maybe you can leave it as optional with default value `None` and use `self.config.tools` in case it is `None`

src/any_agent/frameworks/google.py
```diff
@@ -20,53 +19,39 @@
...
-        if not adk_available:
-            msg = "You need to `pip install 'any-agent[google]'` to use this agent"
-            raise ImportError(msg)
         self.managed_agents = managed_agents
         self.config = config
         self._agent = None
-        self._mcp_servers = None
-        self._managed_mcp_servers = None
+        self._mcp_servers = []
+        self.framework = AgentFramework.GOOGLE
```
ELC, line 31: Adding this change seem like a good oportunity to introduce Discriminated Unions with Pydantic, I can draft a PR after this one is merged.

That will simplify several places of the logic when creating framework-specific agents

src/any_agent/frameworks/agno.py
daavoo: um. no strong opinion but I think it is more clear to leave it as a required arg

src/any_agent/frameworks/google.py
njbrake: Cool yeah that would be good. We put Discriminated Unions into our Lumigator project codebase a few months ago and have been happy with that design pattern 👍

src/any_agent/frameworks/any_agent.py
```diff
@@ -63,32 +89,22 @@ def create(
...
+    async def _load_agent(self) -> None:
+        """Load the agent instance."""
+
     @abstractmethod
     async def run_async(self, prompt: str) -> Any:
         """Run the agent asynchronously with the given prompt."""
 
     @property
     @abstractmethod
-    def tools(self) -> list[Tool]:
```
daavoo, line 80: @njbrake I felt like dropping this, we can revisit later if we want to provide some utils to access framework internals.

My reasoning is that:
- It is hard to get a consistent behavior (tools was not accounting for managed_agents, returns different stuff depending on the framework)
- We should focus on standardizing input (AgentConfig) and output #92 , but not "internals".

WDYT?

njbrake: I'm ok with this, just so long as you have some way to assure that the mcp tools are correctly getting merged in with the regular tools, I want to have tests that ensure that both regular and mcp tools are being given to each agent framework.

daavoo: I will update the integration test for that

PR #105, opened 2025-04-17
fix(lint): refactor imports and remove circular imports

ELC:
This should be merged after #95


njbrake:
Hi @ELC, thank you for the contribution! Can you explain a bit more the rational behind these changes? I'm not sure about the changes to the any_agent/__init file, only exporting a few main classes was to make it a bit simpler for the user of the package to know what main classes to use.

I don't have a good handle of the pro/cons of making some of the changes you recommend here. Thanks!

ELC:
Hi @njbrake, functionality wise, this has no impact more than allowing all imports to be from the top level package. It is somewhat standard in the Python ecosystem, think of numpy being everything np.xxxx rather than np.xxx.yyy.zzz, other popular libraries like matplotlib and pandas do the same.

However there are project which do not follow this, like scikit-learn or langchain. Other take a hibrid approach like pydantic.

I propose this PR because I believe it adds flexibility to the maintaners, by encouring final users to always import from the top level module, the project could be easily refactored or totally re-structured without affecting any consumer as their imports will always stay the same (provided they did it at the top level).

One of such refactors could be a new feature `X` which is only implemented in framework `W`, here in any-agent, we could start with a file `x.py`, then framework `Y` also incorporates feature `X`, now we could modify `x.py` to have a base class and two concrete implementations, one for `W` and one for `Y`, soon after more frameworks add support for feature `X`, now the `x.py` contains too many things and we would like to divide it into a folder `x` with files `base.py`, `w.py`, `y.py` and so on, it would also need a `__init__.py` with the relevant imports from those files.

If we only exposed imports on a "exposed-by-need" basis, such a refactor could break many imports somewhere else in the project or final users, with all the imports being exposed all the way up to the top level and, telling users to always use the most generic import possible, that refactor will not break any imports anywhere as the import system in Python does not care if `x` is a file or a `module` with `__init__.py`

This precise situation is happening right now with MCPs, for telemetry there is a file per framework, same for the agents themselves, but all MCPs live in a single file. I am doing the same refactor I mentioned above in #105 particularly [this file](https://github.com/mozilla-ai/any-agent/pull/105/files#diff-6cc7212af1ea272719466e0b06a3dc54fb0ba100d859e4a11e73efac885afdc7).

Let me know your thoughts, I don't have any strong opinions and I'm fine which watchever you decide as maintainer, but I thought as fluctuant as the AI Agent ecosystem is, this could be a good quality of life improvement on the repo as a whole.

njbrake:
Thanks for the clarification @ELC. Its not clear what this PR is addressing now, since it's out of date with the main branch. It may be best to close this one out in favor of some of the other smaller PRs that you've made.

ELC:
This only modify inits, I can make the update after merging the MCP #112

PR #107, opened 2025-04-17
Tracing as a setting when creating the Agent

njbrake:
In order to simplify the process for users, refactor setup_tracing to be a part of creating an agent, so that the user doesn't need to call setup_tracing, they can pass a TracingConfig object into the create function, and then the trace path will be stored as a member variable of the AnyAgent object.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,13 +40,10 @@ pip install 'any-agent[all]'
 To define any agent system you will always use the same imports:
 
 ```py
-from any_agent import AgentConfig, AgentFramework, AnyAgent
-from any_agent.tracing import setup_tracing  # Optional, but recommended
+from any_agent import AgentConfig, AgentFramework, AnyAgent, TracingConfig
 
 # See all options in https://mozilla-ai.github.io/any-agent/frameworks/
 framework = AgentFramework("smolagents")
-
-setup_tracing(framework)
 ```
 
 ### Single agent
@@ -60,6 +57,7 @@ agent = AnyAgent.create(
         instructions="Use the tools to find an answer",
         tools=[search_web, visit_webpage]
     )
+    TracingConfig(output_dir="output") # Optional, but recommended for saving and viewing traces
 )
 
 agent.run("Which Agent Framework is the best??")
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ agent = AnyAgent.create(
         instructions="Use the tools to find an answer",
         tools=[search_web, visit_webpage]
     )
-    TracingConfig(output_dir="output") # Optional, but recommended for saving and viewing traces
+    TracingConfig(output_dir="traces") # Optional, but recommended for saving and viewing traces
 )
 
 agent.run("Which Agent Framework is the best??")
--- a/docs/tracing.md
+++ b/docs/tracing.md
@@ -5,31 +5,29 @@ standardized [OpenTelemetry](https://opentelemetry.io/) traces for any of the su
 
 ## Example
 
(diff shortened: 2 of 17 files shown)


README.md
```diff
@@ -60,6 +57,7 @@ agent = AnyAgent.create(
         instructions="Use the tools to find an answer",
         tools=[search_web, visit_webpage]
     )
+    TracingConfig(output_dir="output") # Optional, but recommended for saving and viewing traces
```
daavoo, line 60: 💅 nitpick: I would prefer to use `traces` as default (`output` sounds too generic)

docs/tracing.md
```diff
@@ -5,31 +5,29 @@ standardized [OpenTelemetry](https://opentelemetry.io/) traces for any of the su
...
-        AgentConfig(
+        agent_framework=framework,
+        agent_config=AgentConfig(
                 model_id="gpt-4o",
-                tools=[search_web]
-            )
-        )
+                tools=[search_web],
+        ),
+        tracing=TracingConfig(enabled=True)
```
daavoo, line 22: um, this `enabled` doesn't exists (and I don't think it should)?

src/any_agent/frameworks/any_agent.py
```diff
@@ -47,11 +52,21 @@ def create(
...
+                    "Tracing is not yet supported for AGNO and GOOGLE frameworks. "
+                )
+            else:
+                agent.trace_filepath = setup_tracing(agent_framework, tracing)
+        asyncio.get_event_loop().run_until_complete(agent.load_agent())
         return agent
 
     @abstractmethod
-    async def _load_agent(self) -> None:
+    async def load_agent(self) -> None:
```
daavoo, line 69: I guess this is unrelated (but I agree is a good change).

We should document the API for people using `any_agent` within async code

tests/integration/test_load_and_run_agent.py
```diff
@@ -26,11 +25,6 @@ def test_load_and_run_agent(framework: AgentFramework, tmp_path: Path) -> None:
     if "OPENAI_API_KEY" not in os.environ:
         pytest.skip(f"OPENAI_API_KEY needed for {framework}")
 
-    # Agno not yet supported https://github.com/Arize-ai/openinference/issues/1302
```
daavoo, line 29: I would still like to use `tracing` in these tests

src/any_agent/config.py
```diff
@@ -27,6 +27,7 @@ class MCPSseParams(BaseModel):
 
 
 class TracingConfig(BaseModel):
+    output_dir: str = "traces"
```
daavoo, line 30: @njbrake WDYT if we allow this to be `None` as a way to disable the JSON exporter?
Feel like it might be useful for someone to have a straightforward way to disable it and would be consistent with the other props (None means disabled)

daavoo (commented):
LGT

docs/tracing.md
njbrake: Good catch... In order for our github action to catch errors in the md files, the code blocks have to be  ` ```python`, ` ```py` doesn't get checked 🤷

src/any_agent/config.py
njbrake: Sounds good to me 👍

src/any_agent/frameworks/any_agent.py
njbrake: https://github.com/mozilla-ai/any-agent/issues/121

PR #113, opened 2025-04-18
chore(pre-commit): simplify mypy dependencies

ELC:
- Remove specific instrumentation dependencies for mypy.
- Consolidate to a single dependency: any-agent[dev, all].


ELC:
Most of the remaining warning require non-trivial refactors so I added specific type ignores wherever needed. Before publishing this PR, we would need to merge #112 as types are heavily changed there

ELC:
Ready for Review, now all the checks with mypy are enabled given that the proper dependencies are installed.

This will allow for smaller contributions in the future from the community: "Pick any type: ignore and try to fix it in a PR", there are some that are straightforward and others that require some refactoring.

No functionality was changed in this PR, only adding comments and minor refactors.

CC @daavoo @njbrake

src/any_agent/frameworks/smolagents.py
```diff
@@ -86,9 +94,15 @@ async def load_agent(self) -> None:
             **self.config.agent_args or {},
         )
 
+        assert self._agent
```
daavoo, line 97: why are this asserts needed?

daavoo (approved):

PR #115, opened 2025-04-21
Refactor MCP Servers to use discriminated unions

ELC:
The purpose of this PR is to use Discriminated Unions for MCP, onces that is finished we could leverage the same pattern for telemetry, tracing and agents


ELC:
I believe this screenshot summarizes the intent of this PR:

image

Using Unions in combination with pydantic's TypeAdapter allows for Discriminated Unions which eliminate the need for mappings.

I believe there are multiple advantages of this approach but I would like to hear your thoughts @daavoo and @njbrake.

The main benefit is that now adding a framework to MCP only implies changes within the mcp folder and nowhere else. Moreover, now each MCP Type is less than 30 lines of code, which makes it much easier to maintain long term.

After merging this I will refactor the MCP tests and add exhaustive checks to make sure all cases are covered

src/any_agent/frameworks/any_agent.py
```diff
@@ -12,7 +12,7 @@
 if TYPE_CHECKING:
     from collections.abc import Sequence
 
-    from any_agent.tools.mcp import MCPServerBase
+    from any_agent.tools.mcp.mcp_server import MCPServerBase
```
daavoo, line 15: Shouldn't we use `MCPServer` here?

src/any_agent/tools/mcp/frameworks/google.py
```diff
@@ -0,0 +1,73 @@
+"""Tools for managing MCP (Model Context Protocol) connections and resources."""
```
daavoo, line 1: 💅 nitpick: Can we drop these docstrings?

src/any_agent/tools/wrappers.py
```diff
@@ -76,21 +71,7 @@ async def wrap_mcp_server(
     based on the specified agent_framework
     """
     # Select the appropriate manager based on agent_framework
```
daavoo, line 73: 💅 nitpick: (not part of this PR, but just spotted it now) I would drop this comment

src/any_agent/config.py
```diff
@@ -34,11 +34,15 @@ class MCPStdioParams(BaseModel):
...
+
 
 class MCPSseParams(BaseModel):
     url: str
-    headers: dict[str, str] | None = None
-    tools: list[str] | None = None
+    headers: Mapping[str, str] | None = None
+    tools: Sequence[str] | None = None
+
+    model_config = ConfigDict(frozen=True)
```
daavoo, line 45: I guess we could also use this in other parts (i.e. `TracingConfig`?)

daavoo (commented):
Great stuff @ELC 💘 

The refactor lgtm, left some comments but I would wait for @njbrake to review (as he is more familiar with this part)

src/any_agent/tools/mcp/frameworks/agno.py
```diff
@@ -16,30 +17,45 @@
     mcp_available = True
 
 
-class AgnoMCPServer(MCPServerBase):
-    """Implementation of MCP tools manager for Agno agents."""
+class AgnoMCPServerBase(MCPServerBase, ABC):
+    server: AgnoMCPTools | None = None
+    exit_stack: AsyncExitStack = AsyncExitStack()
```
njbrake, line 22: I'm a little worried about this change, since now all instantiations of AgnoMCPServer are going to share the same exitstack. Was this your intention?

njbrake (commented):
Nice! From a review only one potential issue, regarding AsyncExitStack creation. other than that, looks good. I'll need to do a more in depth review to look over the code locally since a lot of code was moved around and it's hard to verify that nothing was lost just by reading on github.

I should be able to finish and get it merged before the end of my day. Thanks @ELC! 🚀

PR #124, opened 2025-04-21
All providers use LiteLLM under the hood.

njbrake:
The fact that both OpenAI and Google are adding support for LiteLLM seems like a promising sign that LiteLLM is now the accepted standard for talking to a model. All of our agents now use LiteLLM, making it nice and easy for people to specify a model type once and have it work everywhere 🎉 .

This PR:

* OpenAI agent now uses LiteLLM under the hood
* api_key and api_base are upgraded to top level keys in AgentConfig, to clearly separate them from generation parameters.


docs/frameworks.md
```diff
@@ -97,19 +97,20 @@ Here you can find the frameworks currently supported in `any-agent`, along with
...
+import os
 from any_agent import AnyAgent, AgentFramework, AgentConfig
 agent = AnyAgent.create(
     "smolagents",
     AgentConfig(
         model_id="llama3.2",
-        model_args={
-            "api_base": "http://localhost:11434/v1"
-        }
+        model_args={},
```
daavoo, line 111: Let's either drop this line or use some of the actual args (i.e. temperature)

daavoo (approved):

Issue #144, opened 2025-04-23
multi agent integration tests should check that multiple agents are being used


njbrake:
When trying out the smolagents multi agent system, I found out that main agent is never actually calling multiple agents. It's just returning an output directly. It's not clear what the error is coming from just yet but it was a good reminder that we should probably have our integration tests in someway, verify that multiple agents are actually being used, so that tests in a little bit more detail.

daavoo:
so @javiermtorres , to give some extra info, we currently have a test single agent that does a dummy check to verify traces have been created:

https://github.com/mozilla-ai/any-agent/blob/ed2fc163720e03355fddf756698f8338d1c590aa/tests/integration/test_load_and_run_agent.py#L42-L44

I feel like we could take that approach and extend https://github.com/mozilla-ai/any-agent/blob/main/tests/integration/test_load_and_run_multi_agent.py to:

- Actually load the generated trace.
   A json.load just works.
- Verify that the multi-agent has been use.
   This I am not really sure about the best way. Each framework produces a slightly different trace, so the work here is checking them all and see if there is a single way to verify or we need framework-specific verification

njbrake:
To expand on David's second point: Being able to check that mulitple agents were used will probably involve making an update to add more information to the trace extraction code https://github.com/mozilla-ai/any-agent/tree/main/src/any_agent/telemetry

Basically right now we generally just extract the input/output/reasoning, but it would be good to also maybe extract some sort of "agent name" information if available. Figuring this out will involve looking at the traces that are generated to see if that info exists anywhere in a trace.

PR #145, opened 2025-04-23
Don't let smolagents crash loop

njbrake:
Multi agent system still isn't working like I expect but that's a problem for a different issue, probably https://github.com/mozilla-ai/any-agent/issues/144

### DIFF
--- a/src/any_agent/frameworks/smolagents.py
+++ b/src/any_agent/frameworks/smolagents.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
 
 
 DEFAULT_AGENT_TYPE = "CodeAgent"
+DEFAULT_MULTI_AGENT_TYPE = "ToolCallingAgent"
 DEFAULT_MODEL_CLASS = "LiteLLMModel"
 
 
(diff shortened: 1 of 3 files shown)


src/any_agent/frameworks/smolagents.py
```diff
@@ -50,12 +51,25 @@ async def load_agent(self) -> None:
...
         tools, _ = await self._load_tools(self.config.tools)
 
+        main_agent_type = getattr(
+            smolagents,
+            self.config.agent_type or DEFAULT_AGENT_TYPE,
+        )
         managed_agents_instanced = []
         if self.managed_agents:
+            if self.config.agent_type != "ToolCallingAgent":
+                msg = "Currently, the main agent in a multiagent system must be of type ToolCallingAgent"
```
daavoo, line 61: I think this is not True, I have just run this from main (just dropping instructions from main agent):

```
from any_agent import AgentConfig, AnyAgent, TracingConfig
from any_agent.tools import search_web, visit_webpage

agent = AnyAgent.create(
    "smolagents",  # See all options in https://mozilla-ai.github.io/any-agent/frameworks/
    AgentConfig(
        model_id="gpt-4.1-mini",
    ),
    managed_agents=[
        AgentConfig(
            name="search_web_agent",
            description="An agent that can search the web",
            model_id="gpt-4.1-nano",
            tools=[search_web]
        ),
        AgentConfig(
            name="visit_webpage_agent",
            description="An agent that can visit webpages",
            model_id="gpt-4.1-nano",
            tools=[visit_webpage]
        )
    ],
    tracing=TracingConfig()
)

agent.run("Which Agent Framework is the best??")
```

njbrake: Yep you're totally right, I was confused.

PR #147, opened 2025-04-23
If the interaction is empty, don't log the bars

njbrake:
Sometimes the interaction is an empty dict, so there's nothing to log

### DIFF
--- a/src/any_agent/tracing.py
+++ b/src/any_agent/tracing.py
@@ -75,7 +75,7 @@ class RichConsoleSpanExporter(SpanExporter):
 
                 style = getattr(self.tracing_config, span_kind.lower(), None)
 
-                if not style:
+                if not style or interaction == {}:
                     continue
 
                 self.console.rule(


njbrake:
To fix this:

Screenshot 2025-04-23 at 12 03 58 PM

njbrake:
> @njbrake But why is empty? Is something missing from our telemetry parser?

It was coming from openai where openinference span kind was LLM but the output message was a tool call, so we excluded it since the tool call would come next.

Unlike some of the other frameworks, openai doesn't provide back the reasoning alongside a tool call, if the tool is called there is no content.

### Open AI Tool Call associated with LLM span

"output.value": "[{\"content\": null, \"role\": \"assistant\", \"tool_calls\": [{\"function\": {\"arguments\": \"{\\\"query\\\":\\\"breweries in Lawrenceville Pittsburgh\\\"}\", \"name\": \"search_web\"}, \"id\": \"call_mQdoCCmDQF8NEUS03fDAuFk2\", \"type\": \"function\"}], \"function_call\": null, \"annotations\": []}]"


### Llama Index Tool Call associated with LLM Span

"output.value": "assistant: Thought: The current language of the user is English. I need to use a tool to search for breweries in the Lawrenceville area of Pittsburgh.\nAction: search_web\nAction Input: {\"query\": \"breweries in Lawrenceville Pittsburgh\"}",

PR #151, opened 2025-04-24
feat(any_agent): Expose `create_async`.

### LINKED ISSUE #121: Documentation for using any-agent async
Our docs talk about using any-agent with our sync api, but lets add info about using the async interface.

### LINKED ISSUE #150: Implement and expose `create_async`
> @fryz I think the issue here is that we are missing a `create_async` method.
> 
> `create` should not be used in an async context.
> 
> We exposed `load_agent` as async, but that is missing the logic of setting up the framework-specific class and tracing 

 _Originally posted by @daavoo in [#121](https://github.com/mozilla-ai/any-agent/issues/121#issuecomment-2827769878)_

daavoo:
Closes #121
Closes #150

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -89,3 +89,41 @@ agent = AnyAgent.create(
 
 agent.run("Which Agent Framework is the best??")
 ```
+
+## Async
+
+If you are running in `async` context, you should use the equivalent `create_async` and `run_async` methods:
+
+```py
+import asyncio
+from any_agent.tools import search_web, visit_webpage
+
+async def main():
+    agent = await AnyAgent.create_async(
+        "openai",
+        AgentConfig(
+            model_id="gpt-4.1-mini",
+            instructions="You are the main agent. Use the other available agents to find an answer",
+        ),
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -89,3 +89,41 @@ agent = AnyAgent.create(
 
 agent.run("Which Agent Framework is the best??")
 ```
+
+## Async
+
+If you are running in `async` context, you should use the equivalent `create_async` and `run_async` methods:
```
njbrake, line 95: Probably just a preference thing but I kinda like `acreate` and `arun` insteat of `create_async` and `run_async`, wdyt?

daavoo: Let's revisit this later (as this PR actually blocks any async usage). As in any other things, it looks like each framework has their own way of defining it (i.e. Openai `run` is actually the async version and then they have `run_sync`) 😅

daavoo:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/14645431513

PR #152, opened 2025-04-24
Filter MCP Tools and write a unit test to check it

njbrake:
Most of the work here went into the unit testing. I figured out how to mock the underlying part of the MCP lib that everyone uses.

I only mocked and test the stdio functionality because SSE isn't yet supported across frameworks. For instance, Agno just added support for it in a commit from 5 hours ago 😆  https://github.com/agno-agi/agno/pull/2892/files 

Llama Index doesn't yet support SSE but I think it's on their roadmap.

### DIFF
--- a/tests/unit/tools/test_unit_mcp.py
+++ b/tests/unit/tools/test_unit_mcp.py
@@ -5,10 +5,12 @@
 import asyncio
 import unittest
 from contextlib import AsyncExitStack
+from pathlib import Path
 from typing import Any
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
+from mcp import ClientSession
 
 from any_agent.config import AgentConfig, AgentFramework, MCPSseParams, MCPStdioParams
 from any_agent.frameworks.any_agent import AnyAgent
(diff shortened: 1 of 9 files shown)


njbrake:
The test is failing because the current latest version of Agno has a bug in it where it's not correctly including/excluding tools. It is fixed by this PR https://github.com/agno-agi/agno/commit/541a02cb1e25dc244f11230b80f178cf989700d6 which I assume will be rolled into a new version soon.

tests/unit/tools/test_unit_mcp.py
```diff
@@ -348,3 +350,91 @@ async def test_agno_mcp_sse() -> None:
...
+@pytest.mark.asyncio
+async def test_stdio_tool_filtering(
+    agent_framework: AgentFramework, tmp_path: Path, monkeypatch: Any
+) -> None:
+    setup_mock_server(monkeypatch)
+    mock_stdio_params = MagicMock(spec=MCPStdioParams)
+    mock_stdio_params.command = "test_command"
+    mock_stdio_params.args = ["arg1", "arg2"]
+    mock_stdio_params.tools = ["write_file", "read_file"]
+    mock_stdio_params.client_session_timeout_seconds = 10
```
daavoo: Why do you have to mock this? Can't you just pass an actual MCPStdioParams?

njbrake: Good point, no reason

daavoo:
> The test is failing because the current latest version of Agno has a bug in it where it's not correctly including/excluding tools. It is fixed by this PR [agno-agi/agno@541a02c](https://github.com/agno-agi/agno/commit/541a02cb1e25dc244f11230b80f178cf989700d6) which I assume will be rolled into a new version soon.

It looks like it was not fixed on that PR. 
I am sending a PR to Agno to properly fix the filtering for MCP

daavoo:
> > The test is failing because the current latest version of Agno has a bug in it where it's not correctly including/excluding tools. It is fixed by this PR [agno-agi/agno@541a02c](https://github.com/agno-agi/agno/commit/541a02cb1e25dc244f11230b80f178cf989700d6) which I assume will be rolled into a new version soon.
> 
> It looks like it was not fixed on that PR. I am sending a PR to Agno to properly fix the filtering for MCP

Pending of https://github.com/agno-agi/agno/pull/2985

PR #153, opened 2025-04-25
feat: initial test refactor for google and agno mcp servers

ELC:


ELC:
Could you guys tell me if this is a refactor you would like me to continue?

I only had the time to work on google and agno mcp servers, please check these files:

- tests/unit/tools/mcp/conftest.py
- tests/unit/tools/mcp/test_agno.py
- tests/unit/tools/mcp/test_google.py

If this looks good to you, I can continue with the other frameworks

My idea is to move all the patches/mocks to the conftest as fixtures and separate mock-centric tests from output-oriented tests. I strongly believe this way of structuring tests makes them much easier to understand and follow, even though some of the locality is lost due to the indirections.

Moreover since each framework has now its own test file, it is more straightforward to modify and keep updated.

CC @njbrake / @daavoo

njbrake:
Hi @ELC! Thanks for the PR: it may be relevant for you to look at what I'm about to merge in https://github.com/mozilla-ai/any-agent/pull/152: 

I wrote a unit test that iterates over every framework to check that tool filtering works correctly for `stdio_client`s. I like the idea of having separate files and tests for each framework's MCP code (since there may be some internals specific to a framework that we would want to check), but I think there is also space for a category of unit tests that will iterate over all the frameworks. What do you think?

In the coming week I plan to follow a similar path for sse clients (i.e. writing a single test that will iterate over each framework).

tests/unit/tools/mcp/test_agno.py
```diff
@@ -0,0 +1,50 @@
...
+from any_agent.config import AgentFramework, MCPSseParams, Tool
+from any_agent.tools import get_mcp_server
+
+
+@pytest.mark.asyncio
+@pytest.mark.usefixtures(
+    "enter_context_with_transport_and_session",
+    "agno_mcp_tools",
+    "_path_client_session",
+)
```
njbrake, line 20: I like this design, very clean and nice!

tests/unit/tools/mcp/test_agno.py
```diff
@@ -0,0 +1,50 @@
+# pylint: disable=unused-argument, unused-variable
+# Test MCP Tools Classes.
+# Disclaim
```
njbrake, line 3: This currently exists elsewhere in our code but reminder to myself to remove these 😆

tests/__init__.py
njbrake, line 1: Good call on adding the __init__ files 👍 🙏

ELC:
Ready for review

CC @njbrake

njbrake (approved):
This is a nice clean refactor 💯  🙏

PR #154, opened 2025-04-25
feat(tracing): Add `cost_info` option.

daavoo:
Can be disabled with `TracingConfig(cost_info=False)`.

Example output:

image

### DIFF
--- a/src/any_agent/config.py
+++ b/src/any_agent/config.py
@@ -56,6 +56,7 @@ class TracingConfig(BaseModel):
     tool: str | None = "blue"
     agent: str | None = None
     chain: str | None = None
+    cost_info: bool = True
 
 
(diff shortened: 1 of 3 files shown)


njbrake (commented):
If I read this right, it is not providing a "total cost" summation of the cost of all calls. I think that the total cost should be provided either as an attribute of the agent or printed at the end of the run (or maybe both).

daavoo:
> If I read this right, it is not providing a "total cost" summation of the cost of all calls. 

Correct.

> I think that the total cost should be provided either as an attribute of the agent or printed at the end of the run (or maybe both).

I will add a print total at the end in this PR.

IMO, instead of an attribute of the agent it should be part of whatever comes from #92

PR #158, opened 2025-04-25
Implement an SSE Fixture and split mcp tests into their own files

njbrake:
Adds a simple FastMCP server which is used as a session level fixture for easy MCP testing of the SSE things. Also reorganizes the tests into single files per framework


njbrake:
@ELC  This PR is very much related to your work on https://github.com/mozilla-ai/any-agent/pull/153 . 

This PR should take care of splitting the mcp tests into separate files per framework, so your PR can be a little bit more focused on reorganization of fixtures, and doesn't have to do quite as much rearranging of the test code.

tests/conftest.py
```diff
@@ -90,3 +91,55 @@ def mock_stdio_client() -> Generator[
...
+
+
+@pytest.fixture(
+    scope="session"
+)  # This means it only gets created once per test session
+async def echo_sse_server() -> AsyncGenerator[dict[str, str]]:
+    """This fixture runs a FastMCP server in a subprocess.
+    I thought about trying to mock all the individual mcp client calls,
+    but I went with this because this way we don't need to actually mock anything.
+    This is similar to what MCPAdapt does in their testing https://github.com/grll/mcpadapt/blob/main/tests/test_core.py
```
daavoo, line 129: dam. This looks hacky 😅 
No objections as I don't know a better alternative

njbrake: Believe it or not, it was much less hacky than trying to mock up all the low level parts of the mcp sse_client code 😅

PR #159, opened 2025-04-25
feat: Support `**kwargs` in `AnyAgent.run` method.

daavoo:
To be passed to the underlying framework method used to run the query.


src/any_agent/frameworks/smolagents.py
```diff
@@ -99,10 +99,10 @@ async def load_agent(self) -> None:
         if self.config.instructions:
             self._agent.prompt_templates["system_prompt"] = self.config.instructions
 
-    async def run_async(self, prompt: str) -> Any:
```
daavoo, line 102: ups

docs/index.md
```diff
@@ -127,3 +127,11 @@ async def main():
...
     asyncio.run(main())
 ```
+
+
+## Run args
+
+You can pass arbitrary `key=value` arguments to `agent.run` and they will be forwarded
+to the corresponding method used by the underlying framework.
+
+For example, you can pass `max_turns=30` when using the [OpenAI Agents SDK](https://openai.github.io/openai-agents-python/).
```
njbrake, line 137: Since this is a common setting across all(or most) of the frameworks I feel like maybe we should add this as a specific param? By allowing the passing of arbitrary key=val arg, it basically allows someone to bypass the anyagent api, which then may lock them into a specific framework. For common args like this maybe we should make a specific paramter for it.

daavoo: > For common args like this maybe we should make a specific paramter for it.

I thought about the same, but then it turns out that not all frameworks support it 😅 .

We could still update the docs to show a tab-style block with how to achieve the same in the frameworks that support it?

PR #169, opened 2025-04-28
refactor(mcp): add dependency injection to improve maintainability

ELC:
With this refactor, the idea is to divide the MCPServers into two layers:
- Server: used to leverage discriminate unions and auto-select based on:
  - MCP Tool Type [StdioIO | SSE]
  - Framework
- Connection: used to group all the framework specific logic and allow the server to inject a Fake connection to seamlessly test in unit test.

With this refactor the tests using patches can be still be used but now a more general approach can be used to test all MCP SSE Servers at once by injecting a fake Connection.

I believe this is the change needed to go in the direction you suggested @njbrake , please let me know your thoughts. I only did Agno to gather some feedback first before proceeding with the rest.

Particularly this is the test that will validate all frameworks:
https://github.com/ELC/any-agent/blob/374061656faa27f2c58b504ac6e7204f5ffcd22b/tests/unit/tools/mcp/test_unit_all_mcp.py#L45-L59

After this is merged, it could be further refactored to test all potential connection types as well, i.e. Stdio and SSE.


njbrake:
Hi @ELC, thanks for the PR: I think this design refactor looks good.

If we do this refactor we'll stop testing some of the underlying MCP implementation of each framework, which is what helped us catch bugs in the framework's code (i.e. https://github.com/agno-agi/agno/pull/2985). However, catching bugs in frameworks' code isn't something our unit tests need to be responsible for.

Only comment is that if possible, can you move the FakeMCPConnection class out of the `src` folder and into the `tests` folder, so that all of the test code exists in the `tests` folder. Thanks!

ELC:
@njbrake Ready for review

> If we do this refactor we'll stop testing some of the underlying MCP implementation of each framework, which is what helped us catch bugs in the framework's code

This is not accurate, we still retain the tests that uses patches, we are now dividing the tests into two groups, one that tests the integration in the separate framework's specific file using patches and then another group that tests without the framework's specific by injecting a fake connection. Basically testing the framework code in isolation and then ensuring all the frameworks are tested equally

ELC:
With this additional refactor by leveraging lazy fixtures we now can have a test like this:

image

image

I believe this is what you wanted @njbrake, it will automatically update everytime there is a new AgentFramework and every time there is a new member in the MCPParams Union we would need to modify the `mcp_params` fixture with the corresponding lazy fixture. Minimal additions with exponentially more tests at minimal effort

src/any_agent/tools/mcp/frameworks/llama_index.py
```diff
@@ -14,51 +17,76 @@
...
+    """Base class for LlamaIndex MCP connections."""
 
-    def _check_dependencies(self) -> None:
-        """Check if the required dependencies for the MCP server are available."""
-        self.libraries = "any-agent[mcp,llama_index]"
-        self.mcp_available = mcp_available
-        super()._check_dependencies()
+    _client: LlamaIndexMCPClient | None = PrivateAttr(default=None)
 
     @abstractmethod
```
njbrake, line 25: Maybe I missed this from a previous PR: this isn't an abstract method so `@abstractmethod` doesn't apply anymore, right?

njbrake (commented):
Thanks for the cleanup and the explanation of what will be tested ❤️ I have one clarification question about why `@abstractmethod` is decorating non-abstract functions, but other than that looks good 💯

ELC:
The reason to keep the `abstractmethod` and to make the framework specific `MCPConnection` `ABC` is more conceptual than technical. I'll elaborate...

`MCPConnnection` is more of an interface than an abstract class, I initially implemented it as a Protocol but then decided to go with the ABC for clarity. That decision was re-enforced after I moved the `_filter_tools` there which would have not been possible should it be a Protocol.

Since `list_tool` is an abstract method without any implementation, all children classes must implemented, and `MCPConnection` has `mcp_tool` which is `MCPParams`, a union, so to leverage the discriminated unions, there should be a concrete children for each member of the `MCPParams` union.

When doing it for one framework, one ends up with at least two concrete classes, one for STDIO and one for SSE, if we keep the two as direct children of `MCPConnection`, there is code duplication in the `list_tools` implementation as both classes have some framework-specific logic that is common across `stdio` and `sse`. That is why I decided to introduced a middle layer of abstraction, e.g. `LangchainMCPConnection`. However, that middle layer does not narrow the type for `mcp_tool` so to avoid abuse and prevent direct instantiation of that class, I made it also abstract, so that there is a still a need to create concrete classes for each member of the `MCPParams` union. That is why there is also a `LangchainMCPStdioConnection` and `LangchainMCPSseConnection`. However, to centralize the framework-specific-but-not-connection-specific code, I put the common logic inside `LangchainMCPConnection`'s `list_tools`, but that method is not meant to be called directly, as the first lines implies (e.g. it assumes some private attribute is set, which is only set by children classes). To ensure that that method is never called and that children classes add the connection-specific logic, I made that method abstract, while allowing the children classes to run that logic by calling `super().list_tools()`

With this approach we end up with a class hierarchy of three levels:

- MCPConnection (Abstract)
  - Framework-Specific (Abstract)
    - Connection-Specific (Concrete)

This way all the concrete classes are simultaneously framework-specific and connection-specific. To make sure this multiple-dispatch is clean the two-fold discrimination happens independenly, i.e. `LangchainMCPServerBase` discriminates by `framework` and `LangchainMCPConnection` discriminates by `mcp_tool`.

Maybe a way to think this is assuming that the `_setup_tools` in a framework specific concrete class (e.g. `LangchainMCPServerStdio`) takes only a framework-specific `mcp_connection` (e.g. `LangchainMCPConnection`) instead of the abstract `MCPConnection` but that is not possible because input parameters should be contravariant, i.e. accepting supertypes.

Let me know your thoughts

CC @njbrake

src/any_agent/tools/mcp/frameworks/llama_index.py
ELC: Replied in the PR main section

njbrake:
> The reason to keep the `abstractmethod` and to make the framework specific `MCPConnection` `ABC` is more conceptual than technical. I'll elaborate...
> 
> `MCPConnnection` is more of an interface than an abstract class, I initially implemented it as a Protocol but then decided to go with the ABC for clarity. That decision was re-enforced after I moved the `_filter_tools` there which would have not been possible should it be a Protocol.
> 
> Since `list_tool` is an abstract method without any implementation, all children classes must implemented, and `MCPConnection` has `mcp_tool` which is `MCPParams`, a union, so to leverage the discriminated unions, there should be a concrete children for each member of the `MCPParams` union.
> 
> When doing it for one framework, one ends up with at least two concrete classes, one for STDIO and one for SSE, if we keep the two as direct children of `MCPConnection`, there is code duplication in the `list_tools` implementation as both classes have some framework-specific logic that is common across `stdio` and `sse`. That is why I decided to introduced a middle layer of abstraction, e.g. `LangchainMCPConnection`. However, that middle layer does not narrow the type for `mcp_tool` so to avoid abuse and prevent direct instantiation of that class, I made it also abstract, so that there is a still a need to create concrete classes for each member of the `MCPParams` union. That is why there is also a `LangchainMCPStdioConnection` and `LangchainMCPSseConnection`. However, to centralize the framework-specific-but-not-connection-specific code, I put the common logic inside `LangchainMCPConnection`'s `list_tools`, but that method is not meant to be called directly, as the first lines implies (e.g. it assumes some private attribute is set, which is only set by children classes). To ensure that that method is never called and that children classes add the connection-specific logic, I made that method abstract, while allowing the children classes to run that logic by calling `super().list_tools()`
> 
> With this approach we end up with a class hierarchy of three levels:
> 
>     * MCPConnection (Abstract)
>       
>       * Framework-Specific (Abstract)
>         
>         * Connection-Specific (Concrete)
> 
> 
> This way all the concrete classes are simultaneously framework-specific and connection-specific. To make sure this multiple-dispatch is clean the two-fold discrimination happens independenly, i.e. `LangchainMCPServerBase` discriminates by `framework` and `LangchainMCPConnection` discriminates by `mcp_tool`.
> 
> Maybe a way to think this is assuming that the `_setup_tools` in a framework specific concrete class (e.g. `LangchainMCPServerStdio`) takes only a framework-specific `mcp_connection` (e.g. `LangchainMCPConnection`) instead of the abstract `MCPConnection` but that is not possible because input parameters should be contravariant, i.e. accepting supertypes.
> 
> To summarize, it was not _needed_ to make those abstractmethod, but it was done so deliberately to enforce a cleaner class struture, ensuring that each new member of the MCPParams union will require their own dedicated (yet small) concrete class, while keeping all the framework-specific logic centralized.
> 
> FYI - `abstractmethods` in Python can have a concrete implementation, they are just a marker to enforce concrete classes overwrite them, it is not mandatory that all `abstractmethods` have no implementation. This is a Python perk that other languages don't have. In other languages that middle layer class could have not been abstract for this very purpose.
> 
> Let me know your thoughts
> 
> CC @njbrake

Thanks for your thorough explanation 🙏  , I didn't know that info about python abstractmethods, cool to know! Ok in that case, no more comments from me at the moment, I'll try to take some time today to do a deep dive to make sure I understand what's happening, and then merge it.

njbrake:
HI @ELC there are a few issues with the implementation that show up in the integration tests 

```bash
export ANY_AGENT_INTEGRATION_TESTS=TRUE
export MCP_INTEGRATION_TESTS=TRUE
export OPENAI_API_KEY=<YOUR_KEY>
pytest tests/integration
```

If you don't yet have an OpenAI API key, they're pretty [easy to get](https://www.codecademy.com/article/creating-an-openai-api-key)

Here's a log of the failing tests, if that helps
[failing_tests.log](https://github.com/user-attachments/files/19974967/failing_tests.log)

ELC:
@njbrake Currently I'm not planning on spending on an OpenAI key, I believe the latest commit should address the issue in the logs.

Could you please re-test and let me know if there are any other issues?

src/any_agent/tools/mcp/frameworks/agno.py
```diff
@@ -15,44 +23,39 @@
...
-            msg = "MCP server is not set up. Please call `setup` from a concrete class."
+    async def list_tools(self) -> list[Tool]:
+        """List tools from the MCP server."""
+        if self._server is None:
+            msg = "MCP server is not set up. Please call `list_tools` from a concrete class."
             raise ValueError(msg)
 
-        self.tools = [await self._exit_stack.enter_async_context(self.server)]
+        tools = await self._exit_stack.enter_async_context(self._server)
+        return tools if isinstance(tools, list) else [tools]  # type: ignore[return-value, list-item]
```
njbrake, line 37: Is there any instance where tools is a list? It's not clear to me why this logic is required

src/any_agent/tools/mcp/frameworks/agno.py
```diff
@@ -15,44 +23,39 @@
...
-    server: Any | None = None  # Using `Any` to avoid circular import issues
-    framework: Literal[AgentFramework.AGNO] = AgentFramework.AGNO
-
-    def _check_dependencies(self) -> None:
-        """Check if the required dependencies for the MCP server are available."""
-        self.libraries = "any-agent[mcp,agno]"
-        self.mcp_available = mcp_available
-        super()._check_dependencies()
+class AgnoMCPConnection(MCPConnection, ABC):
+    _server: AgnoMCPTools | None = PrivateAttr(default=None)
```
njbrake, line 27: This change would undo my bugfix from https://github.com/mozilla-ai/any-agent/pull/195/files: I used the type of Any because otherwise this code will throw an error if Agno isn't installed, even if the agno agent isn't being used. I also didn't use the TYPE_CHECKING blog because doing so was throwing pydantic circular import issues.

njbrake: This comment applies to the use of "Any" in all of the MCP framework classes.

src/any_agent/tools/mcp/frameworks/google.py
```diff
@@ -19,54 +27,79 @@
...
-    framework: Literal[AgentFramework.GOOGLE] = AgentFramework.GOOGLE
+class GoogleMCPConnection(MCPConnection, ABC):
+    """Base class for Google MCP connections."""
 
-    def _check_dependencies(self) -> None:
-        """Check if the required dependencies for the MCP server are available."""
-        self.libraries = "any-agent[mcp,google]"
-        self.mcp_available = mcp_available
-        super()._check_dependencies()
+    _params: GoogleStdioServerParameters | GoogleSseServerParameters | None = (
```
njbrake, line 33: See my comment about typing in the agno mcp class

njbrake (commented):
Thanks for the updates, it's so close! The integration tests all pass now 🎉 . One thing remains to clean up, which is a regression of a bug I patched yesterday, see my comments in the MCP framework code. Thanks!

njbrake:
> @njbrake Currently I'm not planning on spending on an OpenAI key, I believe the latest commit should address the issue in the logs.
> 
> The issue was that I was returning something different than the input value in the `_filter_tools` method, due to an unintended variable shadowing on my side.
> 
> Could you please re-test and let me know if there are any other issues?

Sounds good. My apologies, I didn't mean to imply that you should spend money on the OpenAI API key. I thought there was a free usage level, but maybe that isn't the case. Anyhow, your updates fixed the integration tests so thank you 🙏

src/any_agent/tools/mcp/frameworks/agno.py
ELC: I took a different approach and use strings for the typing, that allows to keep the original type, avoid circular imports and prevent the missing import error

Checked with the uv sync command from #195

src/any_agent/tools/mcp/frameworks/agno.py
ELC: I introduced this change to make one test pass which was using several patches, now it is no longer needed, I'm reverting it to a similar implementation as before

ELC:
@njbrake Address all the comments, let me know if there is anything else I should take on to merge this

I'm pretty satisfied with how this PR turned out :D

src/any_agent/tools/mcp/mcp_server.py
```diff
@@ -33,18 +52,3 @@ def _check_dependencies(self) -> None:
 
         msg = f"You need to `pip install '{self.libraries}'` to use MCP."
         raise ImportError(msg)
-
-    def _filter_tools(self, tools: Sequence[Any]) -> Sequence[Any]:
-        # Only add the tools listed in mcp_tool['tools'] if specified
-        requested_tools = list(self.mcp_tool.tools or [])
-        if not requested_tools:
-            return tools
-        tools = [tool for tool in tools if tool.name in requested_tools]
```
njbrake, line 42: Link to #298  We missed this because we didn't have unit tests 🤦  but this logic was the thing doing the filtering of tools, which was accidentally removed in this PR

Issue #160, opened 2025-04-25
Compute and show total cost of the agent trace


daavoo:
> I will add a print total at the end in this PR.

So it turns out this is harder than I expected 😅 
I would need to move the logic to the telemetry processor(s). I will be opening an issue as a follow-up to refactor and add a total value.

_Originally posted by @daavoo in https://github.com/mozilla-ai/any-agent/issues/154#issuecomment-2830653162_

PR #174, opened 2025-04-28
Tracing as a class

njbrake:
This is the beginnings of the changes to support

https://github.com/mozilla-ai/any-agent/issues/92 

and 

https://github.com/mozilla-ai/any-agent/issues/160

By storing the trace object as a member of the agent, we'll be able to access the trace information (and maybe the cost information from the traces, not sure yet) in a clean way. The existing implementation is storing the trace_filepath as a parameter of AnyAgent separate from the tracing functions, which imo is making our lives a little more complicated than it needs to be.

### DIFF
--- a/src/any_agent/config.py
+++ b/src/any_agent/config.py
@@ -2,7 +2,7 @@ from collections.abc import Callable, Mapping, MutableMapping, Sequence
 from enum import Enum, auto
 from typing import Any, Self
 
-from pydantic import BaseModel, ConfigDict, Field,  model_validator
+from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 
 class AgentFramework(str, Enum):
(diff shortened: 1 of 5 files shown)


src/any_agent/config.py
```diff
@@ -51,13 +51,22 @@ class MCPSseParams(BaseModel):
 
 class TracingConfig(BaseModel):
     model_config = ConfigDict(extra="forbid")
-    output_dir: str | None = "traces"  # None for no json saved trace
+    output_dir: str = "traces"  # None for no json saved trace
+    enable_console: bool = True
+    enable_file: bool = True
```
daavoo, line 56: No strong opinion but I would prefer to not have these args and instead rely on `None` value.

I know this is more explicit but I think is simple enough API for someone to understand `None = disabled`.
Also, some corner cases are opened and not handled here (i.e. `enable_console=True` + all the console options set to `None / False`)

njbrake: Similar to you I don't have a strong opinion here. My thinking about adding explicit enable/disable booleans is that as we add more args that relate to each log setting, it might get confusing about how to turn it off.

For example, if right now a person wanted to disable console logging but continue with file logging, its not clear how they would disable the console. Needing to set all values to None might be a bit confusing. 

I'm thinking this will also make more sense when we incorporate some sort of file rotation functionality into the Tracer, mentioned in https://github.com/mozilla-ai/any-agent/issues/175.

njbrake: Either way, you make a good point about the corner case of enable=True with all console options set to None

PR #177, opened 2025-04-28
Consistent return no matter which agent framework

njbrake:
Now when the Agent.run returns, you always can get a reasonable answer at result.final_output, and some tracing when available at `raw_responses`.

We'll add some more keys to this output object soon, like execution time, cost, etc. 

We also will improve on raw_responses by populating it with the normalized trace object. But this PR should be nice enough to get us started.


daavoo (commented):
Good start !

Let's update the integration tests to assert for `AgentResult.final_output` and `raw_responses`

PR #180, opened 2025-04-29
Check test traces for multiagent handoffs

javiermtorres:
Traces in multiagent tests should be checked for handoffs across agents.

### DIFF
--- a/src/any_agent/frameworks/openai.py
+++ b/src/any_agent/frameworks/openai.py
@@ -13,8 +13,9 @@ try:
         Handoff,
         Model,
         ModelSettings,
-        Runner,
         RunContextWrapper,
+        Runner,
+        Tool,
         function_tool,
     )
     from agents.extensions.models.litellm_model import LitellmModel
@@ -122,7 +123,7 @@ class OpenAIAgent(AnyAgent):
             kwargs_["model_settings"] = ModelSettings(**self.config.model_args)
 
         # TODO create tool function calls from function calls
-        wrapped_tools = [
+        wrapped_tools: list[Tool] = [
             function_tool(
                 tool, failure_error_function=self.any_agent_tool_error_function
             )
--- a/src/any_agent/frameworks/openai.py
+++ b/src/any_agent/frameworks/openai.py
@@ -2,9 +2,6 @@ import traceback
 from typing import TYPE_CHECKING, Any
 from uuid import uuid4
 
-from agent.run_context import RunContextWrapper
-from agent.tools import function_tool
-
 from any_agent.config import AgentConfig, AgentFramework, TracingConfig
 from any_agent.tools import search_web, visit_webpage
 
@@ -17,6 +14,8 @@ try:
         Model,
         ModelSettings,
         Runner,
+        RunContextWrapper,
+        function_tool,
     )
     from agents.extensions.models.litellm_model import LitellmModel
 
--- a/src/any_agent/frameworks/openai.py
+++ b/src/any_agent/frameworks/openai.py
@@ -1,5 +1,9 @@
+import traceback
 from typing import TYPE_CHECKING, Any
 from uuid import uuid4
+
+from agent.run_context import RunContextWrapper
+from agent.tools import function_tool
 
 from any_agent.config import AgentConfig, AgentFramework, TracingConfig
 from any_agent.tools import search_web, visit_webpage
@@ -57,6 +61,13 @@ class OpenAIAgent(AnyAgent):
             api_key=agent_config.api_key,
         )
 
+    @staticmethod
+    def any_agent_tool_error_function(
+        ctx: RunContextWrapper[Any], error: Exception
+    ) -> str:
+        traceback.print_exception(error)
+        return f"An error occurred while running the tool. Please try again. Error: {error!s}"
+
     async def _load_agent(self) -> None:
         """Load the OpenAI agent with the given configuration."""
         if not agents_available:
@@ -111,13 +122,21 @@ class OpenAIAgent(AnyAgent):
         if self.config.model_args:
             kwargs_["model_settings"] = ModelSettings(**self.config.model_args)
 
+        # TODO create tool function calls from function calls
+        wrapped_tools = [
+            function_tool(
+                tool, failure_error_function=self.any_agent_tool_error_function
+            )
+            for tool in tools
+        ]
+
         self._main_agent_tools = tools
         self._agent = Agent(
             name=self.config.name,
             instructions=self.config.instructions,
             model=self._get_model(self.config),
             handoffs=handoffs,
-            tools=tools,
+            tools=wrapped_tools,
             mcp_servers=[mcp_server.server for mcp_server in mcp_servers],
             **kwargs_,
         )
--- a/src/any_agent/frameworks/openai.py
+++ b/src/any_agent/frameworks/openai.py
@@ -1,4 +1,3 @@
-import traceback
 from typing import TYPE_CHECKING, Any
 from uuid import uuid4
 
@@ -13,10 +12,7 @@ try:
         Handoff,
         Model,
         ModelSettings,
-        RunContextWrapper,
         Runner,
-        Tool,
-        function_tool,
     )
     from agents.extensions.models.litellm_model import LitellmModel
 
@@ -60,13 +56,6 @@ class OpenAIAgent(AnyAgent):
             base_url=agent_config.api_base,
             api_key=agent_config.api_key,
         )
-
-    @staticmethod
-    def any_agent_tool_error_function(
-        ctx: RunContextWrapper[Any], error: Exception
-    ) -> str:
-        traceback.print_exception(error)
-        return f"An error occurred while running the tool. Please try again. Error: {error!s}"
 
     async def _load_agent(self) -> None:
         """Load the OpenAI agent with the given configuration."""
@@ -122,21 +111,13 @@ class OpenAIAgent(AnyAgent):
         if self.config.model_args:
             kwargs_["model_settings"] = ModelSettings(**self.config.model_args)
 
-        # TODO create tool function calls from function calls
-        wrapped_tools: list[Tool] = [
-            function_tool(
-                tool, failure_error_function=self.any_agent_tool_error_function
-            )
-            for tool in tools
-        ]
-
         self._main_agent_tools = tools
         self._agent = Agent(
             name=self.config.name,
             instructions=self.config.instructions,
             model=self._get_model(self.config),
             handoffs=handoffs,
-            tools=wrapped_tools,
+            tools=tools,
             mcp_servers=[mcp_server.server for mcp_server in mcp_servers],
             **kwargs_,
         )
--- a/src/any_agent/tools/web_browsing.py
+++ b/src/any_agent/tools/web_browsing.py
@@ -17,7 +17,7 @@ def _truncate_content(content: str, max_length: int) -> str:
 
 
 def search_web(query: str) -> str:
-    """Perform a duckduckgo web search based on your query (think a Google search) then returns the top search results.
+    """Perform a duckduckgo web search based on your query (think a Google search) then returns the top search results. Do this even if the model cannot answer the query by itself.
 
     Args:
         query (str): The search query to perform.
--- a/src/any_agent/tools/web_browsing.py
+++ b/src/any_agent/tools/web_browsing.py
@@ -17,7 +17,7 @@ def _truncate_content(content: str, max_length: int) -> str:
 
 
 def search_web(query: str) -> str:
-    """Perform a duckduckgo web search based on your query (think a Google search) then returns the top search results. Do this even if the model cannot answer the query by itself.
+    """Perform a duckduckgo web search based on your query (think a Google search) then returns the top search results.
 
     Args:
         query (str): The search query to perform.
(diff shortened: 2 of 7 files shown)


src/any_agent/tools/web_browsing.py
```diff
@@ -17,7 +17,7 @@ def _truncate_content(content: str, max_length: int) -> str:
 
 
 def search_web(query: str) -> str:
-    """Perform a duckduckgo web search based on your query (think a Google search) then returns the top search results.
+    """Perform a duckduckgo web search based on your query (think a Google search) then returns the top search results. Do this even if the model cannot answer the query by itself.
```
njbrake, line 20:
```suggestion
    """Perform a duckduckgo web search based on your query (think a Google search) then returns the top search results.
```

The function description shouldn't bias the model to use the tool. The info about whether the model should use the tool and how often belongs in the system_prompt/instructions, not the function definition

tests/conftest.py
```diff
@@ -93,6 +93,53 @@ def _patch_stdio_client() -> Generator[
...
+def check_multi_tool_usage_langchain(json_logs: List[Dict]):
+    # find trace that has 
+    for json_log in json_logs:
+        attrs = json_log["attributes"]
+        name = json_log["name"]
+        if "metadata" in attrs:
+            metadata = json.loads(attrs["metadata"])
+            if attrs["openinference.span.kind"] == "AGENT" and metadata["langgraph_step"] == 2 and name.startswith("transfer_to_"):
+                return
+    raise Exception("")
```
njbrake, line 109: This type of extraction belongs in the src/any_agent/telemetry code. All access of trace specific info should be interfaced through those classes imo

njbrake (commented):
Good start! I think the trace extraction logic may be best suited for the TelemetryProcessor

src/any_agent/tools/web_browsing.py
javiermtorres: Ok

tests/conftest.py
javiermtorres: I'd argue that this is test code, no src code, so a separate function should be used.

tests/conftest.py
```diff
@@ -93,6 +93,53 @@ def _patch_stdio_client() -> Generator[
...
+
+
+def check_multi_tool_usage_langchain(json_logs: List[Dict]):
+    # find trace that has 
+    for json_log in json_logs:
+        attrs = json_log["attributes"]
+        name = json_log["name"]
+        if "metadata" in attrs:
+            metadata = json.loads(attrs["metadata"])
+            if attrs["openinference.span.kind"] == "AGENT" and metadata["langgraph_step"] == 2 and name.startswith("transfer_to_"):
```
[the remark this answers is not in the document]
javiermtorres: Totally brittle. I changed to just locating the "TOOL" kinds in the traces, which should be consistent with the OI conventions; but frameworks will sometimes add their own tools (like final answers) so I assume we'll need to change the criteria every now and then. For now, using the span kind sorta works.

src/any_agent/tools/web_browsing.py
njbrake: I don't think this is resolved?

tests/conftest.py
```diff
@@ -93,6 +94,63 @@ def _patch_stdio_client() -> Generator[
...
+
+check_multi_tool_usage_dict = {
+    AgentFramework.GOOGLE: check_multi_tool_usage_google,
+    AgentFramework.LANGCHAIN: check_multi_tool_usage_langchain,
+    AgentFramework.LLAMA_INDEX: check_multi_tool_usage_llamaindex,
+    AgentFramework.OPENAI: check_multi_tool_usage_openai,
+    AgentFramework.AGNO: check_multi_tool_usage_agno,
+    AgentFramework.SMOLAGENTS: check_multi_tool_usage_smolagents,
+    AgentFramework.TINYAGENT: check_multi_tool_usage_tinyagent,
+}
```
njbrake, line 144: If the only difference in the calls is that `check_multi_tool_usage_llamaindex` passes in the number 2 instead of 1, can this dict and the calling mechanism be simplified?

tests/integration/test_load_and_run_multi_agent.py
```diff
@@ -23,29 +28,33 @@ def test_load_and_run_multi_agent(
             f"Skipping test for {agent_framework.name} because it does not support multi-agent"
         )
 
-    kwargs["model_id"] = "gpt-4.1-nano"
+    agent_model = "gpt-4.1-nano"
+    # kwargs["model_id"] = "gemini/gemini-2.0-flash-lite"
+    # kwargs["model_id"] = "gemini/gemini-2.5-flash-preview-04-17"
+    kwargs["model_id"] = agent_model
```
njbrake, line 34: This will be reverted prior to merge right? Same question for the commented out model_args a few lines below

tests/integration/test_load_and_run_multi_agent.py
```diff
@@ -54,22 +63,33 @@ def test_load_and_run_multi_agent(
...
     )
-    result = agent.run("Which agent framework is the best?")
 
     assert result
     assert result.final_output
     if agent_framework not in [AgentFramework.LLAMA_INDEX]:
         # Llama Index doesn't currently give back raw_responses.
         assert result.raw_responses
         assert len(result.raw_responses) > 0
+    if agent_framework not in (AgentFramework.AGNO, AgentFramework.GOOGLE):
```
njbrake, line 89:
```suggestion
    if agent_framework not in (AgentFramework.AGNO, AgentFramework.GOOGLE, AgentFramework.TINYAGENT):
```

tests/conftest.py
```diff
@@ -93,6 +94,63 @@ def _patch_stdio_client() -> Generator[
...
+
+def check_multi_tool_usage_all(json_logs: list[dict[str, Any]], min_tools: int) -> None:
+    tools = 0
+    for json_log in json_logs:
+        attrs = json_log["attributes"]
+        if attrs["openinference.span.kind"] == "TOOL":
+            tools += 1
+    assert tools < min_tools, (
+        "Count of tool usage is too low, managed agents were not used"
+    )
```
njbrake, line 133: Can logic about extracting the tools used/how many agents were used be put into the TelemetryProcessor class? Logic to figure out how many tools are used and what agents used them is valuable not just to unit tests but for evaluation and general developer access. 

Is there no other key in the span that indicates that an agent was used?  It's not clear to me why multiple tools being used means that multiple agents were used. If you ran these checks on the `test_load_and_run_agent.py` tests (single agent configs), would these also pass for them?

src/any_agent/tools/web_browsing.py
javiermtorres: No, not yet.

tests/conftest.py
javiermtorres: I expected this to be much more customized for every platform, even if right now we're only using a very simple check. I'd leave it as it is, since we're still checking how OI traces look like in different frameworks.

tests/conftest.py
javiermtorres: Hm. I'd not use the same exact code in source and tests as it can make bugs difficult to spot later on. I can add a similar function to src, but I'd wait until we really check if the TOOL span kind is consistent across OI instrumentation packages.

Right now in this multi agent test, we have two tools that use agents, so yes, in our case the tool count is the same as the agent count.

I had the impression that the AGENT span kinds are used for more purposes, so it's not so easy to count. Furthermore, current frameworks afaict are concerned mainly about single agents with tools, so the TOOL span kind may have a more defined and stable meaning.

tests/integration/test_load_and_run_multi_agent.py
```diff
@@ -99,5 +99,12 @@ def test_load_and_run_multi_agent(
             assert cost_sum.total_cost < 1.00
             assert cost_sum.total_tokens > 0
             assert cost_sum.total_tokens < 20000
+        if agent_framework not in (AgentFramework.AGNO, AgentFramework.GOOGLE, AgentFramework.TINYAGENT):
```
njbrake, line 91: We now have a convenient `is_tracing_supported` function that you should use here.

tests/conftest.py
njbrake: Thanks for the input 👍 . At a minimum then, if we keep this logic and don't do any more detailed checks, can you file a Github issue on the openinference repo to explain this problem so that they can fix it?

tests/conftest.py
njbrake: I would prefer to keep the code as concise as possible here until we need to expand it out. If you don't mind, would it be ok if you simplified the logic this time? Feel free to come tell me "i told you so" if this has to be re-split out in the future 😆

javiermtorres: NP to change it 🙂 And no need to tell you anything!!!! 😋

tests/conftest.py
javiermtorres: At least I can try to provide the traces for our multi agent case and show the differences across frameworks, yes. But I guess this is going to be very dependent on the framework :-/

tests/integration/test_load_and_run_multi_agent.py
```diff
@@ -85,5 +115,15 @@ def test_load_and_run_multi_agent(agent_framework: AgentFramework) -> None:
             assert cost_sum.total_cost < 1.00
             assert cost_sum.total_tokens > 0
             assert cost_sum.total_tokens < 20000
+        if agent_framework not in (
+            AgentFramework.AGNO,
+            AgentFramework.GOOGLE,
+            AgentFramework.TINYAGENT,
+        ):
```
njbrake, line 122: I think you can use the _is_tracing_supported() helper function here?

njbrake (approved):

src/any_agent/frameworks/openai.py
```diff
@@ -57,6 +61,13 @@ def _get_model(
             api_key=agent_config.api_key,
         )
 
+    @staticmethod
```
daavoo, line 64: What is this change about? seems unrelated to the test?

javiermtorres: It enables us to get a traceback in the case of errors. Otherwise we have a very concise "there was an error around here" that doesn't really give any information :-/

daavoo: I get that, I was asking why added as part of this PR 😅 

BTW, it seems to me that this is a problem with the openinference instrumentation, as OpenAI does expose exception info when the tool usage fails. Will take a look as part of the work I am currently doing

javiermtorres: > I was asking why added as part of this PR 

Well, it helped me to troubleshoot something in this PR :D

tests/integration/test_load_and_run_multi_agent.py
```diff
@@ -1,19 +1,54 @@
...
+    level=logging.DEBUG,
+    format=FORMAT,
+    datefmt="[%X]",
+    handlers=[RichHandler(rich_tracebacks=True)],
+)
+logger = logging.getLogger("any_agent_test")
+logger.setLevel(logging.DEBUG)
+
+
+CHILD_TAG = "any_agent.children"
```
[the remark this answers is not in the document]
njbrake: Why is this test the correct place to put this logic?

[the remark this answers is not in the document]
javiermtorres: We could set a test-wide logger at conftest, if that's the question. Since other tests don't make use of a logger right now, I didn't want to impose design choices (and make further changes). But we can change the scope of this logger.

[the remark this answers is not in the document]
njbrake: No worries! In that case yep I'm fine to keep it here for the time being 👍

src/any_agent/frameworks/openai.py
njbrake: For the purposes of this PR, would you be ok removing this code, and then you can make a new PR where you suggest this change and we discuss in that place?

javiermtorres: Ok, I'll remove it.

tests/unit/frameworks/test_openai.py
```diff
@@ -16,10 +16,14 @@ def test_load_openai_default() -> None:
     mock_agent = MagicMock()
     mock_function_tool = MagicMock()
     mock_litellm_model = MagicMock()
+    mock_ctx_wrapper = MagicMock()
```
njbrake, line 19:
```suggestion
```

tests/unit/frameworks/test_openai.py
```diff
@@ -16,10 +16,14 @@ def test_load_openai_default() -> None:
...
+    mock_tool = MagicMock()
 
     with (
         patch("any_agent.frameworks.openai.Agent", mock_agent),
         patch("agents.function_tool", mock_function_tool),
+        patch("agents.tool", mock_tool),
+        patch("agents.RunContextWrapper", mock_ctx_wrapper),
```
njbrake, line 26:
```suggestion
```

njbrake (commented):
so close!

njbrake (approved):
Looks good! Once linting/tests pass, clear to merge. (you may want to run the integration tests locally or via manual trigger in github, because they dont' run automatically)

javiermtorres:
Integration tests don't pass in some cases :( But it's a known issue; I sometimes cannot get the agent to use tools, and the main agent will try to answer by itself. Let's deal with that in a follow up PR.

PR #183, opened 2025-04-29
AgentResults contains token cost and usage, when available

njbrake:


src/any_agent/config.py
```diff
@@ -53,18 +53,18 @@ class MCPSseParams(BaseModel):
 class TracingConfig(BaseModel):
     model_config = ConfigDict(extra="forbid")
     output_dir: str = "traces"  # None for no json saved trace
-    enable_console: bool = True
-    enable_file: bool = True
+    console: bool = True
+    file: bool = True
+    cost_info: bool = True
```
njbrake, line 58: @daavoo  does this rename make it a little less painful to have the bool toggles? cost_info now controls both console printing and general cost tracking.

src/any_agent/tracing.py
```diff
@@ -176,15 +245,23 @@ def __init__(
...
+        )  # Fail fast if framework is not supported
         self.tracing_config = tracing_config
         self.trace_filepath: str | None = None
         self._setup_tracing()
 
+    def __del__(self) -> None:
+        """Stop the openinference instrumentation when the tracer is deleted."""
+        if self.instrumentor:
+            self.instrumentor.uninstrument()
+
```
njbrake, line 259: @daavoo  Keep an eye out for how instrumentation is handled as you deep dive into openinference. I think we will have an issue if a single process tries to load multiple AnyAgent() because the openinference instrumenter is managing some sort of global state and gets confused if you try to double instrument it (aka re-instrument opeanai when openai is already instrumented)

README.md
```diff
@@ -69,7 +69,7 @@ agent = AnyAgent.create(
     tracing=TracingConfig(output_dir="traces") # Optional, but recommended for saving and viewing traces
 )
 
-agent.run("Which Agent Framework is the best??")
+result: AgentResult = agent.run("Which Agent Framework is the best??")
```
njbrake, line 72: I added the typing not only for readability but also because then our CI tests will throw an error if we change the import location or name of the result class "which we discuss in the text part of the docs"

src/any_agent/config.py
daavoo: sounds good to me. 
Btw, we should probably add docstrings for each attribute so they are rendered in the API reference

src/any_agent/tracing.py
daavoo: > I think we will have an issue if a single process tries to load multiple AnyAgent() because the openinference instrumenter is managing some sort of global state and gets confused if you try to double instrument it 

Yes, the instrumentations don't expect that and we should handle it on our side , somehow. 

This was one of the benefits of having setup_tracing outside (not saying we should move it back)

README.md
daavoo: um. but do you think we should put it in the README? it feels like just more verbose

src/any_agent/config.py
daavoo: one minor thing is I would replace `file` with something more clear, don't know what (json export?)

daavoo (commented):
I am not 100% sold on this change, but no strong opinion.

In my mind, it would be cleaner to not make choices about what fields should be returned and instead either:

-  Provide a way for the user to load the trace into an standard class (i.e. something like `agent.load_trace()` 
- Return the standard trace class directly as result (probably makes more sense since we already started this direction)
  Exposing helper method/properties to get stuff from it like the final output / cost / etc.

njbrake:
> I am not 100% sold on this change, but no strong opinion.
> 
> In my mind, it would be cleaner to not make choices about what fields should be returned and instead either:
> 
>     * Provide a way for the user to load the trace into an standard class (i.e. something like `agent.load_trace()`
> 
>     * Return the standard trace class directly as result (probably makes more sense since we already started this direction)
>       Exposing helper method/properties to get stuff from it like the final output / cost / etc.

Thanks for the input, yeah I don't have a super strong opinion about my exact implementation. I think I would like the result of agent.run to contain the information needed to access the information from the trace, instead of it being attached to the agent: this is because there is a scenario where agent.run() is being called many times in parallel (if it's running in a server) and having to call `agent.get_info_I_need()` might be tricky for the agent to keep track of what run the client is asking about. 

I like your idea of returning a trace class directly and then having a user call methods on that object to retrieve info like cost/usage/etc. Let me jam on that for a bit and update this PR. 🙏

src/any_agent/tracing.py
njbrake: Ok I think I have a working design now, at least something to iterate on.

README.md
njbrake: I'm fine with removing it to keep the README simple

src/any_agent/telemetry/types.py
njbrake, line 1: This is the ugliest file in the whole PR and is responsible for the large file count difference. I got stuck in this pickle where I wanted to turn the OTEL ReadableSpan into something that was serializable with the pydantic model, so I turned several OTEL types into pydantic models so that they would be serializable. See the `AnyAgentSpan.from_readable_span` method. 

As I get more familiar with how OTEL works and what we want to do with our telemetry hopefully we can streamline this approach and get rid of this code.

src/any_agent/tracing.py
```diff
@@ -4,56 +4,177 @@
...
+
+class AnyAgentTrace(BaseModel):
+    """A trace that can be exported to JSON or printed to the console."""
+
+    spans: list[AnyAgentSpan]
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    def get_cost_summary(self) -> TotalTokenUseAndCost:
```
daavoo, line 100: 💅 

```suggestion
    def get_total_cost(self) -> TotalTokenUseAndCost:
```

src/any_agent/tracing.py
```diff
@@ -4,56 +4,177 @@
...
+            if key in self.attributes:
+                logger.warning("Overwriting attribute %s with %s", key, value)
+            self.attributes[key] = value
+
+    def set_attribute(self, key: str, value: AttributeValue) -> None:
+        """Set a single attribute for the span."""
+        return self.set_attributes({key: value})
+
+
+class AnyAgentTrace(BaseModel):
```
daavoo, line 93: (Not a blocker for this PR) 

I wonder if we really need this class instead of a list of `AnyAgentSpan` 🤔  

If `AnyAgentSpan` implements `get_cost` then we would just need to do a sum() somewhere. Not sure if it's worth the abstraction layer we introduce.

We could instead document how users can get info from that list instead of implementing ourselves.

Thinking about other potential use cases, like counting the number of tool calls, I would prefer if we go more in this direction:

```py
# agent_trace is just list[AnyAgentSpan]
tool_calls = [span for span in agent_trace if span.is_tool]
```

Than this one:

```py
tool_calls = agent_trace.get_tool_calls()
```

@njbrake WDYT?

src/any_agent/telemetry/types.py
daavoo: TBH I am kind of surprised no one has implemented some pydantic models for the OTEL base classes.

> and get rid of this code.

I don't see why, it feels like this would be useful. Maybe not for the "base types" but we could implement pydantic models for "openinference types" / "otel genai types" to validate

daavoo: Right, I now see most of these classes are just to be able to json.dump?.

An alternative approach would be to use a custom JSONEncoder (https://docs.python.org/3/library/json.html#json.JSONEncoder) and handle the serialization of each OTEL class using `isinstance` or something

daavoo (commented):
Overall LGTM, left some comments for discussion

src/any_agent/tracing.py
```diff
@@ -176,46 +272,65 @@ def __init__(
     ):
         """Initialize the Tracer and set up tracing filepath, if enabled."""
         self.agent_framework = agent_framework
+        # Set it to None at first To avoid AttributeError on __del__ if _get_instrumenter_by_framework throws exception
+        self.instrumentor: Instrumenter | None = None
```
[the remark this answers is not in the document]
njbrake: Oooh great find, I didn't even realize I was mixing the spelling of it 🙈 . Googling tells me that it's a en-us vs en-uk word difference. I think I like instrumentEr better so I'll update everything to that.

src/any_agent/tracing.py
njbrake: Hmm. So I did think about avoiding the creation of `AnyAgentTrace.` However, it seemed to me that it would be nice for end users not to need to understand what's in each span to get information out from the trace.

I think there is some opportunity for more cleanup of the `TelemetryProcessor` classes too that may inform more of the design decisions arount AnyAgentTrace

README.md
```diff
@@ -99,7 +99,7 @@ agent = AnyAgent.create(
     ]
 )
 
-agent.run("Which Agent Framework is the best??")
+result = agent.run("Which Agent Framework is the best??")
```
njbrake, line 102:
```suggestion
agent_trace = agent.run("Which Agent Framework is the best??")
```

PR #184, opened 2025-04-29
Unit Testing of Evaluation Framework

njbrake:
In order to enable testing, it's also helpful to re-organize into a more class based approach.


njbrake:
@daavoo  this is still a draft and not a working re-factor yet, but if you have time, take a quick look over to see if you like the direction it's heading.

src/any_agent/evaluation/runner.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def _run_telemetry_eval(self, test_case: TestCase, telemetry_path: str) -> None:
+        with open(telemetry_path) as f:
+            telemetry: list[dict[str, Any]] = json.loads(f.read())
+        logger.info(f"Telemetry loaded from {telemetry_path}")
+
+        agent_framework = TelemetryProcessor.determine_agent_framework(telemetry)
+
+        processor = TelemetryProcessor.create(agent_framework)
+        hypothesis_answer = processor._extract_hypothesis_answer(trace=telemetry)
```
daavoo, line 43: IMO, we should decouple the processor logic that we use to generate console outputs from any evaluation logic. 

Maybe we need to prioritize what I commented in https://github.com/mozilla-ai/any-agent/pull/183#pullrequestreview-2807512775 and have a standard `AgentTrace` class that can be loaded from the trace file.

Then, evaluation works from there and the `extract_stuff` for evaluation should be implemented in the `evaluation` module

njbrake: This is a good comment. I will keep this in mind while I work on #183 💯

njbrake: Ok cool I think I have it implemented in #183, take a look at that and let me know what you think. Sorry it's such a big PR now 🙈

docs/evaluation.md
```diff
@@ -33,21 +33,52 @@ agent = AnyAgent.create(
...
-test_case = TestCase.from_yaml("./docs/examples/test_case.yaml")
-evaluate_telemetry(test_case, '/path/to/telemetry/output')
+from any_agent.evaluation import EvaluationRunner
+from any_agent.evaluation.evaluation_case import EvaluationCase
+output_path="tmp/path/result.json"
+evaluation_case = EvaluationCase(
+    ground_truth=[{"name": "Test Case 1", "value": 1.0, "points": 1.0}],
+    checkpoints=[{"criteria": "Check if value is 1.0", "points": 1}],
+    llm_judge="gpt-4o-mini",
+    final_answer_criteria=[]
```
daavoo, line 76: 💅 Maybe use `final_output` instead of `final_answer` for consistency with the name of the AgentTrace property

docs/evaluation.md
```diff
@@ -33,21 +33,52 @@ agent = AnyAgent.create(
     tracing=TracingConfig(output_dir="traces")
 )
 
-agent.run("How many seconds would it take for a leopard at full speed to run through Pont des Arts?")
+agent_trace = agent.run("How many seconds would it take for a leopard at full speed to run through Pont des Arts?")
+# If doing an evaluation via a separate python process, the trace can also be dumped to a json file:
+# with open("agent_trace.json", "w", encoding="utf-8") as f:
```
daavoo, line 38: Can't we load the exported trace inside `traces`?

pyproject.toml
```diff
@@ -83,6 +83,10 @@ dev = [
   "pre-commit",
 ]
 
+evaluation = [
```
daavoo, line 86: I am fine with adding this to the default deps instead of creating a new group

src/any_agent/evaluation/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .evaluation_runner import EvaluationRunner
```
daavoo, line 1: 💅 Maybe also expose `EvaluationCase` here?

src/any_agent/evaluation/cli.py
```diff
@@ -0,0 +1,47 @@
...
+from any_agent.config import AgentFramework
+from any_agent.evaluation import EvaluationRunner
+from any_agent.evaluation.evaluation_case import EvaluationCase
+from any_agent.logging import logger
+from any_agent.tracing.trace import AgentSpan, AgentTrace
+
+
+def do_eval(
+    evaluation_case_paths: list[str],
+    telemetry_paths: list[str],
```
daavoo, line 14: 💅 I was trying to use `tracing` instead of `telemetry` elsewhere

src/any_agent/evaluation/evaluation_case.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+# This can't go into a TYPE_CHECKING block because it's used at runtime by Pydantic to build the fields
```
daavoo, line 3: btw: https://docs.pydantic.dev/latest/concepts/performance/#sequence-vs-list-or-tuple-with-mapping-vs-dict

I think we could type with `list`

src/any_agent/evaluation/evaluation_runner.py
```diff
@@ -0,0 +1,169 @@
...
+        self.qa_evaluator = QuestionAnsweringSquadEvaluator()
+
+    def add_evaluation_case(self, evaluation_case: EvaluationCase) -> None:
+        """Add test case file path to the evaluation runner."""
+        if evaluation_case not in self._evaluation_cases:
+            self._evaluation_cases.append(evaluation_case)
+        else:
+            logger.warning("Test case %s already added.", evaluation_case)
+
+    def add_trace(self, trace: AgentTrace, agent_framework: AgentFramework) -> None:
```
daavoo, line 39: So, the idea is that you might use different traces inside a same evaluation run?

src/any_agent/evaluation/evaluation_runner.py
```diff
@@ -0,0 +1,169 @@
...
+        else:
+            logger.warning("Telemetry %salready added.", trace)
+
+    def _run_trace_eval(
+        self,
+        evaluation_case: EvaluationCase,
+        trace: AgentTrace,
+        agent_framework: AgentFramework,
+    ) -> None:
+        processor = TracingProcessor.create(agent_framework)
```
daavoo, line 53: not sure if we already discuss this, but ideally here we would not need to import `TracingProcessor` and have the extract logic defined within `evaluation`

tests/unit/evaluation/sample_traces/OPENAI.json
daavoo, line 1: not a blocker: Can we use instead a list of span fixtures (similar to the `llm_span` that already exists)? So we can reuse them across tests.
Or is there something in particular that we would be missing if we don't load from JSON?

tests/unit/evaluation/test_evaluation_runner.py
```diff
@@ -0,0 +1,150 @@
...
+    second_case.ground_truth[0]["value"] = 2.0
+
+    runner.add_evaluation_case(evaluation_case)
+    runner.add_evaluation_case(second_case)
+
+    # Add multiple traces
+    second_trace = agent_trace.model_copy(deep=True)
+    second_trace.spans[0].name = "Different Span"
+
+    runner.add_trace(agent_trace, AgentFramework.OPENAI)
```
daavoo, line 71: Why is the framework needed in `add_trace`?

src/any_agent/evaluation/evaluation_runner.py
```diff
@@ -0,0 +1,169 @@
...
+        hypothesis_answer_results = self.hypothesis_evaluator.evaluate(
+            hypothesis_final_answer=hypothesis_answer,
+            ground_truth_answer_dict=evaluation_case.ground_truth,
+            ground_truth_checkpoints=evaluation_case.final_answer_criteria,
+        )
+
+        if evaluation_case.ground_truth:
+            if not self.qa_evaluator:
+                msg = "QuestionAnsweringSquadEvaluator not initialized."
+                raise ValueError(msg)
```
daavoo, line 78: How/When can these exceptions be reached? Can we add a test for them?

daavoo (approved):
Just several nitpick (💅 ) comments and a couple of questions.
Nothing blocker

tests/unit/evaluation/test_results_saver.py
```diff
@@ -0,0 +1,60 @@
+import os
+from pathlib import Path
+
+import pandas as pd
```
daavoo, line 4: do we really need pandas here?

PR #186, opened 2025-04-29
feat: add uv to create envs for pre-commit

ELC:


ELC:
@njbrake Ready for review

njbrake (commented):
Is there a more "uv-centric" approach? For instance, should we switch the lint github action to use UV instead of base python? https://docs.astral.sh/uv/guides/integration/github/#installation 

See my recent addition of UV into the tests github action: https://github.com/mozilla-ai/any-agent/blob/main/.github/workflows/tests.yaml#L24-L28

ELC:
@njbrake Let me know your thoughts about this second iteration

.github/workflows/lint.yaml
```diff
@@ -15,18 +15,21 @@ jobs:
...
 
       - uses: actions/cache@v4
         with:
           path: ~/.cache/pre-commit/
           key: pre-commit-4|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
 
       - name: pre-commit
-        run: pre-commit run --all-files
+        env:
+          FORCE_PRE_COMMIT_UV_PATCH: 1
```
njbrake, line 34: Can you add a comment or something to explain why this is necessary? I see it listed in here https://github.com/tox-dev/pre-commit-uv but I'm not an expert with pre-commit and don't really understand the explanation in that repo 😬

.github/workflows/lint.yaml
```diff
@@ -15,18 +15,21 @@ jobs:
...
         with:
-          python-version: '3.11'
+          python-version: 3.13
+          activate-environment: true
 
       - name: Install pre-commit
-        run: pip install pre-commit
+        run: uv sync --extra lint
 
       - uses: actions/cache@v4
```
daavoo, line 27: Is this cache still needed? It seems that if we use UV we should follow:
https://docs.astral.sh/uv/guides/integration/github/#caching

ELC:
I made several tests, the conclusion is that caching does not affect the execution time in any significant way - impact was at most 18s (10%). I suggest not implementing caching at all unless the benefit outweights the maintenance burden.

Below are the results of the tests.

The big bottleneck is running mypy but Astral is already working in [an alternative](https://github.com/astral-sh/ty) that is rust based, like ruff. It would be worth revisiting in a couple of months.

@njbrake please review.

--- 

Result of comparing caching strategies:

Using pre-commit cache and uv run - cold start: 2min 15s - [Source](https://github.com/mozilla-ai/any-agent/actions/runs/14815426383/job/41595237966#step:6:1)
image

Using pre-commit cache and uv run - warm: 2min 5s - [Source](https://github.com/mozilla-ai/any-agent/actions/runs/14833484896/job/41639641510#step:6:1)
image

Using uv caching and uv run - cold start: 2min 20s - [Source](https://github.com/mozilla-ai/any-agent/actions/runs/14918916212/job/41910454647?pr=186#step:5:1)
image

Using uv caching and uv run - warm:  2 min 23s - [Source](https://github.com/mozilla-ai/any-agent/actions/runs/14918999773/job/41910692533?pr=186#step:5:1)
image

No explicit caching and uv run: 2 min 18s - [Source](https://github.com/mozilla-ai/any-agent/actions/runs/14919049160/job/41910827280?pr=186#step:5:1)
image

PR #191, opened 2025-05-02
docs: Create `agents` page. Add separate page for each framework.

daavoo:


docs/agents.md
```diff
@@ -0,0 +1,181 @@
...
+    ]
+)
+```
+
+## Running Agents
+
+Regardless of the definition (single-agent or multi-agent), you can run the
+agent as follows:
+
+```py
```
njbrake, line 80: Currently, the code block has to be ` ```python` not `  ```py` for it to be run by our docs tester

PR #197, opened 2025-05-02
Add agent.exit and delete MCP test by merging into existing test

njbrake:
While implementing the cleanup of tests I found the bug where we weren't properly uninstrumenting the agent or releasing the mcp servers when we were done with them, which meant that integration tests were stomping on each other. By adding an exit method this allows application developers to control when resources get cleaned up, without needing to rely on garbage collection timing of when __del__ would get called.


docs/agents.md
```diff
@@ -133,7 +133,7 @@ if __name__ == "__main__":
 
 For example, you can pass `output_type` when using the [OpenAI Agents SDK](https://github.com/openai/openai-agents-python):
 
-```py
+```pythnon
```
njbrake, line 136:
```suggestion
```python
```

tests/integration/test_load_and_run_agent.py
```diff
@@ -1,31 +1,75 @@
...
+        return True if result.returncode == 0 else False  # noqa: TRY300
+    except Exception:
+        return False
 
 
 @pytest.mark.skipif(
     os.environ.get("ANY_AGENT_INTEGRATION_TESTS", "FALSE").upper() != "TRUE",
     reason="Integration tests require `ANY_AGENT_INTEGRATION_TESTS=TRUE` env var",
 )
+# @pytest.mark.asyncio
```
njbrake, line 30:
```suggestion
```

PR #198, opened 2025-05-05
Add-generics-for-mcp-tools

ELC:
This PR [is stacked ](https://graphite.dev/blog/stacked-prs) on top of #169


ELC:
This PR aims to make things more explicit and transparent, every framework has their own class for "Tool" and the current implementation naively assumes everything will be compliant with any_anget.config.Tool, this may not be the case for all frameworks.

By introducing Generics, we make this transition more transparent, using string annotations prevents requiring the libraries to be installed and it allows mypy to detect issues early on ([shifting left](https://en.wikipedia.org/wiki/Shift-left_testing)) instead on solely relying on integration tests.

It would be great to have another discriminated union for the tools which provides a unified api for all frameworks, but that seems either a [leaky abstraction ](https://en.wikipedia.org/wiki/Leaky_abstraction) or something that be [brittle](https://en.wikipedia.org/wiki/Software_brittleness) will likely break on framework non-backwards compatible changes. We can explore this in a separate PR if there is insterest though.

The idea is to only appply that to MCPs in this PR and have a separate one for Agents, to avoid changing too many files at once

@njbrake What are your thoughts?

PR #202, opened 2025-05-05
Refactor around returning and using `AgentTrace`

daavoo:
TODO:

- [ ] Extract evaluation logic from `tracing/processors`

### DIFF
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -5,7 +5,7 @@
 To define any agent system you will always use the same imports:
 
 ```python
-from any_agent import AgentConfig, AnyAgent, TracingConfig
+from any_agent import AgentConfig, AnyAgent
 # In these examples, the built-in tools will be used
 from any_agent.tools import search_web, visit_webpage
 ```
@@ -35,7 +35,6 @@ agent = AnyAgent.create(
         instructions="Use the tools to find an answer",
         tools=[search_web, visit_webpage]
     ),
-    tracing=TracingConfig(output_dir="traces") # Optional, but recommended for saving and viewing traces
 )
 ```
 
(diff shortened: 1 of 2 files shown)


src/any_agent/tracing/exporter.py
```diff
@@ -0,0 +1,102 @@
...
+from any_agent.logging import logger
+
+from .processors import TracingProcessor
+from .trace import AgentSpan, AgentTrace
+
+if TYPE_CHECKING:
+    from opentelemetry.sdk.trace import ReadableSpan
+
+
+class AnyAgentExporter(SpanExporter):
```
daavoo, line 25: @njbrake You could implement here the:

> include things like the agent framework that was used, which would be a much more reliable method to load the trace when it comes to evaluation

However, let's make sure that the resulting JSON is compatible with OTEL, whatever that means

njbrake (approved):

src/any_agent/frameworks/any_agent.py
```diff
@@ -41,21 +40,46 @@ def __init__(
     ):
         self.config = config
         self.managed_agents = managed_agents
+
         self._mcp_servers: list[MCPServerBase] = []
-        self._last_tracer: Tracer | None = (
-            None  # holds the trace from the most recent run
-        )
-        self._tracing_config = tracing
```
njbrake, line 48: Whoops I missed this during the initial review. Making this change means that now all the `agent.run()` commands from an agent are going to get merged into a single trace. That's going to be an issue for people that are running the agent in a server or want to have multiple runs happening in a single instantation of the agent.

```   
 result = agent.run("What's the best agent framework")
 result2 = agent.run("What's the best agent framework")
```
result2 is going to be result and result2 appended on top of it

PR #205, opened 2025-05-05
Reset tracing at the beginning of each run

njbrake:
This doesn't have to be the final design but at least for the moment it solves the problem of `agent.run`s appending on top of each other

### DIFF
--- a/src/any_agent/frameworks/agno.py
+++ b/src/any_agent/frameworks/agno.py
@@ -109,7 +109,7 @@ class AgnoAgent(AnyAgent):
         if not self._agent:
             error_message = "Agent not loaded. Call load_agent() first."
             raise ValueError(error_message)
-
+        self._setup_tracing()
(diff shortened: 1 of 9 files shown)


daavoo (commented):
@njbrake This works, so I am 👍 with it.

No strong opinion but maybe `AnyAgentExporter` could have a `reset` method or something and we call that.

My concern with the current code is that it calls `TraceProvider` stuff multiple times (you should see a OTEL warning about it) when we would only need to call it the first time and then reset the spans of the exporter

njbrake:
> @njbrake This works, so I am 👍 with it.
> 
> No strong opinion but maybe `AnyAgentExporter` could have a `reset` method or something and we call that.
> 
> My concern with the current code is that it calls `TraceProvider` stuff multiple times (you should see a OTEL warning about it) when we would only need to call it the first time and then reset the spans of the exporter

Sounds good. I'm totally open for more changes, this was slightly rushed so that we would get the bug fixed in main.

PR #206, opened 2025-05-05
Re-organize console logging in exporter

njbrake:


src/any_agent/tracing/exporter.py
```diff
@@ -34,71 +34,78 @@ def __init__(  # noqa: D107
...
+        if self.tracing_config.console:
+            self.console = Console()
+
         if self.tracing_config.save:
             if not os.path.exists(self.tracing_config.output_dir):
                 os.makedirs(self.tracing_config.output_dir)
+            timestamp = datetime.now().strftime("%Y-%m-%d-%H-%M-%S")
+            self.output_file = f"{self.tracing_config.output_dir}/{self.agent_framework.name}-{timestamp}.json"
+            # Currently, we pass this file back to the user via the AgentTrace object returned by agent.run.
+            # This is flaky since nothing prevents the user from moving the file after it's saved, which then would
```
daavoo, line 52: I think is not a big worry, any file output is susceptible to that

src/any_agent/tracing/trace.py
```diff
@@ -182,6 +182,15 @@ def get_total_cost(self) -> TotalTokenUseAndCost:
             total_cost_completion=total_cost_completion,
         )
 
+    def add_span(self, span: AgentSpan) -> None:
```
daavoo, line 185: I guess do you plan to expand this somehow?

daavoo (approved):

src/any_agent/tracing/trace.py
njbrake: This ties into the conversation I was having with Javier on slack: if we go down the route of keeping all the important information inside of OTEL, then we would maybe need to add an attribute for things like `agent_framework` into every span that we add. Still thinking about it.

daavoo:
@njbrake should we close this one?

njbrake:
> @njbrake should we close this one?

Let's leave it open and I'll repurpose it so that it's just the reorg of the console writing code. I'll take care of it tomorrow

src/any_agent/tools/user_interaction.py
```diff
@@ -1,3 +1,5 @@
+from rich.prompt import Prompt
```
daavoo, line 1: ❤️

PR #209, opened 2025-05-06
Saving AgentTrace to a file is now the user/developers responsibility.

njbrake:
Per our discussion, it is no longer the responsibility of any-agent to manage and save output files for an AgentTrace. the developer who calls `agent.run()` is now responsible to save it to a file if they want it.

### DIFF
--- a/tests/unit/tracing/test_unit_exporter.py
+++ b/tests/unit/tracing/test_unit_exporter.py
@@ -34,14 +34,6 @@ def test_rich_console_span_exporter_disable(llm_span: ReadableSpan):  # type: ig
         console_mock.return_value.rule.assert_not_called()
 
 
-def test_save_default(llm_span: ReadableSpan):  # type: ignore[no-untyped-def]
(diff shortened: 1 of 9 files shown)


tests/unit/tracing/test_unit_exporter.py
```diff
@@ -34,31 +28,28 @@ def test_rich_console_span_exporter_disable(llm_span: ReadableSpan):  # type: ig
...
         exporter.export([llm_span])
         console_mock.return_value.rule.assert_not_called()
 
 
-def test_save_default(tmp_path, llm_span: ReadableSpan):  # type: ignore[no-untyped-def]
-    exporter = AnyAgentExporter(
-        AgentFramework.LANGCHAIN,
-        TracingConfig(output_dir=str(tmp_path), console=False, save=True),
-    )
+def test_save_default(llm_span: ReadableSpan):  # type: ignore[no-untyped-def]
```
daavoo, line 37: I guess we can drop this test?

daavoo (approved):
9t1q01

PR #210, opened 2025-05-06
feat: Add initial implementation of `serving` module.

daavoo:
Before going down too many rabbit holes, I decided to open this for early discussion.

How to test it:

---

- Run this in one terminal:

```python
from any_agent.config import ServingConfig
from any_agent.tools import search_web

agent = AnyAgent.create(
    "google",
    AgentConfig(
        name="google_expert",
        model_id="gpt-4.1-nano",
        instructions="You must use the available tools to find an answer",
        description="An agent that can answer questions specifically about the Google Agents Development Kit (ADK).",
        tools=[search_web]
    )
)

agent.serve(ServingConfig(port=5001))
```

- Run this in another terminal:

```py
from any_agent.tools import search_web

agent = AnyAgent.create(
    "openai",
    AgentConfig(
        name="openai-expert",
        model_id="gpt-4.1-nano",
        instructions="You must use the available tools to find an answer",
        description="An agent that can answer questions specifically about the OpenAI Agents SDK.",
        tools=[search_web]
    )
)

agent.serve()
```

- Follow the setup in https://github.com/google/A2A/blob/main/demo/README.md#running-the-examples

- Run the demo UI

- Add both agents through the UI (this tests the Agent Card stuff https://google.github.io/A2A/specification/agent-card/):

Screenshot 2025-05-06 204455

- Start a conversation through the UI

Screenshot 2025-05-06 205424

- You should see the traces in the corresponding terminal and check the events logs in the UI:

Screenshot 2025-05-06 205412


src/any_agent/serving/agent_card.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def _get_agent_card(agent: "AnyAgent", serving_config: "ServingConfig") -> AgentCard:
+    skills = []
+    for tool in agent.config.tools:
+        # TODO: handle MCP tools
+        if isinstance(tool, MCPParams):
+            continue
+        skills.append(
+            # TODO: find what other arguments can be set.
```
daavoo, line 20: The remaining fields are `tags`, `examples`, `inputModes` and `outputModes`

src/any_agent/serving/agent_card.py
```diff
@@ -0,0 +1,37 @@
...
+                name=tool.__name__,
+                description=inspect.getdoc(tool),
+            )
+        )
+    return AgentCard(
+        name=agent.config.name,
+        description=agent.config.description,
+        version=serving_config.version,
+        url=f"http://{serving_config.host}:{serving_config.port}/",
+        # TODO: extend default capabilities
```
daavoo, line 32: I wanted to keep this to the minimum for now. 

Adding `streaming` capabilities would require us to implement some sort of `run_stream` method in AnyAgent, which I am not sure it's worth it.

I haven't check what the other 2 capabilities mean in practice.

src/any_agent/serving/agent_card.py
```diff
@@ -0,0 +1,37 @@
...
+
+if TYPE_CHECKING:
+    from any_agent import AnyAgent
+    from any_agent.config import ServingConfig
+
+
+def _get_agent_card(agent: "AnyAgent", serving_config: "ServingConfig") -> AgentCard:
+    skills = []
+    for tool in agent.config.tools:
+        # TODO: handle MCP tools
```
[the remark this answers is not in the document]
njbrake: You would need to first connect to the MCP server to retrieve the tools available. Could this function be called after the tools are loaded by the agent?

src/any_agent/serving/server.py
```diff
@@ -0,0 +1,28 @@
+from typing import TYPE_CHECKING
+
+try:
+    from common.server import A2AServer
```
njbrake, line 4: So this is some sort of example server from the A2A examples repo?

njbrake:
Cool stuff! Since the server we're using is a part of the "samples" folder of the A2A repo, should we basically copy all the code from that "common" folder?

src/any_agent/serving/agent_card.py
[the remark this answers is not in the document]
daavoo: > Could this function be called after the tools are loaded by the agent?

It is already called after agent creation, so I guess I will add a helper to do that

njbrake:
How about we merge this code, and then request to have the example code you provide here added as a folder in https://github.com/google/A2A/tree/main/samples/python/agents?

pyproject.toml
```diff
@@ -150,12 +156,12 @@ disallow_untyped_calls = true
...
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "session"
 
-
 [project.scripts]
 any-agent-evaluate = "any_agent.evaluation.cli:main"
+
+[tool.uv.sources]
+a2a-samples = { git = "https://github.com/google/A2A", subdirectory = "samples/python" }
```
njbrake, line 167: In the absence of the a2a sdk this seems ok for now.

src/any_agent/serving/agent_card.py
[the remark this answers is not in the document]
njbrake: The `_load_tools` function returns all the tools (both MCP and callables) in the `tools` list, so you could iterate over that list

src/any_agent/serving/task_manager.py
```diff
@@ -0,0 +1,98 @@
+from collections.abc import AsyncIterable
+from typing import TYPE_CHECKING
+
+try:
```
njbrake, line 4: You could use the `with suppress(ImportError):` syntax that we adopted in the MCP code

njbrake:
In terms of final usage, I was thinking something like this

Config 1 (equivalent to MCPStdio): this process spins up and serves the child agents
```python

agent_1 = AnyAgent.create(...)
agent_2 = AnyAgent.create(...)

host_agent = AnyAgent.create(..., connected_agents=[agent_1,agent_2])
host_agent.run("help me answer this thing")
```

Config 2: I'm gonna mess up the python syntax but you'll get the idea (this is like MCPSse)
```python
agent_1 = AnyAgent.create(...)
agent_2 = AnyAgent.create(...)

proc_1 = subprocess.run(agent_1.serve(localhost:8000),detach=True) # this is the wrong syntax
proc_2 = subprocess.run(agent_2.server(localhost:8001),detach=True) # still wrong syntax

host_agent = AnyAgent.create(..., connected_agents=["localhost:8000", "localhost:8001"])
host_agent.run("help me answer this thing")

proc_1.kill()
proc_2.kill()
```

Maybe we should also add an "examples" folder in any-agent to have sample scripts about how to do this stuff, idk

daavoo:
> Config 1 (equivalent to MCPStdio): this process spins up and serves the child agents

I don't think this is worth implementing. We provide `.serve`, people can figure (we can have examples) on best way to consume it (see next point)

> Config 2: I'm gonna mess up the python syntax but you'll get the idea (this is like MCPSse)

For this, the way I was thinking is more like:

```py
from any_agent.tools import A2ATool

host_agent = AnyAgent.create(..., tools=[A2ATool("localhost:8000"), A2ATool("localhost:8001")])
```

This would create an A2A client and expose it as a function that can be called like another tool.
This is basically what all frameworks already do with "Agent as tools", but we would providing a generic wrapper instead of relying on the framework-specific wrapping

njbrake:
> > Config 1 (equivalent to MCPStdio): this process spins up and serves the child agents
> 
> I don't think this is worth implementing. We provide `.serve`, people can figure (we shall have examples) on different ways to consume it (see next point)

I think it's probably ok if we don't implement this, but comparing in terms of MCP, the MCPStdio is advertised as "this gives you the option to easily develop, but you probably wouldn't want to use it for production".  If we decide to skip this implementation option with A2A I think it risks making the process more complicated for a developer to build and debug their system when they're getting up and running. 

I imagine that early on in the process there will be a lot of tweaking happening with each agent, and forcing a developer to manage the restarting of each agent as they make changes seems like undesireable overhead.

I guess before we discard this configuration I'd like to get an idea of how hard it is to implement, because if it's simple enough, this seems like a valuable configuration to help lower the barrier for entry.

> 
> > Config 2: I'm gonna mess up the python syntax but you'll get the idea (this is like MCPSse)
> 
> For this, the way I was thinking is more like:
> 
> ```python
> from any_agent.tools import A2ATool
> 
> host_agent = AnyAgent.create(..., tools=[A2ATool("localhost:8000"), A2ATool("localhost:8001")])
> ```
> 
> This would create an A2A client and expose it as a function that can be called like another tool. This is basically the same all frameworks do today with "Agent as tools" (without a client), but we would providing a generic wrapper instead of relying on the framework-specific wrapping


Cool, works for me 👍 . In the A2A demo, that "host_agent" that is the demo UI, are the "connected agents" added as tools for that agent? I guess it's not clear to me (because I don't know A2A that well yet) whether its correct to think of an A2A connected agent as a tool, or whether it should be a separate category.

njbrake (approved):
I wonder whether there is a cleaner implementation that using the new _main_agent_tools member variable, but I don't think it's a blocker. So exciting to see this come together!

I'd also love to see an integration test for A2A but I'm ok also with that coming later.

PR #211, opened 2025-05-07
feat: refactor wrap tools to use discriminated unions

ELC:
In this PR I am replacing all the logic for the tool wrappers by a similar approach taken in the Agents and in the MCP connections, that is, using a generic abstract AnyTool class that the rest of the codebase can rely on.

So instead of dynamically wrapping any callable, we can statically map each framework to a specific class by using a discriminated union and then return tools using a unified interface while keeping compatibility with the underlying framework.

I believe the main benefit of this is to have a unified api to interact with tools regardless of the framework used under the hood, it will allow for stronger typing controls both at the agent and at the MCP level.

This is still work in progress although all the unit tests pass, there are several changes needed in MCP and Agents to use `AnyTool` instead of `Tool` or simly `Any`.

How do you like this refactor @njbrake? Any guideline or direction you want me to follow? 

For starters, I will need to move the MCP folder outside of tools to avoid circular imports with this approach, but that is a simple change, I won't do it right away to avoid polluting the diff.


njbrake:
Hi @ELC , thank you for the input. From the user usability perspective, I now see code like 

```python
tools=[SmolagentsTool(tool=search_web), SmolagentsTool(tool=visit_webpage)],
```

where it used to be 
```python
tools=[search_web,visit_webpage],
```

I like the idea of being able to use an `AnyTool` type for consistent validation and usage of tools, but I wouldn't want it to come at the expense of making it more complex for the developer. Was adding that `SmolagentsTool` wrapper just a convenience for the tests, or would this be a new requirement on the developer when they are creating the agent?

njbrake:
@ELC Thanks for the clarification. In that case, since there isn't any impact to the API for the developer building on top of any-agent, I think adding the AnyTool typing makes sense 👍

PR #214, opened 2025-05-07
feat(config): Add `validate_console_flags` to `TracingConfig`.

daavoo:


src/any_agent/config.py
```diff
@@ -99,6 +99,13 @@ class TracingConfig(BaseModel):
     cost_info: bool = True
     """Whether traces should include cost information"""
 
+    @model_validator(mode="after")
+    def validate_console_flags(self) -> Self:
+        if self.console and not any([self.llm, self.tool, self.agent, self.chain]):
```
njbrake, line 104: would it be a valid case if all these were turned off but cost_info was turned on?

daavoo: No. cost wouldn't be displayed in the console (but it would be computed and available in the AgentTrace)

PR #217, opened 2025-05-07
Add pytest timeout

njbrake:
Add a timeout so that a hanging test won't hang the process.


pyproject.toml
```diff
@@ -155,6 +156,7 @@ follow_untyped_imports = true
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "session"
+timeout = 300
```
daavoo, line 159: we could be more "restrictive" (and add timeout to specific tests )

PR #225, opened 2025-05-07
Concurrent `agent.run` should not overlap

njbrake:
We have a bug where the tracing gets replaced by whichever `agent.run` happens last, which contaminates the results of concurrent agent.runs.

### DIFF
--- a/tests/integration/test_load_and_run_agent.py
+++ b/tests/integration/test_load_and_run_agent.py
@@ -1,3 +1,4 @@
+import asyncio
 import os
 import subprocess
 from datetime import datetime
@@ -100,8 +101,8 @@ def test_load_and_run_agent(agent_framework: AgentFramework, tmp_path: Path) -> 
     finally:
         agent.exit()
 
-
-def test_run_agent_twice(agent_framework: AgentFramework) -> None:
+@pytest.mark.asyncio
+async def test_run_agent_twice(agent_framework: AgentFramework) -> None:
     """When an agent is run twice, state from the first run shouldn't bleed into the second run"""
     model_id = "gpt-4.1-nano"
     env_check = validate_environment(model_id)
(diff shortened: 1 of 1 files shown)


njbrake:
@daavoo Ok, this is obviously an incorrect fix and can't be the final product, but it helps to illustrate the current disconnect. There are two problems in our current code:

1. We only had a single exporter per agent instantiation, which meant that even if we replaced the _instrumentor for each agent.run call, that would replace the instrumentor for everybody, so it breaks in concurrent situations
2. We don't have any way (that I know of) so that the exporter knows which spans it should pay attention to vs ignore. Each trace comes with a trace id, but if the agent.run_async doesn't know which trace id is theres, how do we appropriately retrieve their trace when the time comes.

Ideally we would have a single exporter that handles and separates out all of the in-process traces, but I'm still trying to figure out how the agent.run call will know what their trace_id is, which is what it would need in order to retrieve its trace. The hacky thing in this PR is that I do the matching based on whether their prompt string is in the span, but that doesn't work as soon as two concurrent agent.run calls are using similar prompts (and in general its just a bad design).

src/any_agent/frameworks/agno.py
```diff
@@ -109,7 +108,7 @@ async def run_async(self, prompt: str, **kwargs: Any) -> "AgentTrace":
         if not self._agent:
             error_message = "Agent not loaded. Call load_agent() first."
             raise ValueError(error_message)
-        self._setup_tracing()
         result: RunResponse = await self._agent.arun(prompt, **kwargs)
-        self._exporter.trace.final_output = result.content
-        return self._exporter.trace
```
njbrake, line 115: Since Agno doesn't yet support tracing, we directly create the AgentTrace here

src/any_agent/frameworks/any_agent.py
```diff
@@ -134,27 +134,18 @@ async def _load_tools(
 
     def _setup_tracing(self) -> None:
```
njbrake, line 135: We now follow the original design where setup_tracing only gets called in the AnyAgent __init__. We will still get those OTEL warnings when multiple AnyAgents are created in a single process, but this at least improves the design

src/any_agent/frameworks/langchain.py
```diff
@@ -132,12 +133,16 @@ async def run_async(self, prompt: str, **kwargs: Any) -> "AgentTrace":
...
             error_message = "Agent not loaded. Call load_agent() first."
             raise ValueError(error_message)
-        self._setup_tracing()
         inputs = {"messages": [("user", prompt)]}
-        result = await self._agent.ainvoke(inputs, **kwargs)
+        run_id = str(uuid4())
+        tracer = self._tracer_provider.get_tracer("any_agent")
+        with tracer.start_as_current_span("agent_run") as span:
+            span.set_attribute("any_agent.run_id", run_id)
+            result = await self._agent.ainvoke(inputs, **kwargs)
```
njbrake, line 141: Here's how this works:

1. Make run UUID
2. Make all spans created by ainvoke be associated with the agent_run span, with which we attached the run_id attribute
3. The exporter will see this span and then have the info we need to associate the run_id (which the Agent knows about) with the trace_id (that the Agent doesn't know about)
4. Pop that trace out of the exporter so that we free up the memory since we don't need it anymore once the agent retrieves it.

daavoo: Do we need to create the parent span ourselves?

Afaik, openinference will (correctly?) create a parent span (ideally, type "Agent") already with an unique ID. 

In our exporter, we could just:

- If the trace doesn't have parent id -> Store it in `run_trace_mapping`.
- If the trace has a parent id -> Get it from `run_trace_mapping`.

I am missing something?

njbrake: If we don't create the parent span ourselves, how do we set the run_id attribute that we need so that the agent knows which trace_id is associated with the run?

PR #230, opened 2025-05-09
fix(AgnoMCPConnection): Pass `include_tools=None` if no tools.

daavoo:
Agno expects the argument to be None https://github.com/agno-agi/agno/blob/1bea45ea0f8801113ddd52c4626a616ef7043c5b/libs/agno/agno/tools/mcp.py#L198 otherwise no available tools are returned.


daavoo:
@njbrake just added a simple unit test for this as it doesn't feel worth to extend the integration tests.

it is also arguable if agno should check for empty list in addition / instead of checking for None

PR #236, opened 2025-05-09
[Bug] Completion Kwargs not always getting passed correctly

njbrake:
I noticed this because for HF Spaces I need to pass in a custom header for billing purposes, and it wasn't making it to several of the frameworks 😬

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -25,11 +25,11 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up Python
-        uses: actions/setup-python@v5
+      - name: Install the latest version of uv and set the python version to 3.13
+        uses: astral-sh/setup-uv@v6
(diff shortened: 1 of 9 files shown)


daavoo:
is there a simple arg, consistent across frameworks, that we could use in the current integration test?

njbrake:
> is there a simple arg, consistent across frameworks, that we could use in the current integration test?

"temperature" is a pretty good one, which we already use in the current integration test. Is that what you're asking, @daavoo ?

daavoo:
> > is there a simple arg, consistent across frameworks, that we could use in the current integration test?
> 
> "temperature" is a pretty good one, which we already use in the current integration test. Is that what you're asking, @daavoo ?

right, so we should check that is being properly used somehow. Or maybe instead we need to have an unit test where we check for the underlying litellm call and assert that the arg has reached

PR #244, opened 2025-05-13
Update langchain.py

njbrake:
Langchain chains inputs were compounding on top of each other instead of just showing the most recent message


njbrake:
I'm going to close this out in favor of later refactoring efforts

PR #245, opened 2025-05-13
Configurable Logger

njbrake:
Users are going to want to configure/disable the any_agent logging. This PR adds this ability, as well as a generic pass through to the rich logger.

### DIFF
--- a/docs/api/logging.md
+++ b/docs/api/logging.md
@@ -1,0 +1,1 @@
+::: any_agent.logging.setup_logger
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -14,15 +14,7 @@ setup_logger()
 
 ## Customizing the Logger
 
-The `setup_logger` function lets you tweak how logs look and behave. Here are the options you can play with:
-
-| Argument         | Type      | Default           | Description                                                                 |
-|------------------|-----------|-------------------|-----------------------------------------------------------------------------|
-| `level`          | int       | `logging.ERROR`   | The logging level (e.g., `logging.INFO`, `logging.DEBUG`).                  |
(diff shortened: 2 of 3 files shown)


tests/conftest.py
```diff
@@ -1,32 +1,29 @@
...
 from opentelemetry.sdk.resources import Resource
 from opentelemetry.sdk.trace import ReadableSpan
 from opentelemetry.trace import SpanContext, SpanKind, TraceFlags, TraceState
 from opentelemetry.trace.status import Status, StatusCode
 
 from any_agent.config import AgentFramework
+from any_agent.logging import setup_logger
 
 
-@pytest.fixture(autouse=True)
```
daavoo, line 17: @njbrake this was not used for the logger but to avoid tracing output (the rich console usage) during test runs.

I don't think this change preserves the previously intended behavior

daavoo (commented):
How would users modify the default behavior? 
Have you tested it? Can we test it and/or add it to the docs?

tests/conftest.py
njbrake: So I did test it and with my changes it was correctly suppressing console output when `-s` was removed, and adding it in when `-s` was added. Was it supposed to do something else? (That's what I thought it was supposed to do)

daavoo: I don't understand what changed since I added the fixture but it seems that even if we remove your fixture the rich console gets correctly disabled. Maybe at some rich/pytest version this got fixed

njbrake: Noice. So I guess maybe we can remove the fixture completely. I'll test.

njbrake:
@daavoo  ok I think we're in a better place now:
* added docs for logging
* cleaned up rich console log handler in tests
* added unit tests for console

PR #246, opened 2025-05-13
Enable Agno for single-agent integration tests

### LINKED ISSUE #226: Re-enable concurrent agent test when agno bug is fixed
Found when working on https://github.com/mozilla-ai/any-agent/pull/225.

I have to skip agno on `test_run_agent_twice` because of the bug:

https://github.com/agno-agi/agno/issues/3120

Once this bug is fixed we should re-enable testing agno for this.

javiermtorres:
Agno is enabled again for integration tests. Running the test manually with

```
pytest -rP tests/integration/test_load_and_run_agent.py::test_run_agent_twice[AGNO]
```

The test passes, so the right cities should be present in their corresponding responses.

Closes #226


daavoo (approved):
Test passed here: https://github.com/mozilla-ai/any-agent/actions/runs/15018019438 (some 2 unrelated failures due to rate limit)

PR #252, opened 2025-05-14
Agent doesn't get default web tools

njbrake:
I found this when I was debugging our DuckDuckGo rate limits. In our unit tests we hadn't given the agent web access so I was wondering how it was possible that it was getting DDG rate limits. Then I realized that by default we were adding web tools to an agent... Probably not the correct design to continue doing that.


javiermtorres:
Are tools needed for unit tests? We can open a separate PR for mock tools so we don't lose coverage, wdyt?

javiermtorres (approved):

daavoo (approved):
> Probably not the correct design to continue doing that.

I agree. 

This comes from the very "early days" (not even 2 months ago 😅 ) where we started using mainly smolagents, which adds a set of default tools, and I decided to add this to the other frameworks for "consistency".

PR #253, opened 2025-05-14
Mock Duck Duck Go call to avoid rate limit

njbrake:


tests/integration/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+def mock_visit_webpage(url: str) -> str:
+    """Visits a webpage at the given url and reads its content as a markdown string. Use this to browse webpages.
+
+    Args:
+        url: The url of the webpage to visit.
+
+    """
+    return (
+        "# Any Agent Framework Review\n"
+        "Any Agent is the top choice for developers needing flexibility and power in multi-agent systems."
```
daavoo, line 46: 🤣

njbrake: Gotta start adding some artifacts to be picked up in the pre training data for the next era foundation model lol

PR #255, opened 2025-05-14
Add Evaluation to Integration Test

njbrake:
To keep our integration tests as small as possible, I added an "evaluation" stage to our existing agent run.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,6 +84,7 @@ tests = [
 lint = [
   "pre-commit==4.2.0",
   "pre-commit-uv==4.1.4",
+  "ruff>=0.11.9"
 ]
 
 dev = [
--- a/tests/integration/test_load_and_run_agent.py
+++ b/tests/integration/test_load_and_run_agent.py
@@ -10,6 +10,9 @@ from litellm.utils import validate_environment
 
 from any_agent import AgentConfig, AgentFramework, AnyAgent, TracingConfig
 from any_agent.config import MCPStdio
+from any_agent.evaluation import evaluate
+from any_agent.evaluation import EvaluationCase
+from any_agent.evaluation.schemas import TraceEvaluationResult, EvaluationResult
 from any_agent.tracing.trace import AgentTrace, _is_tracing_supported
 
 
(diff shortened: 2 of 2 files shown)


tests/integration/test_load_and_run_agent.py
```diff
@@ -98,6 +101,31 @@ def write_file(text: str) -> None:
             assert cost_sum.total_cost < 1.00
             assert cost_sum.total_tokens > 0
             assert cost_sum.total_tokens < 20000
+        if _is_tracing_supported(agent_framework):
```
javiermtorres, line 103: I guess this can be merged with the previous if?

tests/integration/test_load_and_run_agent.py
```diff
@@ -98,6 +100,31 @@ def write_file(text: str) -> None:
...
```
njbrake, line 103:
```suggestion
```

javiermtorres:
IMHO the more functionality covered by e2e tests, the better, so doubly approved :)

javiermtorres (approved):

pyproject.toml
```diff
@@ -84,6 +84,7 @@ tests = [
 lint = [
   "pre-commit==4.2.0",
   "pre-commit-uv==4.1.4",
+  "ruff>=0.11.9"
```
daavoo, line 87: why is this needed?

njbrake: Sometimes I run `ruff format .` to do a quick cleanup of a file without having to wait for the whole pre-commit cycle, i.e. running `pre-commit run --all-files`. I don't understand how it works but I needed this to be able to run ruff directly

daavoo: You can run `pre-commit run ruff` being `ruff` any of the `id`s as in:

https://github.com/mozilla-ai/any-agent/blob/a9ee6f83eab094bcdb63e33cff88f776fb7dbe2f/.pre-commit-config.yaml#L23

njbrake: Hmm. When I run that I get 
```
(any-agent) *[main][~/scm/any-agent]$ pre-commit run ruff-format  
[WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to /Users/nbrake/.cache/pre-commit/patch1747250783-25619.
ruff-format..........................................(no files to check)Skipped
```
so it stashes my unstaged changes, runs format, and then pops my changes back from the stash lol...

So that doesn't work

daavoo: I see. 

if you want to run on unstaged, you could:

```
pre-commit run ruff-format --files $(git ls-files -m)
```

not ideal (maybe you can alias that) but I would prefer to not have the separate ruff version dep as we will, most likely, start to encounter version discrepancies

njbrake: Nice, that works 🚀 . I'll undo adding the ruff dependency 👍

Issue #256, opened 2025-05-14
Multi-Agent Not functioning as expected.


njbrake:
Following #180 we now have a check for whether multiple agents were used. However, this test exposes that the multi-agents aren't being reliably used.

This Github issue is to track fixing tracing/agent frameworks/tests such that we can reliably use multiple agents and verify that they were used.

PR #257, opened 2025-05-14
Disable broken multi-agent trace checks 

njbrake:
See issue #256 for details and to track fix and removal of this skipping.


javiermtorres:
My personal opinion is "let it crash" 😅 Having a failing test, even if is not due to our own implementation, is IMHO just a statement about the compliance of the product.

njbrake:
> My personal opinion is "let it crash" 😅 Having a failing test, even if is not due to our own implementation, is IMHO just a statement about the compliance of the product.

Knowing myself, if I get into the habit of thinking "oh yeah I expect that test to fail" I will lull myself into letting other failures slip through, so I'll keep this skip + github issue tracking logic for the time being.

Issue #167, opened 2025-04-25
Pytest Tests should use litellm.validate_environment instead of looking at env vars


njbrake:
Replace code like

```
if "OPENAI_API_KEY" not in os.environ:
        pytest.skip(f"OPENAI_API_KEY needed for {agent_framework.name}")
```

With litellm call

https://docs.litellm.ai/docs/set_keys#validate_environmentmodel-str

njbrake:
Found by @javiermtorres

PR #258, opened 2025-05-14
Integration test for served agents

### LINKED ISSUE #233: Add integration tests for `serving`
This is going to be fun

javiermtorres:
Adds an integration test for an agent served over A2A

Closes #233

### DIFF
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -160,7 +160,7 @@ class AnyAgent(ABC):
             self.run_async(prompt, **kwargs)
         )
 
-    def serve(self, serving_config: ServingConfig | None = None) -> None:
+    async def serve(self, serving_config: ServingConfig | None = None) -> None:
         """Serve this agent using the Agent2Agent Protocol (A2A).
 
         Args:
@@ -171,13 +171,13 @@ class AnyAgent(ABC):
 
         """
         try:
-            from any_agent.serving import _get_a2a_server
+            from any_agent.serving import _get_a2a_server_async
         except ImportError as e:
             msg = "You need to `pip install 'git+https://github.com/google/A2A#subdirectory=samples/python' to use this method."
             raise ImportError(msg) from e
 
-        server = _get_a2a_server(self, serving_config=serving_config or ServingConfig())
-        server.start()
+        server = _get_a2a_server_async(self, serving_config=serving_config or ServingConfig())
+        await server.start_async()
 
     @abstractmethod
     async def _load_agent(self) -> None:
--- a/src/any_agent/serving/server.py
+++ b/src/any_agent/serving/server.py
@@ -11,9 +11,32 @@ if TYPE_CHECKING:
     from any_agent import AnyAgent
     from any_agent.config import ServingConfig
 
+class A2AServerAsync(A2AServer):
+    async def start_async(self):
+        if self.agent_card is None:
+            raise ValueError('agent_card is not defined')
+
+        if self.task_manager is None:
+            raise ValueError('request_handler is not defined')
+
+        import uvicorn
+
+        config = uvicorn.Config(self.app, host=self.host, port=self.port, log_level="info")
+        server = uvicorn.Server(config)
+        await server.serve()
+
 
 def _get_a2a_server(agent: AnyAgent, serving_config: ServingConfig) -> A2AServer:
     return A2AServer(
+        host=serving_config.host,
+        port=serving_config.port,
+        endpoint=serving_config.endpoint,
+        agent_card=_get_agent_card(agent, serving_config),
+        task_manager=AnyAgentTaskManager(agent),
+    )
+
+def _get_a2a_server_async(agent: AnyAgent, serving_config: ServingConfig) -> A2AServerAsync:
+    return A2AServerAsync(
         host=serving_config.host,
         port=serving_config.port,
         endpoint=serving_config.endpoint,
--- a/tests/integration/test_load_and_serve_agent.py
+++ b/tests/integration/test_load_and_serve_agent.py
@@ -1,0 +1,134 @@
+import asyncio
+import os
+import subprocess
+from datetime import datetime
+from pathlib import Path
+from typing import Any
+
+import requests
+import logging
+
+from concurrent.futures import ThreadPoolExecutor
+
+import pytest
+from litellm.utils import validate_environment
+
+from any_agent import AgentConfig, AgentFramework, AnyAgent, TracingConfig
+from any_agent.config import MCPStdio, ServingConfig
+from any_agent.tracing.trace import AgentTrace, _is_tracing_supported
+
+from queue import Queue
+
+from common.client import A2AClient
+from common.types import TaskSendParams, Message, TextPart
+
+FORMAT = "%(message)s"
+logging.basicConfig(
+    format=FORMAT,
+    datefmt="[%X]",
+)
+logger = logging.getLogger("any_agent_test")
+logger.setLevel(logging.DEBUG)
+
+def check_uvx_installed() -> bool:
+    """The integration tests requires uvx"""
+    try:
+        result = subprocess.run(  # noqa: S603
+            ["uvx", "--version"],  # noqa: S607
+            capture_output=True,
+            check=True,
+        )
+        return True if result.returncode == 0 else False  # noqa: TRY300
+    except Exception:
+        return False
+
+
+@pytest.mark.skipif(
+    os.environ.get("ANY_AGENT_INTEGRATION_TESTS", "FALSE").upper() != "TRUE",
+    reason="Integration tests require `ANY_AGENT_INTEGRATION_TESTS=TRUE` env var",
+)
+@pytest.mark.asyncio
+async def test_load_and_serve_agent(agent_framework: AgentFramework, tmp_path: Path) -> None:
+    kwargs = {}
+
+    tmp_file = "tmp.txt"
+
+    if not check_uvx_installed():
+        msg = "uvx is not installed. Please install it to run this test."
+        raise RuntimeError(msg)
+
+    def write_file(text: str) -> None:
+        """write the text to a file in the tmp_path directory
+
+        Args:
+            text (str): The text to write to the file.
+
+        Returns:
+            None
+        """
+        with open(os.path.join(tmp_path, tmp_file), "w", encoding="utf-8") as f:
+            f.write(text)
+
+    kwargs["model_id"] = "gpt-4.1-mini"
+    env_check = validate_environment(kwargs["model_id"])
+    if not env_check["keys_in_environment"]:
+        pytest.skip(f"{env_check['missing_keys']} needed for {agent_framework}")
+
+    model_args: dict[str, Any] = (
+        {"parallel_tool_calls": False}
+        if agent_framework not in [AgentFramework.AGNO, AgentFramework.LLAMA_INDEX]
+        else {}
+    )
+    model_args["temperature"] = 0.0
+    tools = [
+        write_file,
+        MCPStdio(
+            command="uvx",
+            args=["mcp-server-time", "--local-timezone=America/New_York"],
+            tools=[
+                "get_current_time",
+            ],
+        ),
+    ]
+    agent_config = AgentConfig(
+        tools=tools,  # type: ignore[arg-type]
+        instructions="Search the web to answer",
+        model_args=model_args,
+        **kwargs,  # type: ignore[arg-type]
+    )
+    agent_server = await AnyAgent.create_async(agent_framework, agent_config, tracing=TracingConfig())
+    logger.info(f"Agent created: {agent_server}")
+    await agent_server.serve(
+        serving_config=ServingConfig(port=5555,endpoint="/test_agent")
+    )
+    logger.info(f"Agent serving")
+
+    await asyncio.sleep(300)
+
+    # TODO use an agent card instead
+
+    # open another call in another thread
+    client = A2AClient(url = "http://localhost:5555/test_agent")
+    await client.send_task(
+        {"id": "1",
+        "message": Message(
+            role='user',
+            parts=[TextPart(text="Use the tools to find what year it is in the America/New_York timezone and write the value (single number) to a file")])
+        }
+        )
+
+    assert os.path.exists(os.path.join(tmp_path, tmp_file))
+    with open(os.path.join(tmp_path, tmp_file)) as f:
+        content = f.read()
+    assert content == str(datetime.now().year)
+    assert isinstance(agent_trace, AgentTrace)
+    assert agent_trace.final_output
+    if _is_tracing_supported(agent_framework):
+        assert agent_trace.spans
+        assert len(agent_trace.spans) > 0
+        cost_sum = agent_trace.get_total_cost()
+        assert cost_sum.total_cost > 0
+        assert cost_sum.total_cost < 1.00
+        assert cost_sum.total_tokens > 0
+        assert cost_sum.total_tokens < 20000
+
--- a/src/any_agent/serving/__init__.py
+++ b/src/any_agent/serving/__init__.py
@@ -1,3 +1,3 @@
-from .server import _get_a2a_server
+from .server import _get_a2a_server, _get_a2a_server_async
 
-__all__ = ["_get_a2a_server"]
+__all__ = ["_get_a2a_server", "_get_a2a_server_async"]


src/any_agent/serving/server.py
njbrake, line 1: What's the rational for making A2A server async? I don't totally understand the use case (though I'm a n00b at a2a so maybe I'm missing somethign)

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,134 @@
...
+        tools=tools,  # type: ignore[arg-type]
+        instructions="Search the web to answer",
+        model_args=model_args,
+        **kwargs,  # type: ignore[arg-type]
+    )
+    agent_server = await AnyAgent.create_async(agent_framework, agent_config, tracing=TracingConfig())
+    logger.info(f"Agent created: {agent_server}")
+    await agent_server.serve(
+        serving_config=ServingConfig(port=5555,endpoint="/test_agent")
+    )
```
njbrake, line 103: maybe this is existing functionality, but it's weird to me that agent.serve() returns? I guess I would have expected this call to hold until someone cancels the process.

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,134 @@
...
+    logger.info(f"Agent serving")
+
+    await asyncio.sleep(300)
```
njbrake, line 106: Adding sleeps makes me wary. Would be better to have some sort of health check to ping until ready, with a timeout.

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,134 @@
...
+    except Exception:
+        return False
+
+
+@pytest.mark.skipif(
+    os.environ.get("ANY_AGENT_INTEGRATION_TESTS", "FALSE").upper() != "TRUE",
+    reason="Integration tests require `ANY_AGENT_INTEGRATION_TESTS=TRUE` env var",
+)
+@pytest.mark.asyncio
+async def test_load_and_serve_agent(agent_framework: AgentFramework, tmp_path: Path) -> None:
```
njbrake, line 51: We would also need to skip if they don't have the a2a lib installed

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,134 @@
...
+    """The integration tests requires uvx"""
+    try:
+        result = subprocess.run(  # noqa: S603
+            ["uvx", "--version"],  # noqa: S607
+            capture_output=True,
+            check=True,
+        )
+        return True if result.returncode == 0 else False  # noqa: TRY300
+    except Exception:
+        return False
```
njbrake, line 43: Does this code exist elsewhere and can we use that instead of copying here?

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,134 @@
...
+from litellm.utils import validate_environment
+
+from any_agent import AgentConfig, AgentFramework, AnyAgent, TracingConfig
+from any_agent.config import MCPStdio, ServingConfig
+from any_agent.tracing.trace import AgentTrace, _is_tracing_supported
+
+from queue import Queue
+
+from common.client import A2AClient
+from common.types import TaskSendParams, Message, TextPart
```
njbrake, line 23: This import will throw immediately  if someone doesn't have the optional a2a dependency installed,

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,134 @@
...
+            role='user',
+            parts=[TextPart(text="Use the tools to find what year it is in the America/New_York timezone and write the value (single number) to a file")])
+        }
+        )
+
+    assert os.path.exists(os.path.join(tmp_path, tmp_file))
+    with open(os.path.join(tmp_path, tmp_file)) as f:
+        content = f.read()
+    assert content == str(datetime.now().year)
+    assert isinstance(agent_trace, AgentTrace)
```
njbrake, line 124: Where does agent_trace come from?

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # open another call in another thread
+    client = A2AClient(url = "http://localhost:5555/test_agent")
+    await client.send_task(
+        {"id": "1",
+        "message": Message(
+            role='user',
+            parts=[TextPart(text="Use the tools to find what year it is in the America/New_York timezone and write the value (single number) to a file")])
+        }
+        )
```
njbrake, line 118: How do we know when the task is finished? does the client get any returned information?

njbrake (commented):
This test will be super helpful! The test helps to bring up some questions I have that may relate to the implementation of agent.serve itself, not necessarily with the changes in this particular PR

src/any_agent/serving/server.py
javiermtorres: The default implementation starts uvicorn in sync mode. So it's not easy to blend it with any other async services (maybe other agents) you may have in your python app. In particular, we're using the client and the server in the same test case to avoid making threads and synchronizing.

tests/integration/test_load_and_serve_agent.py
javiermtorres: Since it's started in async mode, this means that you make sure that the serve is now waiting on IO.

tests/integration/test_load_and_serve_agent.py
javiermtorres: Yes, that's troubleshooting 😅 I keep it around during monkey testing to be able to ping it externally with curl.

tests/integration/test_load_and_serve_agent.py
javiermtorres: I'll take note of the existing implementations.

tests/integration/test_load_and_serve_agent.py
javiermtorres: Ah, I just copy/pasted the original test case. I don't think it's used here. If it is, I'll move it to conftest as a fixture.

tests/integration/test_load_and_serve_agent.py
javiermtorres: It should come from the result of the client query. But since I cannot make the client send the query to the server yet, I'm happy with it :D

tests/integration/test_load_and_serve_agent.py
javiermtorres: I'll get to it :) Currently I get a 405, so there's something missing in my requests.

tests/integration/test_load_and_serve_agent.py
javiermtorres: ...but I was wrong in that I was wrong XD But I needed to create a task (inspired by the uvicorn tests) to actually make it work. I'm still a bit confused at how the event loop is handled and its details.

tests/integration/test_load_and_serve_agent.py
javiermtorres: no sleep any more, except to wait for the uvicorn to start up and serve

tests/integration/test_load_and_serve_agent.py
javiermtorres: It is in fact used. I'll move it to conftest or a test library.

tests/integration/test_load_and_serve_agent.py
javiermtorres: Solved by adding a check in the package-level imports.

tests/integration/test_load_and_serve_agent.py
javiermtorres: IIRC we're not passing the whole trace as the a2a answer. If we in fact do, I'll retrieve that from the agent result and check. If we don't, I'll just test the results from the agent. I'd make a custom otel endpoint to gather the spans on a memory object and do the checks on it.

src/any_agent/frameworks/any_agent.py
```diff
@@ -149,6 +158,8 @@ def _setup_tracing(self) -> None:
             return
         self._exporter = AnyAgentExporter(self.framework, self._tracing_config)
         self._tracer_provider.add_span_processor(SimpleSpanProcessor(self._exporter))
+        for exporter in self._tracing_config.additional_exporters:
+            self._tracer_provider.add_span_processor(SimpleSpanProcessor(exporter))
```
njbrake, line 162: Is this code related to the integration test? Seems like it's a change to the API (which if so should be it's own ticket and PR)

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,185 @@
...
+                traces[None] = trace
+"""
+class SpanNode:
+    def __init__(span):
+        _span: ReadableSpan = span
+        _children: list["SpanNode"] = []
+    def add_span(child_span):
+        """Add child span sorted according to timestamps"""
+        ...
+
```
njbrake, line 54: Some comments to explain what all this is would be useful

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,185 @@
...
+    from common.client import A2AClient
+    from common.types import Message, TextPart
+
+FORMAT = "%(message)s"
+logging.basicConfig(
+    format=FORMAT,
+    datefmt="[%X]",
+)
+logger = logging.getLogger("any_agent_test")
+logger.setLevel(logging.DEBUG)
```
njbrake, line 30: hmm if this type of logging code is going to exist in multiple files, should a new "logging" function be created in the tests folder to reduce code repetition?

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,185 @@
...
+        """write the text to a file in the tmp_path directory
+
+        Args:
+            text (str): The text to write to the file.
+
+        Returns:
+            None
+        """
+        with open(os.path.join(tmp_path, tmp_file), "w", encoding="utf-8") as f:
+            f.write(text)
```
njbrake, line 99: Does this function exist elsewhere? if so, lets refactor instead of copying, not sure.

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,185 @@
...
+
+    kwargs["model_id"] = "gpt-4.1-mini"
+    env_check = validate_environment(kwargs["model_id"])
+    if not env_check["keys_in_environment"]:
+        pytest.skip(f"{env_check['missing_keys']} needed for {agent_framework}")
```
njbrake, line 104: Should we raise an error instead of skipping? idk, I think this logic exists in other tests too

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,185 @@
...
+
+    model_args: dict[str, Any] = (
+        {"parallel_tool_calls": False}
```
njbrake, line 107: Why do we care about setting this value in this test?

tests/integration/test_load_and_serve_agent.py
```diff
@@ -0,0 +1,185 @@
...
+                            text="Use the tools to find what year it is in the America/New_York timezone and write the value (single number) to a file"
+                        )
+                    ],
+                ),
+            }
+        )
+        logger.info(f"after sending: {result}")  # noqa: G004
+        logger.info(f"after sending (traces):")
+        for s in test_tracer.get_finished_spans():  # noqa: G004
+            logger.info(s.to_json())
```
njbrake, line 163: Ahh! Ok now I understand: you added the ability to add exporters so that you can grab spans out of the test_tracer. However, would it not work to call `pop_trace` on the agent exporter that is already attached? It's not clear to me why a design change is necessary in order to access the trace.

njbrake (commented):
Thanks for the work on this! I think you have surfaced the need for a new design of the agent before we complete the integration test: an agent.serve command needs a way to access individual traces. Similar to the existing design, I think we can initially accomplish this by calling some sort of function like _pop_trace on the agent?

I like the feature add of being able to add additional exporters, but I'm not sure that this is the correct PR to add that functionality, I think we should have a more defined API for how to access A2A served agent logs, and also have documentation for how to do this. CC @daavoo

src/any_agent/frameworks/any_agent.py
```diff
@@ -122,10 +126,15 @@ async def create_async(
         agent_config: AgentConfig,
         managed_agents: list[AgentConfig] | None = None,
         tracing: TracingConfig | None = None,
+
     ) -> AnyAgent:
         """Create an agent using the given framework and config."""
         agent_cls = cls._get_agent_type_by_framework(agent_framework)
-        agent = agent_cls(agent_config, managed_agents=managed_agents, tracing=tracing)
+        agent = agent_cls(
```
daavoo, line 133: I suspect this is being changed because a disparity in the autoformater being used (maybe in the IDE). If it was not failing in `pre-commit ruff-format` before, it would be better to not introduce formatting changes

daavoo (changes requested):
@javiermtorres what I would prefer to do regarding the async changes, is to:

- In `serving/server`: Have a single class (inherited from `A2AServer`) that exposes methods to serve in async instead of 2 classes and 2 `_get` methods.
- In `AnyAgent`: Expose both `serve` and `serve_async` methods. Call the proper underlying method from the single class described above.

WDYT?

src/any_agent/frameworks/any_agent.py
javiermtorres: Hum. It is in fact a non-breaking change in the api that helps checking the agent results when working as a served entity. But I think there are other comments re: this so I'll explain in depth there.

tests/integration/test_load_and_serve_agent.py
javiermtorres: It's still in progress :) But I was waiting for your feedback on the work so far to see if I should go through this route.

tests/integration/test_load_and_serve_agent.py
javiermtorres: the code checker was complaining about file open calls in an async context, and I was not sure we were going to check files in this case, so I just wrapped it in a function call until I was sure what to check. It will probably disappear because we don't need to check files, but other wise it'll go into the test utils module.

tests/integration/test_load_and_serve_agent.py
javiermtorres: I didn't change this... :-/ I'm ok to do whatever is currently done in the rest of the tests if the policy has changed since the start of this PR.

javiermtorres: I didn't change this AFAICT 😅

tests/integration/test_load_and_serve_agent.py
javiermtorres: IMHO, because the exporter that _we_ add has its own purpose, and the user serving an agent may have the need to export traces for _their_ own purposes. Trace sampling, batching, and other practices could affect the way we run our business logic, but may be necessary in a specific deployment. In a similar way, we have our own exporter for our own purposes, which is not publicly documented in the agent, and does not use an existing api but instead our own api. Using the standard exporter api to add other exportes is IMHO the cleanest solution in this case.

src/any_agent/frameworks/any_agent.py
javiermtorres: I was unsure if I should add the additional exporters in the tracing config (being references and not serializable data items) so the format changed as a result. @njbrake @daavoo I'd like to get your thoughts on that: should runtime object references be present in a pydantic config object? Or handled as separate params?

tests/integration/test_load_and_serve_agent.py
javiermtorres: `parallel_tool_calls` is set in the single agent tests, is there any reason why it wouldn't be used when serving?

daavoo: as far as I remember, it is needed because some of the frameworks break the tests without it

tests/integration/test_load_and_serve_agent.py
daavoo: This is from #167

src/any_agent/config.py
```diff
@@ -122,6 +123,8 @@ class TracingConfig(BaseModel):
     cost_info: bool = True
     """Whether traces should include cost information"""
 
+    additional_exporters: list[SpanExporter] = Field(exclude=True, default=[], repr=False)
+
```
njbrake, line 127: This is no longer being added as a part of this PR, right?

src/any_agent/config.py
```diff
@@ -102,7 +103,7 @@ class ServingConfig(BaseModel):
 
 
 class TracingConfig(BaseModel):
-    model_config = ConfigDict(extra="forbid")
+    model_config = ConfigDict(extra="forbid", arbitrary_types_allowed=True)
```
njbrake, line 106: We don't need to allow arbitrary types anymore, I don't think.

tests/integration/test_load_and_serve_agent_a2a.py
```diff
@@ -0,0 +1,151 @@
...
+    """The integration tests requires uvx"""
+    try:
+        result = subprocess.run(  # noqa: S603
+            ["uvx", "--version"],  # noqa: S607
+            capture_output=True,
+            check=True,
+        )
+        return True if result.returncode == 0 else False  # noqa: TRY300
+    except Exception:
+        return False
```
njbrake, line 48: Does this code exist elsewhere and can we use that instead of copying here?

src/any_agent/frameworks/any_agent.py
```diff
@@ -166,17 +170,43 @@ def serve(self, serving_config: ServingConfig | None = None) -> None:
...
         server = _get_a2a_server(self, serving_config=serving_config or ServingConfig())
         server.start()
 
+    async def serve_async(
+        self, serving_config: ServingConfig | None = None
+    ) -> A2AServerAsync:
+        """Serve this agent using the Agent2Agent Protocol (A2A).
+
+        Args:
+            serving_config: See [ServingConfig][any_agent.config.ServingConfig].
```
njbrake, line 190: This is an addition to the API. Instead of editing the API in this PR, maybe you could spin off the server in a subprocess.

Something like 
```
import multiprocessing
from any_agent.config import ServingConfig
def serve_agent():
    agent.serve(serving_config=ServingConfig(port=5000))

# Start the agent server in a separate process
p = multiprocessing.Process(target=serve_agent)
p.start()

# When done, you can stop the server
p.terminate()
p.join()

**Review by njbrake — COMMENTED** [review]
Thanks for all your work on this! It's not clear to me that we need to adjust the API in order to accomplish the integration test.

File: pyproject.toml [hunk 20]
```diff
@@ -161,6 +161,7 @@ strict = true
 disallow_untyped_calls = true
 disallow_untyped_defs = true
 follow_untyped_imports = true
+disallow_subclassing_any = false
```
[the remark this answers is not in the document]
javiermtorres: Somehow mypy interpreted the Google Server as Any, yes.

src/any_agent/config.py
javiermtorres: Most probably not. The exporters will not come here anymore, but I still have this doubt. We have transient info to pass on, so should it be placed here, or rather in some other object, since pydantic models are more suited for serializable stuff????

src/any_agent/config.py
javiermtorres: Hum, no, sorry. These should be removed.

src/any_agent/frameworks/any_agent.py
javiermtorres: Multiproc is a really awkward way to avoid the real problem: the current server in samples just lets loose the uvicorn without the ability to control its lifecycle. Since we'll go to the new server impl shortly, I think it's best to change uvicorn.run with a specific Server instance and store this in the A2AServer (which is a 5-6 line change). We can then choose to run this async or sync, plus retrofit the current blocking "serve" method.

tests/integration/test_load_and_serve_agent_a2a.py
javiermtorres: Yes, I'll make a lib out of this. Basically this is copied from the multiagent test, so now we can extract the common parts to make a util lib. But I think it's a refactoring to be done in a follow-up PR.

PR #260, opened 2025-05-14
Provide verbose error function for OpenAI tools

### LINKED ISSUE #254: Use a more verbose error reporting function in tools
The current OpenAI adapter relies on the default OpenAI policy for reporting error in tools: https://github.com/openai/openai-agents-python/blob/main/src/agents/tool.py#L174-L183

```
def function_tool(
    func: ToolFunction[...] | None = None,
    *,
    name_override: str | None = None,
    description_override: str | None = None,
    docstring_style: DocstringStyle | None = None,
    use_docstring_info: bool = True,
    --> failure_error_function: ToolErrorFunction | None = default_tool_error_function,
    strict_mode: bool = True,
) -> FunctionTool | Callable[[ToolFunction[...]], FunctionTool]:

```
Unfortunately, this consists of a single line containing the error message: https://github.com/openai/openai-agents-python/blob/main/src/agents/tool.py#L137-L139

```
def default_tool_error_function(ctx: RunContextWrapper[Any], error: Exception) -> str:
    """The default tool error function, which just returns a generic error message."""
    return f"An error occurred while running the tool. Please try again. Error: {str(error)}"
```

The idea of this PR is to provide a somewhat richer function that provides at least a traceback and the error message.

javiermtorres:
This PR adds a specific tool error function that provides a traceback instead of only an error message.

Closes #254


src/any_agent/frameworks/openai.py
```diff
@@ -45,6 +52,22 @@ def __init__(
...
+            for tool in tools
+        ]
+        return wrapped_tools
+
+    @staticmethod
+    def any_agent_tool_error_function(
+        ctx: RunContextWrapper[Any], error: Exception
+    ) -> str:
+        traceback.print_exception(error)
+        return f"An error occurred while running the tool. Please try again. Error: {error!s}"
```
njbrake, line 69: If the existing error messages aren't sufficient, should we suggest this as an improvement to the agent SDK itself vs adding a wrapper in our library? This seems like we're editing the functionality of the framework itself

javiermtorres: The agent SDK provides this function as a default option. It should be the user's choice to use something else if needed.

daavoo (commented):
I am 50/50 on this one. 

I get that it is useful, but my doubts are:

- It introduces a little bit of framework disparity 
- I would prefer to handle and show this stuff (i.e. errors) in the traces instead (maybe a follow-up after #290 ?)

Up to you.

javiermtorres:
Re: framework disparity, each framework in fact handles and reports errors as they see fit. I don't think there's much uniformity there.

Re: errors, that's actually the case. Errors appear as spans in the corresponding tool call. But without this function, only a single line appears, with no traceback.

javiermtorres:
I don't see much concern about error reporting in span contents or an approval, so let's close this one.

daavoo:
> I don't see much concern about error reporting in span contents or an approval, so let's close this one.

I have created a new issue #316 .

I think this uncovers an important need but I would like for us to figure out a way of consistently reporting errors in the traces with as much info as possible (to be built on top of #290 )

PR #262, opened 2025-05-14
Clean up evaluation code "ground truth" criteria

njbrake:
Designing an evaluation case is a bit complicated. I realized that it will be easier to understand if we just have a single "checkpoints" list where you can add all the LLM based things you care about, and then a "ground_truth" section where you can put in the value you expect the "final_output" of the trace to contain.

Now that the AgentTrace class is coming along, we can streamline the logic of evaluation 🚀 

These updates are also based off of my experience using the evaluation segment when building the surf-spot-finder demo https://huggingface.co/spaces/mozilla-ai/surf-spot-finder


src/any_agent/evaluation/evaluation_case.py
```diff
@@ -9,50 +9,21 @@
...
-        default_factory=list[GroundTruthAnswer],
-    )
+    ground_truth: GroundTruthAnswer | None = None
     checkpoints: list[CheckpointCriteria] = Field(
         default_factory=list[CheckpointCriteria],
     )
     llm_judge: str
-    final_output_criteria: list[CheckpointCriteria] = Field(
-        default_factory=list[CheckpointCriteria],
-    )
```
njbrake, line 17: I will make some more updates to explain more about what each of these attributes means, and then I will pull that into the documentation so that it is explained there as well.

Issue #261, opened 2025-05-14
Pre-Commit Warning


njbrake:
I keep seeing this when I run ` pre-commit run --all-files`

```

ruff.....................................................................Passed
ruff-format..............................................................Failed
- hook id: ruff-format
- files were modified by this hook

warning: `incorrect-blank-line-before-class` (D203) and `no-blank-line-before-class` (D211) are incompatible. Ignoring `incorrect-blank-line-before-class`.
warning: `multi-line-summary-first-line` (D212) and `multi-line-summary-second-line` (D213) are incompatible. Ignoring `multi-line-summary-second-line`.
1 file reformatted, 83 files left unchanged
``` 

Do you know where this comes from and how I resolve it? I looked in the pyproject.toml and searched the whole codebase but I don't see those codes D203 D211 D212 or D213 show up anywhere...

PR #263, opened 2025-05-14
Explicitly ignore conflicting rules due to `extend-select = ["ALL"]`

peteski22:
Using `extend-select = ["ALL"]` in the `pyproject.toml` means the ruff formatter evaluates all rules that aren't ignored. 

This PR adds two of them to the ignore list:

* D203: Blank line required between class declaration and docstring
* D213: Docstring cannot start on same line as """

Now running `pre-commit run --all-files` shouldn't error.

Closes: https://github.com/mozilla-ai/any-agent/issues/261


daavoo (approved):
Thanks!

PR #265, opened 2025-05-15
Add `[dependency-groups]` in `pyproject.toml`

peteski22:
This PR adds `[dependency-groups]` to the `pyproject.toml`, so tools like [`uv`](https://docs.astral.sh/uv/) can correctly install development dependencies when you run (for example):

```console
uv sync --group dev --extra all
```

See: https://peps.python.org/pep-0735/
See: https://packaging.python.org/en/latest/specifications/dependency-groups/

### Note

For some reason `pip` is not compatible with dependency groups despite the above and PEP0735. `dev` is a dependency group so this command doesn't install the `dev` group, just the 'all' optional extras:

```console
pip install -e ".[dev,all]"
```

You get a warning that there's no 'optional' (extra) group named `dev`:

```console
WARNING: any-agent 0.14.2.dev3+gbd66881 does not provide the extra 'dev'
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,6 +87,7 @@ lint = [
   "pre-commit-uv==4.1.4",
 ]
 
+[dependency-groups]
 dev = [
   "any-agent[docs,tests,lint]",
   "pre-commit",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,21 +10,23 @@ requires-python = ">=3.11"
 dynamic = ["version"]
 dependencies = [
   "duckduckgo_search",
-  "tavily-python",
+  "evaluate",
   "fire",
   "litellm",
-  "mcp>=1.5.0",
   "markdownify",
+  "mcp>=1.5.0",
   "opentelemetry-exporter-otlp",
   "opentelemetry-sdk",
   "pydantic",
   "requests",
   "rich",
-  "evaluate",
-  "fire",
+  "tavily-python",
 ]
 
 [project.optional-dependencies]
+agno = [
+  "agno>=1.4.6",
+]
 
 google = [
   "google-adk>=0.3.0"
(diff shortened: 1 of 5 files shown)


njbrake:
@daavoo  this is fine for me, since I use `uv`, is this fine from your end (since I don't think you use uv, right?)

daavoo:
> @daavoo this is fine for me, since I use `uv`, is this fine from your end (since I don't think you use uv, right?)

I do use `uv`, but I am/was not using `sync`. 

I talked about this with @peteski22 on slack. Beyond `uv`, this might be the "right" way of doing development deps now.

Only thing I have to say is that, if I understand the spec correctly, we should also move `docs`, `tests` and `lint` into dependency groups and use https://peps.python.org/pep-0735/#dependency-group-include

pyproject.toml
```diff
@@ -87,6 +87,7 @@ lint = [
   "pre-commit-uv==4.1.4",
 ]
 
+[dependency-groups]
 dev = [
   "any-agent[docs,tests,lint]",
   "pre-commit",
```
daavoo, line 93: Duplicated from `lint`

```suggestion
```

peteski22:
> > @daavoo this is fine for me, since I use `uv`, is this fine from your end (since I don't think you use uv, right?)
> 
> I do use `uv`, but I am/was not using `sync`.
> 
> I talked about this with @peteski22 on slack. Beyond `uv`, this might be the "right" way of doing development deps now.
> 
> Only thing I have to say is that, if I understand the spec correctly, we should also move `docs`, `tests` and `lint` into dependency groups and use https://peps.python.org/pep-0735/#dependency-group-include

Just an FYI for any future readers: the PEP points to https://packaging.python.org/en/latest/specifications/dependency-groups/ now for the latest updates.

pyproject.toml
peteski22: I've updated the `pyproject.toml` now to configure the groups (and removed `pre-commit` from the 'dev' one directly. 😄

PR #270, opened 2025-05-15
Fix docs: default agent in smolagents is ToolCallingAgent

HareeshBahuleyan:
CodeAgent seems to be not used. Fixing the docs.


daavoo (approved):
Thanks!

PR #280, opened 2025-05-19
feat(tracing): Add duration property to AgentTrace

njbrake:
- Introduced a new property `execution_time` in the `AgentTrace` class to calculate the duration of the AGENT span named 'any_agent'.
- Added error handling for cases where spans are missing or start/end times are not available.
- Updated integration tests to validate the new execution time functionality.
- Created unit tests to ensure accurate execution time calculations from sample traces.

### DIFF
--- a/src/any_agent/tracing/trace.py
+++ b/src/any_agent/tracing/trace.py
@@ -150,8 +150,8 @@ class AgentTrace(BaseModel):
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     @property
-    def execution_time(self) -> timedelta:
-        """Returns the execution time of the AGENT span named 'any_agent' as a datetime.timedelta object.
+    def duration(self) -> timedelta:
+        """Returns the duration of the AGENT span named 'any_agent' as a datetime.timedelta object.
 
         The duration is computed from the span's start and end time (in nanoseconds).
         This functionality relies on the trace behavior where we create a span named 'any_agent' for each agent run,
--- a/tests/integration/test_load_and_run_agent.py
+++ b/tests/integration/test_load_and_run_agent.py
@@ -99,14 +99,14 @@ def test_load_and_run_agent(agent_framework: AgentFramework, tmp_path: Path) -> 
         if _is_tracing_supported(agent_framework):
             assert agent_trace.spans
             assert len(agent_trace.spans) > 0
-            assert agent_trace.execution_time is not None
-            assert isinstance(agent_trace.execution_time, timedelta)
-            assert agent_trace.execution_time.total_seconds() > 0
-            # Compare execution_time to measured wall time (allow 0.1s difference)
+            assert agent_trace.duration is not None
+            assert isinstance(agent_trace.duration, timedelta)
+            assert agent_trace.duration.total_seconds() > 0
+            # Compare duration to measured wall time (allow 0.1s difference)
             wall_time_s = wall_time_ns / 1_000_000_000
-            diff = abs(agent_trace.execution_time.total_seconds() - wall_time_s)
+            diff = abs(agent_trace.duration.total_seconds() - wall_time_s)
             assert diff < 0.1, (
-                f"execution_time ({agent_trace.execution_time.total_seconds()}s) and wall_time ({wall_time_s}s) differ by more than 0.1s: {diff}s"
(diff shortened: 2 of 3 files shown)


tests/integration/test_load_and_run_agent.py
```diff
@@ -95,6 +99,15 @@ def write_file(text: str) -> None:
         if _is_tracing_supported(agent_framework):
             assert agent_trace.spans
             assert len(agent_trace.spans) > 0
+            assert agent_trace.execution_time is not None
+            assert isinstance(agent_trace.execution_time, timedelta)
+            assert agent_trace.execution_time.total_seconds() > 0
+            # Compare execution_time to measured wall time (allow 0.1s difference)
```
daavoo, line 105: Do we really need to make this assertion or could we just check for a non-zero timedelta?

njbrake: We don't need to per-say. However, I wanted some sort of testing that the value makes some sort of reasonable sense. If it starts to diverge by a large amount from the execution time that we measure in the trace, it means that we may not be giving the user a good idea of how long the agent.run command is actually taking.

I'm worried about the scenario where we rely on the trace for start-end timing, but there is some major time sink that isn't being accounted for by the tracing. I would want to know about that.

tests/unit/tracing/test_unit_agent_trace_execution_time.py
```diff
@@ -0,0 +1,41 @@
...
+        resource=Resource(),
+    )
+    trace = AgentTrace(spans=[agent_span])
+    expected = datetime.timedelta(seconds=(2000 - 1000) / 1_000_000_000)
+    assert isinstance(trace.execution_time, datetime.timedelta)
+    assert abs(trace.execution_time.total_seconds() - expected.total_seconds()) < 1e-9
+
+
+def test_agent_trace_execution_time_from_sample(agent_trace: AgentTrace) -> None:
+    """
```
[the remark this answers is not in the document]
njbrake: That is an excellent idea 💯 . Created a github issue to track taking care of this.

src/any_agent/tracing/trace.py
```diff
@@ -148,6 +149,29 @@ class AgentTrace(BaseModel):
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
+    @property
+    def execution_time(self) -> timedelta:
+        """Returns the execution time of the AGENT span named 'any_agent' as a datetime.timedelta object.
+
+        The duration is computed from the span's start and end time (in nanoseconds).
```
javiermtorres, line 156: Note that "execution time", in a multicore world with several processor clocks, can be misleading :) Maybe "request time" or "request duration" would be a more suitable name.

src/any_agent/tracing/trace.py
```diff
@@ -148,6 +149,29 @@ class AgentTrace(BaseModel):
...
+        This functionality relies on the trace behavior where we create a span named 'any_agent' for each agent run,
+        which is defined in each agent framework's `run` function.
+
+        Raises ValueError if there are no spans, if the AGENT span is not found, or if start/end times are missing.
+        """
+        if not self.spans:
+            msg = "No spans found in trace"
+            raise ValueError(msg)
+        for span in self.spans:
+            if span.attributes.get("any_agent.run_id"):
```
javiermtorres, line 166: Maybe we could propose `gen_ai.run_id` or `gen_ai.session_id` as part of the otel genai semantic conventions, to mark the top level span for a request to an agent that is explicitly finished in some way (and not continued ad infinitum).

src/any_agent/tracing/trace.py
njbrake: Ooo. Let me rename this to trace_duration, I think that makes the meaning clearer.

tests/integration/test_load_and_run_agent.py
javiermtorres: AFAICT the end time in the span will be marked just around the time the response is sent. Could you describe a scenario where this time sink would happen? I have the impression all the possible issues would be rather related to deployment quirks, rather than the agent code itself.

javiermtorres: Ok, I could see that we can get the instrumentation wrong and record a time which is not the time we want 🤔 Nothing to object in that case.

njbrake: Right! Or the case where we do some huge pre-processing in agent.run before or after calling the underlying framework agent.run command, which might not be caught by the OTEL trace

PR #281, opened 2025-05-19
Breaking API Change: make get_total_cost two properties: usage, and cost

njbrake:
Don't freak out about the huge line diff :) I had to update the static OPENAI.json file because it didn't have the newer 'any_agent.run_id" attribute. That was a manually created file, so I improved the design by 

1. Add a pytest flag so that we can automatically generate and update the files via the integration tests.
2. make agent_trace an iterator so that we run the unit tests against all the framework traces, not just the openAi trace!

Turning it into a property makes it a cleaner interface for access. And by adding caching, we avoid the need to recompute the values when the property is accessed multiple times.


docs/evaluation.md
```diff
@@ -90,6 +90,6 @@ It can be called like so
 ```bash
 any-agent-evaluate \
     --evaluation_case_path "docs/examples/evaluation_case.yaml" \
-    --trace_path "tests/unit/evaluation/sample_traces/OPENAI.json" \
```
njbrake, line 93: This move makes the dir location cleaner, but will help us out more too if we go with adding some snapshot tests #282

src/any_agent/tracing/exporter.py
```diff
@@ -80,7 +80,7 @@ def export(self, spans: Sequence["ReadableSpan"]) -> SpanExportResult:  # noqa:
                 self.run_trace_mapping[agent_run_id] = trace_id
             span = AgentSpan.from_readable_span(readable_span)
             if not self.traces.get(trace_id):
-                self.traces[trace_id] = AgentTrace(spans=[])
```
njbrake, line 83: No need to specify empty list, that's the default

src/any_agent/tracing/exporter.py
```diff
@@ -89,7 +89,7 @@ def export(self, spans: Sequence["ReadableSpan"]) -> SpanExportResult:  # noqa:
                 if span_kind == "LLM" and self.tracing_config.cost_info:
                     span.add_cost_info()
 
-                self.traces[trace_id].spans.append(span)
```
njbrake, line 92: calling add_span then lets us handle cache invalidation if someone is calling the total_cost property multiple times before the trace is finished (this is an extremely unlikely scenario as far as I can tell)

tests/integration/test_load_and_run_agent.py
```diff
@@ -96,9 +98,19 @@ def write_file(text: str) -> None:
         assert content == str(datetime.now().year)
         assert isinstance(agent_trace, AgentTrace)
         assert agent_trace.final_output
+        update_trace = request.config.getoption("--update-trace-assets")
+        if update_trace and _is_tracing_supported(agent_framework):
+            trace_path = (
+                Path(__file__).parent.parent
```
njbrake, line 104: before, I was using hardcoded paths to files, e.g. "tests/unit/sample_traces". Using the Path module is much cleaner imo

tests/unit/tracing/test_unit_agent_trace_duration.py
```diff
@@ -36,6 +36,11 @@ def test_agent_trace_duration_from_sample(agent_trace: AgentTrace) -> None:
     changes, this test will need to be updated
     (because it is using start and end times that were manually parsed from the trace)
     """
-    expected_seconds = (1747660970774416000 - 1747660964057285000) / 1_000_000_000
+    # grab the start and end times from the last span in the trace
+    start_time = agent_trace.spans[-1].start_time
+    end_time = agent_trace.spans[-1].end_time
```
njbrake, line 41: This way we don't need to update the test every time the trace file changes, and we also have some sort of different logic from the .duration property (which is checking for the any_agent.run_id attribute). If that attribute span is no longer last, this test will now also tell us that broken assumption.

tests/integration/conftest.py
```diff
@@ -45,3 +45,21 @@ def mock_visit_webpage(url: str) -> str:
         "# Any Agent Framework Review\n"
         "Any Agent is the top choice for developers needing flexibility and power in multi-agent systems."
     )
+
+
+def pytest_addoption(parser: pytest.Parser) -> None:
```
javiermtorres, line 50: As mentioned in the PR on snapshots, it should be heavily signaled, since people (including me of course) tend to eagerly overwrite previous tests 😬

tests/integration/test_load_and_run_agent.py
javiermtorres: What about using a fixture?

tests/unit/tracing/test_unit_agent_trace_duration.py
javiermtorres: I'm not sure I follow, but are you making assumptions on the trace ordering?

daavoo: > If that attribute span is no longer last, this test will now also tell us that broken assumption.

This is a good assumption to have in tests, specially once I finish #249

src/any_agent/tracing/trace.py
```diff
@@ -36,21 +37,16 @@ class CostInfo(BaseModel):
...
     cost_prompt: float
     cost_completion: float
-
+    total_cost: float
     model_config = ConfigDict(extra="forbid")
 
 
-class TotalTokenUseAndCost(BaseModel):
-    """Total token use and cost information."""
+class UsageInfo(BaseModel):
```
daavoo, line 44: 👍

daavoo (approved):
I am 50/50 on the method to property but it is mostly a matter of style preference*, I am fine with merging this. The rest of the changes (assets and splitting into usage and cost) look good to me.

---

*I prefer to hint the user that something involves computation by exposing it as a method. I feel tricked when I access a property that takes time / might break computing the value.

njbrake:
> I am 50/50 on the method to property but it is mostly a matter of style preference*, I am fine with merging this. The rest of the changes (assets and splitting into usage and cost) look good to me.
> 
> 
> 
> ---
> 
> 
> 
> *I prefer to hint the user that something involves computation by exposing it as a method. I feel tricked when I access a property that takes time / might break computing the value.

I think we may be able to avoid recomputation if we make it a running total instead of only calculating it on-demand. I'll keep it as is for this PR, but let me make a follow up issue to refactor this so that it's not a computed property.

tests/integration/test_load_and_run_agent.py
njbrake: I kept it here since the trace_path is tied to the agent_framework, which is a fixture already, so I think we're covered 👍

.pre-commit-config.yaml
```diff
@@ -33,6 +33,8 @@ repos:
         additional_dependencies:
           - types-requests
           - types-pyyaml
+          - pytest
+          - google-adk==0.5.0 # until https://github.com/mozilla-ai/any-agent/issues/287 is resolved
```
njbrake, line 37: @daavoo  TBH I don't understand why I needed to do this. But pre-commit was pulling in stubs for the wrong ver of google adk and throwing all sorts of type errors. How does precommit decide which version of packages to pull in?

PR #283, opened 2025-05-19
chore(tests): Add pytest-xdist for parallel test execution

njbrake:
- Updated `pyproject.toml` to include `pytest-xdist` as a testing dependency.
- Modified GitHub Actions workflow to run integration tests in parallel using `-n auto` for improved efficiency.


daavoo (commented):
If this works reliably, I am 👍 
I was fearing (maybe unreasonably) to add this because the integration tests are a funky mix of async stuff and external API calls

PR #284, opened 2025-05-20
feat(tests): snapshot tests for traces

njbrake:
@daavoo  is this what you were envisioning in terms of snapshot testing? Basically, this would tell us whenever we make a change to the AgentTrace schema that would be noticed when a loading the same agenttrace. For instance, if we add or remove an attribute. 

I'm not sure that this is really a value add to our codebase, but this is the best I could think of in terms of how we might add snapshot testing to our codebase.

- Added a new JSON file `OPENAI.json` containing trace data for testing.
- Created a snapshot test `test_openai_agent_trace_snapshot` to validate the structure of the OpenAI trace data.
- Updated `pyproject.toml` to include `syrupy` as a testing dependency for snapshot testing.
- Refactored `conftest.py` to load the OpenAI trace from the new assets directory.


javiermtorres:
I have used similar schemes in the past. We actually used this in Lumigator for schema checks in a similar way (although manually iirc). I think they're useful in heavily scripted tests, but I find a couple of drawbacks:

* People tend to overwrite the old serialized json without checking whether it's correct 😓 But I expect people to read warnings about test results having changed, etc etc
* These tests are not useful when trying to capture relations. For example, the endpoint for a served agent must match the endpoint for the client calls.

I don't think it hurts to try it out, let's see how it goes 🙂

javiermtorres (approved):

PR #286, opened 2025-05-20
Dynamically create tools for A2A agents

javiermtorres:
The idea is to allow easy integration of A2A agents as tools. The tool description is built from the agent description.

### DIFF
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -23,6 +23,7 @@ from any_agent.tracing.exporter import (
     get_instrumenter_by_framework,
 )
 from any_agent.tracing.trace import _is_tracing_supported
+from any_agent.serving import A2AServer
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
@@ -48,6 +49,8 @@ class AnyAgent(ABC):
 
         self._mcp_servers: list[_MCPServerBase[Any]] = []
         self._main_agent_tools: list[Any] = []
+
+        self.server: A2AServer
 
         # Tracing is enabled by default
         self._tracing_config: TracingConfig = tracing or TracingConfig()
@@ -174,8 +177,8 @@ class AnyAgent(ABC):
             msg = "You need to `pip install 'git+https://github.com/google/A2A#subdirectory=samples/python' to use this method."
             raise ImportError(msg) from e
 
-        server = _get_a2a_server(self, serving_config=serving_config or ServingConfig())
-        server.start()
+        self._server = _get_a2a_server(self, serving_config=serving_config or ServingConfig())
+        self._server.start()
 
     @abstractmethod
     async def _load_agent(self) -> None:
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -16,6 +16,7 @@ from any_agent.config import (
     TracingConfig,
 )
 from any_agent.logging import logger
+from any_agent.serving import A2AServer
 from any_agent.tools.wrappers import _wrap_tools
 from any_agent.tracing.exporter import (
     AnyAgentExporter,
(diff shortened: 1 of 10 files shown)


tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -0,0 +1,150 @@
...
+
+    serving_config = ServingConfig(
+        port = search_agent_port,
+        endpoint = f"/{search_agent_endpoint}"
+    )
+    thread = Thread(target=lambda: search_agent.serve(serving_config))
+    thread.setDaemon(True)
+    try:
+        thread.start()
+        while not search_agent._server or not search_agent._server.get_uvicorn() or not search_agent._server.get_uvicorn().started:
```
daavoo, line 114: since we are extending the class, why not add a method like `_server.ready()`  instead of these checks?

src/any_agent/tools/a2a_tool.py
```diff
@@ -0,0 +1,41 @@
+# adapted from https://github.com/google/a2a-python/blob/main/examples/helloworld/test_client.py
+import httpx
+from typing import Any, Callable
+from common.client import A2AClient
+from uuid import uuid4
+
+def a2a_query(url: str, description: str) -> Callable[[str], str]:
```
daavoo, line 7: We should not require `description` as arg. There are method to retrieve the agent card from the URL where we could get the info we need to create a descriptive docstring

src/any_agent/frameworks/any_agent.py
```diff
@@ -49,6 +50,8 @@ def __init__(
         self._mcp_servers: list[_MCPServerBase[Any]] = []
         self._main_agent_tools: list[Any] = []
 
+        self.server: A2AServer
```
daavoo, line 53: Not sure I follow what this line is doing, inside `serve` the stuff gets stored in `self._server`

src/any_agent/frameworks/any_agent.py
```diff
@@ -23,6 +23,7 @@
     get_instrumenter_by_framework,
 )
 from any_agent.tracing.trace import _is_tracing_supported
+from any_agent.serving import A2AServer
```
daavoo, line 26: The serving imports should all be lazy / optional otherwise any code will break when the deps are not installed

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: After the previous discussions, I didn't know if you wanted to keep the running infra details (for example, the thread supporting the server, whether the serving functions are async or not, etc) in the server class, so I have kept them in the test class itself. https://github.com/mozilla-ai/any-agent/pull/258 has an async server with lifetime control integrated in the server class itself (to be refactored a bit with your latest comments), which I prefer to a threaded implementation. If it's ok, I'll take that approach in this PR too.

src/any_agent/tools/a2a_tool.py
javiermtorres: ok, let's retrieve the agent card and get the description from there 👍

src/any_agent/frameworks/any_agent.py
javiermtorres: I'll check it out. The A2AServer further wraps a Starlette App deployed on a uvicorn Server, so it could be one of those _or_ I may be getting the name wrong.

src/any_agent/frameworks/any_agent.py
javiermtorres: Ok

src/any_agent/tools/a2a_tool.py
```diff
@@ -0,0 +1,41 @@
+# adapted from https://github.com/google/a2a-python/blob/main/examples/helloworld/test_client.py
+import httpx
```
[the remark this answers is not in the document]
javiermtorres: I'll use feature flags as we do with the rest of the packages to check if this is available. Maybe we could make a couple of extra support IDs, like a2a, including these dependencies.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: I'll keep the threaded implementation right now since it seems it's easier to integrate right now. Let's discuss async/threads when the official a2a sdk is used.

PR #290, opened 2025-05-21
249 add our own instrumentation

### LINKED ISSUE #99: Support Google ADK tracing

### LINKED ISSUE #100: Support Agno tracing

### LINKED ISSUE #162: Add option to disable OpenAI's `BackendSpanExporter`
I started seeing these errors in the logging and I'm not sure where they come from or whether it's a problem or not.

```
[non-fatal] Tracing: request failed: The read operation timed out
[non-fatal] Tracing: server error 500, retrying
```

### LINKED ISSUE #240: Support TinyAgent tracing

### LINKED ISSUE #249: Add our own `instrumentation`.

### LINKED ISSUE #256: Multi-Agent Not functioning as expected.

daavoo:
Closes #99 
Closes #100 
Closes #162
Closes #240 
Closes #249 
Closes #256


src/any_agent/evaluation/evaluators.py
```diff
@@ -87,11 +91,31 @@ def llm_evaluate_with_criterion(
     return EvaluationResult.model_validate(evaluation)
 
 
+def _construct_evidence(trace: AgentTrace) -> str:
```
javiermtorres, line 94: Maybe build a json, and then make a str? It could be reused by another sw at some other stage.

src/any_agent/frameworks/any_agent.py
```diff
@@ -141,23 +139,40 @@ async def _load_tools(
...
-            self._instrumenter = None
-            return
-        self._exporter = AnyAgentExporter(self.framework, self._tracing_config)
-        self._tracer_provider.add_span_processor(SimpleSpanProcessor(self._exporter))
-        self._instrumenter = get_instrumenter_by_framework(self.framework)
-        self._instrumenter.instrument(tracer_provider=self._tracer_provider)
+        self._trace_provider = TRACE_PROVIDER
+        self._tracer = self._trace_provider.get_tracer("any_agent")
+        self._exporter = _AnyAgentExporter(self._tracing_config)
+        self._trace_provider.add_span_processor(SimpleSpanProcessor(self._exporter))
```
javiermtorres, line 145: Let's open a new issue to discuss additional exporters.

src/any_agent/tracing/instrumentation/google.py
```diff
@@ -0,0 +1,215 @@
+from __future__ import annotations
```
javiermtorres, line 1: As mentioned, I'd even instrument this and not use an api that might change at any time.

javiermtorres (approved):
LGTM

src/any_agent/frameworks/agno.py
```diff
@@ -111,11 +110,9 @@ async def _load_agent(self) -> None:
...
-    async def run_async(self, prompt: str, **kwargs: Any) -> "AgentTrace":
+    async def _run_async(self, prompt: str, **kwargs: Any) -> str:
         if not self._agent:
             error_message = "Agent not loaded. Call load_agent() first."
             raise ValueError(error_message)
         result: RunResponse = await self._agent.arun(prompt, **kwargs)
-        return AgentTrace(
-            final_output=result.content,
-        )
+        return str(result.content)
```
njbrake, line 118: Why do you need to cast it to a string?

src/any_agent/frameworks/any_agent.py
```diff
@@ -211,7 +226,7 @@ def agent(self) -> Any:
 
     def exit(self) -> None:
         """Exit the agent and clean up resources."""
-        if self._instrumenter is not None:
-            self._instrumenter.uninstrument()  # otherwise, this gets called in the __del__ method of Tracer
-            self._instrumenter = None
+        if self._instrumentor is not None:
+            self._instrumentor.uninstrument()  # otherwise, this gets called in the __del__ method of Tracer
```
njbrake, line 230: Should the comment be changed? Since Tracer() isn't a class anymore, right

src/any_agent/tracing/instrumentation/agno.py
```diff
@@ -0,0 +1,124 @@
...
+                    {
+                        "gen_ai.operation.name": "call_llm",
+                        "gen_ai.request.model": instance.id,
+                    }
+                )
+
+                assistant_message, has_tool_calls = await wrapped(*args, **kwargs)
+
+                if assistant_message.content:
+                    span.set_attributes(
```
njbrake, line 33: idk can there be some check here to add "genai.input" with the prompt if it's the first call_llm span?

src/any_agent/tracing/instrumentation/langchain.py
```diff
@@ -0,0 +1,190 @@
...
+
+        self._original_ainvoke = langgraph.pregel.Pregel.ainvoke  # type: ignore[attr-defined]
+        wrap_function_wrapper(  # type: ignore[no-untyped-call]
+            "langgraph.pregel", "Pregel.ainvoke", wrapper=wrap_ainvoke
+        )
+
+    def uninstrument(self) -> None:
+        import langgraph
+
+        langgraph.pregel.Pregel.ainvoke = self._original_ainvoke  # type: ignore[attr-defined]
```
njbrake, line 189: Should there be any sort of handling if someone calls uninstrument before they call instrument? Seems like doing that will make this crash.

This comment applies to a few other frameworks too

src/any_agent/tracing/instrumentation/llama_index.py
```diff
@@ -0,0 +1,73 @@
+# mypy: disable-error-code="no-untyped-call, no-untyped-def"
```
njbrake, line 1: Why are these disabled?

tests/integration/test_load_and_run_agent.py
```diff
@@ -29,6 +27,83 @@ def check_uvx_installed() -> bool:
         return False
 
 
+def assert_trace(agent_trace: AgentTrace, agent_framework: AgentFramework) -> None:
```
njbrake, line 30: Nice re-org. Much better.

tests/unit/tracing/test_unit_agent_trace_duration.py
njbrake, line 1: Why did these tests get removed?

njbrake (commented):
Awesome work, I'm so excited for this improvement to the tracing. It's a big refactor so I want to make sure we don't lose testing/features accidentally. I'm ok with merging, but before you do please have a look at my comments.

src/any_agent/evaluation/evaluators.py
daavoo: This is a copy of functionality that was in the removed `TracinProcessor`.

I do think there are things we could revisit (in a different PR) like adapting the evidence headers and content based on whether is an llm or tool call.

cc @njbrake

src/any_agent/tracing/instrumentation/google.py
daavoo: > As mentioned, I'd even instrument this and not use an api that might change at any time.

Not sure I follow what you mean. Is it more likely for a public API (the callbacks) to change than some internal method we would patch?

src/any_agent/tracing/instrumentation/langchain.py
daavoo: > Should there be any sort of handling if someone calls uninstrument before they call instrument?

How could this happen? The instrumentation happens during agent creation.
Someone would need to manually import the instrumentor (which is marked as `_private`), instantiate it and call `uninstrument`?

src/any_agent/tracing/instrumentation/llama_index.py
daavoo: Not sure if they should be disabled in this file 🤔 I was using those in the modules that use `wrapt`

tests/unit/tracing/test_unit_agent_trace_duration.py
daavoo: because, as far as I could tell, they didn't test anything that is not already tested in the integration?

src/any_agent/tracing/instrumentation/langchain.py
njbrake: Yeah I'm referring to that second edge case. It's marked private but that hasn't always stopped their usage. For those using _private fields it's sort of a "use at your own risk" scenario so I'm ok with keeping it as-is, but wanted to draw attention that this is technically a corner case.

tests/unit/tracing/test_unit_agent_trace_duration.py
njbrake: I wanted to have it in the unit testing since we run the unit tests more often, they're faster, and give an earlier signal. I'm not heartbroken if they're removed, but I was trying to have as much coverage as possible in unit tests, so that we don't have to rely on integration tests to catch all the bugs.

src/any_agent/tracing/instrumentation/langchain.py
daavoo: will update the code just for extra safety

docs/tracing.md
```diff
@@ -30,125 +31,13 @@ Tracing will output standardized console output regardless of the
 framework used.
 
 ```console
-──────────────────────────────────────────────────────────────────────────── LLM ─────────────────────────────────────────────────────────────────────────────
```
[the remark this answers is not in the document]
daavoo: i am so happy with the setup (thanks to the `assets` logic you added)

src/any_agent/tracing/instrumentation/langchain.py
daavoo: updated code and added tests

tests/unit/tracing/test_unit_agent_trace_duration.py
daavoo: I see what you mean. My problem with this one is that it relies on the output generated by the integration tests, so it is a little bit of a cat and mouse game (you might catch something only because traces are not updated and vice-versa).

I am happy to try to revisit how to add more unit coverage (but I think it should be based on mocks?)

src/any_agent/tracing/__init__.py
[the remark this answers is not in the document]
njbrake: @daavoo any thoughts about this one?

src/any_agent/evaluation/evaluators.py
njbrake: Yep once this tracing goes into main, I can do some more cleanup of how we provide evidence to the llm-as-judges 🙌

tests/unit/tracing/test_unit_agent_trace_duration.py
njbrake: No problem. We can remove it. You're right it is covered by the integration tests so I'm still confident that we will catch a problem in tests 👍

src/any_agent/tracing/__init__.py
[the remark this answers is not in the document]
daavoo: missed this one 😅 . will do

njbrake (approved):
I ran the code locally and looked at the outputs and omg it is so beautiful. Merge this thing soon I can't wait to have it available!!!

PR #298, opened 2025-05-22
[BUG]: Filter tools not working as expected

njbrake:
Filter tools wasn't actually filtering 🤦 . I think this error returned during the MCP architecture overhaul a few weeks ago. This should fix the behavior, but I will work on writing a good test to go with this before we merge, to make sure we catch it if it happens again.

### DIFF
--- a/src/any_agent/tools/mcp/mcp_connection.py
+++ b/src/any_agent/tools/mcp/mcp_connection.py
@@ -44,20 +44,20 @@ class _MCPConnection(BaseModel, ABC, Generic[T]):
         if not requested_tools:
             return tools
 
-        tool_names = [
-            tool.name if isinstance(tool, HasName) else tool for tool in tools
-        ]
+        # Map tool names to tool objects
+        name_to_tool = {tool.name if isinstance(tool, HasName) else tool: tool for tool in tools}
 
-        found_tools = [tool for tool in tool_names if tool in requested_tools]
-
-        if len(found_tools) != len(requested_tools):
-            error_message = (
-                dedent(
-                    f"""Could not find all requested tools in the MCP server:
-                    Requested ({len(requested_tools)}): {requested_tools}
-                    Set ({len(tool_names)}):   {tool_names}
-                """
-                ),
+        # Check all requested tools are present
+        missing_tools = [name for name in requested_tools if name not in name_to_tool]
+        if missing_tools:
+            error_message = dedent(
+                f"""Could not find all requested tools in the MCP server:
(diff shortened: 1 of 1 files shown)


tests/unit/tools/mcp/test_unit_all_mcp.py
njbrake, line 1: Thanks to Claude here for the nice unit test 🌻

tests/unit/tools/mcp/test_unit_all_mcp.py
```diff
@@ -35,3 +35,38 @@ async def test_mcp_tools_loaded(
     await mcp_server._setup_tools(mcp_connection=mcp_connection)
 
     assert mcp_server.tools == list(tools)
+
+
+def test_filter_tools(
```
daavoo, line 40: Does  `test_sse_tool_filtering` overlap with this?
Also, why was not catching anything? is just a trap name to include `filtering` (I can't really tell what is going on in the test from GitHub UI because of the fixtures used)?

src/any_agent/tools/mcp/mcp_connection.py
```diff
@@ -38,26 +38,25 @@ def server(self) -> "MCPServer | None":
...
+        if missing_tools:
+            error_message = dedent(
+                f"""Could not find all requested tools in the MCP server:
+                Requested ({len(requested_tools)}): {requested_tools}
+                Set ({len(name_to_tool)}):   {list(name_to_tool.keys())}
+                Missing: {missing_tools}
+            """
             )
             raise ValueError(error_message)
-        return tools
```
daavoo, line 63: So, this should have been `found_tools` and it would work, right?

daavoo (approved):

tests/unit/tools/mcp/test_unit_all_mcp.py
njbrake: That test was a trap, because after the fixture redesign, it was checking whether tools == tools (aka no filtering), it wasn't checking whether a tool was being filtered. That's a good call out: I'll fix that test. I'll keep it around because it does check the functionality at the _setup_tools() level (more than just the _filter_tools())  so I think it's nice to have both.

src/any_agent/tools/mcp/mcp_connection.py
njbrake: Correct. I did some extra refactoring to make it maybe easier to read, but yeah all I technically had to do was return found_tools

PR #308, opened 2025-05-22
feat(serving): Introduce A2A Python SDK

njbrake:
Not yet in final state, I need to add some tests and also clean up the serving functionality. But this is basically feature parity with the manual stuff from the a2a/samples/common imports


tests/integration/test_serve_agent.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+    agent.serve(serving_config=ServingConfig(port=SERVER_PORT))
+
+
+@pytest.mark.asyncio
+async def test_agent_serving_and_communication():
+    # Start the agent in a subprocess
+    proc = multiprocessing.Process(target=run_agent, daemon=True)
+    proc.start()
+    time.sleep(5)
```
njbrake, line 34: I know I need to tweak this so that it's not a sleep, but runs health checks

javiermtorres:
Is the purpose of this PR to add the SDK, or to add tests on serving? If it deals mainly with the new SDK, then IMHO this is the right time to move to a proper handling of the underlying uvicorn server. We have used `uvicorn.run` just because https://github.com/google/A2A/blob/09c36bd94f74ea1a1a4e0db5f65ba12b102171ba/samples/python/common/server/server.py#L65 forced us to it, but we don't have that dependency anymore.

tests/integration/test_serve_agent.py
javiermtorres: TBF the client will (or at least can) be reconfigured for retries so it's not so important. If the uvicorn server were available and in the same memory space we could just check https://github.com/encode/uvicorn/blob/fa3d9d27dbe7ecbc68e0992309d0ba49ead5640b/uvicorn/server.py#L57 and then shut it down with https://github.com/encode/uvicorn/blob/fa3d9d27dbe7ecbc68e0992309d0ba49ead5640b/uvicorn/server.py#L262 .

pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "setuptools.build_meta"
 name = "any-agent"
 readme = "README.md"
 license = {text = "Apache-2.0"}
-requires-python = ">=3.11"
+requires-python = ">=3.13"
```
daavoo, line 9: um. I would not do this and just make it clear that a2a requires 3.13

pyproject.toml
```diff
@@ -59,8 +59,12 @@ smolagents = [
   "smolagents[mcp]>=1.14.0",
 ]
 
+serve = [
+  "a2a-sdk",
+]
+
 all = [
-  "any-agent[agno,google,langchain,llama_index,openai,smolagents]",
+  "any-agent[agno,google,langchain,llama_index,openai,smolagents,serve]",
```
daavoo, line 67: I would not put this as part of `all` and instead clarify in the docs how to install for serving.

I would like for us to stop advertising `all` in the docs and instead default to use what works with the minimal built-in deps (tinyagent)

src/any_agent/config.py
```diff
@@ -98,7 +98,7 @@ class ServingConfig(BaseModel):
     endpoint: str = "/"
     """Will be pass as argument to `Starlette().add_route`"""
 
-    version: str = "0.1.0"
+    version: str = "1.0.0"
```
daavoo, line 101: I don't really understand what is this version for 😅

src/any_agent/frameworks/any_agent.py
```diff
@@ -16,6 +16,7 @@
     TracingConfig,
 )
 from any_agent.logging import logger
+from any_agent.serving.server import serve_a2a
```
daavoo, line 19: We should make this import also lazy, like we handle `_get_a2a_server`

src/any_agent/serving/agent_executor.py
```diff
@@ -0,0 +1,32 @@
...
+        event_queue: EventQueue,
+    ) -> None:
+        query = context.get_user_input()
+        agent_trace = await self.agent.run_async(query)
+        assert agent_trace.final_output is not None
+        event_queue.enqueue_event(new_agent_text_message(agent_trace.final_output))
+
+    @override
+    async def cancel(self, context: RequestContext, event_queue: EventQueue) -> None:
+        msg = "cancel not supported"
```
daavoo, line 31: How is this implemented in the examples?

src/any_agent/serving/server.py
```diff
@@ -2,21 +2,33 @@
...
-        agent_card=_get_agent_card(agent, serving_config),
-        task_manager=AnyAgentTaskManager(agent),
+def _get_a2a_server(
+    agent: AnyAgent, serving_config: ServingConfig
+) -> A2AStarletteApplication:
+    agent_card = _get_agent_card(agent, serving_config)
+
+    request_handler = DefaultRequestHandler(
+        agent_executor=AnyAgentExecutor(agent),
+        task_store=InMemoryTaskStore(),
```
daavoo, line 24: 💅 this could probably be an option to be set in `serving_config` like we have `model_type` and `DEFAULT_MODEL_TYPE` in agents.

daavoo (changes requested):
I would not like to merge this with the current `pyproject.toml` updates.

pyproject.toml
njbrake: Ooh good idea. I'll do that.

pyproject.toml
njbrake: https://github.com/mozilla-ai/any-agent/issues/317

src/any_agent/config.py
njbrake: Haha yeah I didn't either, but it's what was being used in the a2a-python demos, and I was trying to figure out why the server wasn't working even though it said it supported python 12 lol. Turns out this version number was not the answer, there's still an issue in a2a-python for py12 support. I'll undo this change because it's not needed

src/any_agent/serving/agent_executor.py
njbrake: This exact same way 😭 https://github.com/google/a2a-python/blob/main/examples/langgraph/agent_executor.py#L91

src/any_agent/frameworks/any_agent.py
javiermtorres: Along the lines of https://github.com/mozilla-ai/any-agent/blob/e7873751c12305d54bb906d398e2e04c3a7a55ff/src/any_agent/serving/server.py#L20 maybe?

src/any_agent/serving/agent_executor.py
javiermtorres: Wot 😱

src/any_agent/serving/server.py
njbrake: Since we're going to be updating the way we serve via uvicorn, let's take care of this in a future PR. https://github.com/mozilla-ai/any-agent/issues/318 to track

njbrake:
Integration test ran and passed on github action:

```
[gw3] [ 60%] PASSED tests/integration/test_serve_agent.py::test_agent_serving_and_communication
```

✅

PR #315, opened 2025-05-23
Add async implementation with retro compatibility

javiermtorres:
This PR allows a finer control of the underlying uvicorn server that any-agent uses to handle agents. It is backwards compatible but also provides a way to handle the server explicitly in an async context. It will help unblocking https://github.com/mozilla-ai/any-agent/pull/286 and https://github.com/mozilla-ai/any-agent/pull/258.


src/any_agent/serving/server.py
```diff
@@ -27,8 +38,35 @@ def _get_a2a_server(
...
+
+def serve_a2a(
+    server: A2AStarletteApplication, host: str, port: int, log_level: str = "warning"
+) -> None:
     """Serve the A2A server."""
-    import uvicorn
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:
+        loop = asyncio.new_event_loop()
```
njbrake, line 66: To be consistent with how we use async in any_agent, can we do
```
return asyncio.get_event_loop().run_until_complete(
            self.run_async(prompt, **kwargs)
        )
```

If we want to change how we manage the asyncio loop, lets handle that in a separate PR, and change all the sync run/serve implementations.

src/any_agent/frameworks/any_agent.py
```diff
@@ -159,26 +166,53 @@ def run(self, prompt: str, **kwargs: Any) -> AgentTrace:
             self.run_async(prompt, **kwargs)
         )
 
-    def serve(self, serving_config: ServingConfig | None = None) -> None:
+    def serve(
+        self, serving_config: ServingConfig | None = None, log_level: str = "warning"
```
njbrake, line 170: Does it make sense for the `log_level` be a part of the ServingConfig, instead of a new parameter?

javiermtorres: Yep 👍🏼 Didn't think about that one.

src/any_agent/serving/server.py
javiermtorres: Ok, _now_ I get what you mean. I was protecting against the loop not having been created, but if you think that's not an issue I'll make the style consistent  👍🏼

njbrake: I'm not totally sure which design is better, but at least for this PR lets make the implementation consistent across any-agent, and we can consider a change to all implementations in any-agent at a later PR 🚀

njbrake (approved):
Looks good to me!

src/any_agent/serving/server.py
```diff
@@ -27,8 +30,34 @@ def _get_a2a_server(
...
 
-    uvicorn.run(server.build(), host=host, port=port)
+    # Note that the task should be kept somewhere
+    # because the loop only keeps weak refs to tasks
+    # https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
+    async def run() -> None:
+        (task, _) = await serve_a2a_async(server, host, port, log_level)
+        await task
+
+    return asyncio.get_event_loop().run_until_complete(run())
```
daavoo, line 63: Sorry if I missed something.

Do we really need to do this? We do the wrapping in the `run` / `run_async` because some frameworks didn't provide a sync API.

Couldn't we just use the uvicorn sync API?

javiermtorres: We can go from an async api to a sync api, but not vice versa. In tests, and in a foreseeable future, we need to set up a server in an async context to allow other async tasks to be run; for example, a client and a server, or more than one server.

A sync run call will just block the process, which works for simple scenarios but prevents more complex setups.

njbrake: If I understand, @daavoo  isn't questioning the need for an async server, we're on the same page that we need serve_async. The question is whether serve() should call serve_async under the hood, or should have it's own uvicorn.run command that uses the sync uvicorn command.

Personally, I am ok with dealing with this question at a later point.

PR #319, opened 2025-05-23
feat: Introduce a new action for setting up Python and dependencies

njbrake:
Two things:

1. Consolidate logic for setting up dependencies into action.yml
2. Run tests in parallel

Shoutout to Cursor here for suggesting the "action" yml, I didn't know about that.

This commit adds a new GitHub Action to streamline the setup of Python 3.13 and manage dependencies for different groups (tests, docs, lint). The action is integrated into existing workflows, replacing previous steps for checking out the repository and installing dependencies. This enhances maintainability and reduces redundancy across workflows.


daavoo (commented):
Not fully on board with this:

- Not sure how much actual total time it saves (would like to see the numbers) since we have to rebuild the test env 🤔 
- Don't like that we are installing a full env now for docs and lint (we could update the action.yml to have an input that accepts `--group` but TBH it already feels a little overkill without that

njbrake:
> Not fully on board with this:
> 
>     * Not sure how much actual total time it saves (would like to see the numbers) since we have to rebuild the test env on each parallel job 🤔
> 
>     * Don't like that we are installing a full env now for docs and lint (we could update the action.yml to have an input that accepts `--group` but TBH it already feels a little overkill without that

I'm ok with killing this one. It only saves about 30 seconds, which I understand atm may not be worth the added complexity.

PR #340, opened 2025-05-26
fix(tracing): Show final answer in the console.

daavoo:
It was not being shown because we were passing a non-JSON string to `rich.JSON` class (and `rich.JSON` was failing silently).

Screenshot 2025-05-26 162115


tests/unit/tracing/test_unit_exporter.py
```diff
@@ -13,12 +13,26 @@ def readable_spans(agent_trace: AgentTrace) -> list[ReadableSpan]:
...
         exporter = _AnyAgentExporter(TracingConfig())
         exporter.export(readable_spans)
         console_mock.return_value.print.assert_called()
+        if request.node.callspec.id not in ("TINYAGENT_trace", "SMOLAGENTS_trace"):
+            panel_mock.assert_any_call(
+                agent_trace.final_output,
+                title="OUTPUT",
+                style="white",
+                title_align="left",
+            )
```
njbrake, line 35: why are these two frameworks being skipped?

daavoo: They don't make a final LLM call but rather a tool call, which produces a different Panel

njbrake: Nitpick but would be good to have that explanation inside a comment so that its easy to find if/when smolagents changes their implementation or we change the tinyagent implementation

PR #344, opened 2025-05-27
fix(tracing): Handle tool outputs of different types.

daavoo:
- Create `instrumentation/common.py` with reusable `_set_tool_outputs`.
- Update `exporter` to properly handle different `output.type`. Extract `_get_output_panel` as separate function.
- Update tests and traces.


daavoo:
@HareeshBahuleyan could you test installing from this branch?

PR #345, opened 2025-05-27
Add end-of-file newline to JSON generated by `--update-trace-assets`.

daavoo:
Every time I updated the traces the pre-commit was updating the json because of the missing newline.


njbrake:
Hmm. Maybe we could add an ignore just like we do with embr files? https://github.com/mozilla-ai/any-agent/blob/main/.pre-commit-config.yaml#L18-L19

Your option might be prefereable to the ignore setting in the pre-commit config

daavoo:
> Hmm. Maybe we could add an ignore just like we do with embr files? https://github.com/mozilla-ai/any-agent/blob/main/.pre-commit-config.yaml#L18-L19
> 
> Your option might be prefereable to the ignore setting in the pre-commit config

In theory (https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline/729795#729795) we should add the newline character

PR #351, opened 2025-05-27
Fix minor details to align with Google ADK 1.0.0

javiermtorres:
In the pass to 1.0.0, some of the calls have been made async. Also some names have changed. Finally, the way the session is handled is now slightly different. Changing a few mocks to AsyncMocks also fixes almost all tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ agno = [
 ]
 
 google = [
-  "google-adk==1.0.0"
+  "google-adk>=1.0.0, <2"
 ]
 
 langchain = [


pyproject.toml
```diff
@@ -29,7 +29,7 @@ agno = [
 ]
 
 google = [
-  "google-adk==0.5.0" # until https://github.com/mozilla-ai/any-agent/issues/287 is resolved
+  "google-adk==1.0.0"
```
daavoo, line 32:
```suggestion
  "google-adk>=1.0.0, <2"
```

Issue #232, opened 2025-05-09
Add support for `streaming` with A2A


daavoo:
I wanted to keep this to the minimum for now. 

Adding `streaming` capabilities would require us to implement some sort of `run_stream` method in AnyAgent, which I am not sure it's worth it.

I haven't check what the other 2 capabilities mean in practice.

_Originally posted by @daavoo in https://github.com/mozilla-ai/any-agent/pull/210#discussion_r2076090951_

njbrake:
I think we need this in order to allow for a continued conversation with an A2A agent. I was looking at the example  https://google-a2a.github.io/A2A/tutorials/python/7-streaming-and-multiturn/#running-the-langgraph-server

The code is here: https://github.com/google-a2a/a2a-samples/blob/main/samples/langgraph/agent_executor.py  .

We need some way to indicate that the agent server may need to get more feedback the client? It's not clear to me how we could do a multiturn conversation with our current design. I get a response back like this:

```
{
  "id": "755cad71-8528-4169-bada-d001b3219683",
  "jsonrpc": "2.0",
  "result": {
    "contextId": null,
    "kind": "message",
    "messageId": "dea72c7d-fbe5-4488-9fc4-103ccb0363e0",
    "metadata": null,
    "parts": [
      {
        "kind": "text",
        "metadata": null,
        "text": "The current time in the America/New_York timezone is 7:28 PM on May 27, 2025. Would you like to know the time in any other timezone?"
      }
    ],
    "referenceTaskIds": null,
    "role": "agent",
    "taskId": null
  }
}
```

Since both the taskId and contextId are null, I don't think I can make a second request from the client side to continue the conversation.

njbrake:
Ok, so I think we were both right and wrong.

Push notifications is a mechanism for handling long running tasks, so that the client agent can subscribe to be notified with their mechanism of choice.

Streaming is a way to hold the connection open via SSE (https://google-a2a.github.io/A2A/specification/#72-messagestream) 

Even without push notifications or streaming, you can still support a multiturn conversation, but it affects what responsibility lands on the client.

**push_note=False and streaming=False**
Client gets response back which may be a message or a task, and must handle polling for updates as appropriate. https://google-a2a.github.io/A2A/specification/#71-messagesend

**streaming=True**
Client responds to whatever comes back via SSE, no polling https://google-a2a.github.io/A2A/specification/#72-messagestream

**push_note=True**
[...]

Issue #331, opened 2025-05-24
failed to add in a2a demo app


NiuBlibing:
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/test/ref/llm/client/A2A/demo/ui/.venv/lib/python3.12/site-packages/pydantic/main.py", line 253, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for AgentCard
skills.0.tags
  Field required [type=missing, input_value={'id': 'openai-expert-sea...he top search results.'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/missing

NiuBlibing:
> Hi [@NiuBlibing](https://github.com/NiuBlibing) , thanks for the issue! I can take a look: Last week Google released a python SDK for a2a, and I think it may now be incompatible with their A2A demo code. We may need to update our docs with a new usage example. Would you be able to provide any additional information about what documentation you were following that resulted in this error?

@njbrake Hi
My steps is the following:

1. pip install 'any-agent[a2a]' 'any-agent[serve]' openinference-instrumentation-openai openinference-instrumentation-openai-agents  openai-agents （Here [docs ](https://mozilla-ai.github.io/any-agent/serving/)mising some package, I added it)
2. add openai key and base url by set env
3. start openai_expert.py and google_expert.py
4. start a2a demo ui (The [link](https://github.com/google/A2A/blob/main/demo/README.md) is moved for the latest A2A version )
5. add url to a2a demo ui (failed)

PR #360, opened 2025-05-28
docs: Update serving documentation and add A2A tutorial notebook

njbrake:
- Enhanced the serving documentation by clarifying the installation command and adding the `serve_async` method.
- Introduced a new Jupyter notebook tutorial for serving an agent with A2A, demonstrating setup and interaction.
- Skipped tests for `serving.md` due to multi-process server limitations.
View the rendered notebook here https://github.com/mozilla-ai/any-agent/blob/359-simple-a2a-cookbook-example/docs/cookbook/serve_a2a.ipynb

I'd like to make a more advanced cookbook example where you can have a multiturn conversation with the agent, but I think we first need https://github.com/mozilla-ai/any-agent/issues/232 to be finished to enable that behavior.


njbrake:
This should also fix the docs so that it's clear that https://github.com/mozilla-ai/any-agent/issues/331 isn't supported

daavoo:
> This should also fix the docs so that it's clear that #331 isn't supported

Do you know why is not supported? I can take a quick look today, maybe it is something simple we can update

njbrake:
> > This should also fix the docs so that it's clear that #331 isn't supported
> 
> Do you know why is not supported? I can take a quick look today, maybe it is something simple we can update

@daavoo #331 has provided some more info about how the error was raised: I don't mind going through to try to see what is going wrong. You make a good point that if it's a simple fix (or exposing something that's wrong in our implementation) we should adjust appropriately.

PR #376, opened 2025-05-28
Add evaluation agent for streamiined evaluation capabilities

njbrake:
- Introduced a new `AgentTooling` class to streamline access to agent trace outputs, including final output, token usage, step count, and evidence from spans.
- Implemented a `get_agent` function to create an `AnyAgent` instance with the new tooling, enhancing the evaluation process.
- Updated `evaluate.py` to use the renamed `evaluate_checkpoints` function for clarity and consistency.

This update improves the overall evaluation framework, making it easier to gather and analyze agent performance data!

### DIFF
--- a/src/any_agent/evaluation/agent.py
+++ b/src/any_agent/evaluation/agent.py
@@ -113,5 +113,5 @@ def get_agent(trace: AgentTrace, model: str) -> AnyAgent:
     return AnyAgent.create(
         "tinyagent",
         agent_config=agent_config,
-        tracing=TracingConfig(console=True),
+        tracing=TracingConfig(console=False),
     )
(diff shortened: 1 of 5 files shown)


src/any_agent/evaluation/agent.py
```diff
@@ -0,0 +1,117 @@
...
+
+        This includes information about the input, output, and tool calls for each step.
+
+        Returns:
+            str: The evidence of all the spans in the trace
+
+        """
+        evidence = ""
+        for idx, span in enumerate(self.trace.spans):
+            evidence += f"### Step {idx}\n"
```
daavoo, line 52: @njbrake should the header use `gen_ai.operation.name` instead of / in addition to`idx`? 
Would make a clear distinction regarding llm calls vs tool executions

PR #377, opened 2025-05-29
Change name of A2A tool and generated function

### LINKED ISSUE #362: Name the tool created for A2A agents
In https://github.com/mozilla-ai/any-agent/pull/336, the created tool has the name of the callable, mainly `_send_query`. However, this makes using several A2A agents via tools confusing. The name of the created tool should be configurable.

javiermtorres:
This PR changes the name of the generated A2A tool as suggested in the PR. By default, uses the name of the A2A agent as retrieved from its card. All consecutive groups of whitespace are removed from the leading and trailing positions, and replaced by `_` if occurring in intermediate positions. The name of the function itself is changed from `a2a_query` to `a2a_tool`.

Closes #362

### DIFF
--- a/tests/unit/tools/test_a2a.py
+++ b/tests/unit/tools/test_a2a.py
@@ -42,3 +42,63 @@ def test_a2a_tool_name_specific():
         )
         created_fun = asyncio.run(a2a_tool("http://example.com/test", other_name))
         assert created_fun.__name__ == f"call_{other_name}"
+
+
+def test_a2a_tool_name_whitespace():
+    fun_name = "  some_n  ame  "
+    corrected_fun_name = "some_n_ame"
+    with patch(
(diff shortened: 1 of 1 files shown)


tests/unit/tools/test_a2a.py
daavoo, line 1: Since these are cheap to run, we could also add some test for the `re` logic

daavoo (approved):
LGTM. I think some docs might also need update?

tests/unit/tools/test_a2a.py
javiermtorres: Ok, checking a few variants

javiermtorres:
Ahhh... Depends on https://github.com/mozilla-ai/any-agent/pull/365 for the docs :-/

javiermtorres:
NP, I had to check it myself too!!! 😅🙈

PR #380, opened 2025-05-29
feat(evaluation): Allow `criteria` to be a callable.

daavoo:
Simplest way I found to support code-based checkpoints. We can revisit the interface later (thought this looks kind of clean to me).

I added this because I think it would be useful to only rely on Agent/LLM as Judge for cases that can't be handled with simple trace checks.


javiermtorres:
Unless we need the criteria to be visible to some external party (and even then, I believe we could somehow keep the original code for the function somewhere), using a function for validation is imho clean and flexible 👍

njbrake (approved):
Nice! I like it.

Issue #382, opened 2025-05-29
Add `AnyAgent.as_tool()`


daavoo:
> to clarify, my idea here would be a first approach where we expose some sort of agent as tool wrapper so people would pass it in `tools`. But we will be removing `managed_agents` as an argument and stop relying on the underlying framework logic for that argument (except for those that already do agent as tools, which would behave the same) 

 _Originally posted by @daavoo in [#219](https://github.com/mozilla-ai/any-agent/issues/219#issuecomment-2897968578)_

javiermtorres:
A section will be added to the docs detailing how to wrap a second local agent in a function, as discussed. A reference to using agents as tools via the A2A (using as_tool) and MCP (directly) protocols will be included.

PR #383, opened 2025-05-29
feat?: Drop `managed_agents`.

daavoo:
snap

Opened https://github.com/mozilla-ai/any-agent/issues/382 as a separate issue. As discussed we could just go ahead and drop this even without having implemented that one first.


daavoo:
> "Multi-agent communication will be supported via A2A communication, see xxxx to follow on the progress"

I added a note. Unless I am missing something, there is nothing to follow and we already have the components. We should update the current serving docs and/or add a notebook to showcase how to link `.serve()` with `a2a_tool` end-to-end

njbrake:
> > "Multi-agent communication will be supported via A2A communication, see xxxx to follow on the progress"
> 
> I added a note. Unless I am missing something, there is nothing to follow and we already have the components. We should update the current serving docs and/or add a notebook to showcase how to link `.serve()` with `a2a_tool` end-to-end

Good with me 👍

Issue #219, opened 2025-05-07
Drop managed_agents


daavoo:
This question probably ties more into the A2A implementation (which I haven't looked at yet), but I would maybe be in favor of dropping the "managed_agents" setting entirely if it was possible to have an easy design where each agent is spun up individually and then communicates with each other using A2A

_Originally posted by @njbrake in https://github.com/mozilla-ai/any-agent/pull/213#pullrequestreview-2821486632_

PR #384, opened 2025-05-29
Fix issues in A2A integration tests

### LINKED ISSUE #357: Fix broken integration tests from #336
#336 merged to main unfortunately has some broken integration tests. Disabling these tests on main to avoid having a broken main branch testing. This ticket is to track fixing the tests and unskipping them.

javiermtorres:
This PR aims to fix those integration tests that are not working correctly at the moment. The different fixes will be explained in further comments.

Closes #357

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,10 +13,19 @@ from any_agent.config import AgentFramework
 from any_agent.logging import setup_logger
 from any_agent.tracing.agent_trace import AgentSpan, AgentTrace
 
+# TODO use several base ports, one per suite/test
+BASE_PORT = 5800
+PORT_PER_FRAMEWORK = {fw: BASE_PORT + index for index, fw in enumerate(AgentFramework)}
+
 
 @pytest.fixture(params=list(AgentFramework), ids=lambda x: x.name)
 def agent_framework(request: pytest.FixtureRequest) -> AgentFramework:
     return request.param  # type: ignore[no-any-return]
+
+
+@pytest.fixture
+def tool_agent_port(agent_framework):
+    return PORT_PER_FRAMEWORK[agent_framework]
 
 
 @pytest.fixture
--- a/tests/integration/test_load_and_run_multi_agent_a2a_tool.py
+++ b/tests/integration/test_load_and_run_multi_agent_a2a_tool.py
@@ -28,23 +28,12 @@ logger.setLevel(logging.DEBUG)
 )
 @pytest.mark.asyncio
 async def test_load_and_run_multi_agent(
-    agent_framework: AgentFramework,
+    agent_framework: AgentFramework, tool_agent_port: int
 ) -> None:
     """Tests that an agent contacts another using A2A using the adapter tool.
 
     Note that there is an issue when using Google ADK: https://github.com/google/adk-python/pull/566
     """
-    if agent_framework in [
-        AgentFramework.GOOGLE,
-        AgentFramework.TINYAGENT,
-        AgentFramework.SMOLAGENTS,
-        AgentFramework.AGNO,
-        AgentFramework.OPENAI,
-        AgentFramework.LLAMA_INDEX,
-    ]:
-        pytest.skip(
-            "https://github.com/mozilla-ai/any-agent/issues/357 tracks fixing so these tests can be re-enabled"
-        )
     kwargs = {}
 
     kwargs["model_id"] = "gpt-4.1-nano"
@@ -64,7 +53,6 @@ async def test_load_and_run_multi_agent(
     served_task = None
 
     try:
-        tool_agent_port = 5800
         tool_agent_endpoint = "tool_agent"
 
         # DATE AGENT


javiermtorres:
Since integration tests are run in parallel, different ports need to be assigned in each test instance (that is, a test with grounded parameters). Currently this is only done for A2A integration tests but this may need to be done in other tests.

javiermtorres:
SmolAgents does not support async functions as tools, apparently.

javiermtorres:
Google seems to be working ok, but the traces are not correctly assembled. I will open a new issue with some more detail.

javiermtorres:
Things work except for the Agno uninstrument:
https://github.com/mozilla-ai/any-agent/actions/runs/15400757746/job/43332530643

javiermtorres:
...but this seems a transient failure. I'll open an issue for it.
https://github.com/mozilla-ai/any-agent/actions/runs/15400757746

src/any_agent/tools/a2a.py
```diff
@@ -70,6 +72,14 @@ async def _send_query(query: str) -> str:
             result: str = response.model_dump_json()
             return result
 
+    def _send_query_sync(query: str) -> str:
```
daavoo: This is a separate issue, right?

tests/conftest.py
```diff
@@ -173,10 +182,16 @@ def build_tree(items: list[AgentSpan]) -> AgentSpan:
         trace.attributes[CHILD_TAG] = {}
         traces[k] = trace
     for trace in items:
+        k = trace.context.span_id
         if trace.parent:
             parent_k = trace.parent.span_id
             if parent_k:
-                traces[parent_k].attributes[CHILD_TAG][trace.context.span_id] = trace
+                try:
+                    traces[parent_k].attributes[CHILD_TAG][trace.context.span_id] = (
```
daavoo, line 189: what are these changes about?

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -27,20 +27,17 @@
...
+    agent_framework: AgentFramework, tool_agent_port: int
 ) -> None:
     """Tests that an agent contacts another using A2A using the adapter tool.
 
     Note that there is an issue when using Google ADK: https://github.com/google/adk-python/pull/566
     """
     if agent_framework in [
-        AgentFramework.GOOGLE,
-        AgentFramework.TINYAGENT,
         AgentFramework.SMOLAGENTS,
```
daavoo, line 39: Why are these still skipped?

daavoo (commented):
The fix makes sense. A couple questions about other changes and the frameworks that are still skipped

tests/conftest.py
javiermtorres: These deal with the Google inconsistencies when building the spans. Apparently the parents are all out of place, but the test will otherwise work ok. Would you leave the Google FW in then?

src/any_agent/tools/a2a.py
javiermtorres: Yes, I'll remove this. IIRC this was part of the troubleshooting but I forgot to undo it later, sorry.

tests/conftest.py
daavoo: Are we using `_build_tree` for any assertion?

javiermtorres: Not right now, only manual inspection. But for example this will fail for Google but not for the rest of the frameworks, so it's a sort of implicit assumption :-/ 
And it's wrong that it fails. The parent spans should be correctly setup.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: > SmolAgents does not support async functions as tools, apparently. I will remove it from the A2A integration test.

Let me check about TinyAgent and LlamaIndex.

javiermtorres: In the case of LlamaIndex, the tool execution does not appear in the spans, although it does seem to be present since the answer is ok. I'll make a separate issue for this.

TinyAgent works ok, I'll add it in again.

tests/conftest.py
```diff
@@ -13,12 +13,21 @@
 from any_agent.logging import setup_logger
 from any_agent.tracing.agent_trace import AgentSpan, AgentTrace
 
+# TODO use several base ports, one per suite/test
```
njbrake: If this is a todo, please make a github issue, and remove this comment.
```suggestion
```

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -27,19 +27,15 @@
     reason="Integration tests require `ANY_AGENT_INTEGRATION_TESTS=TRUE` env var",
 )
 @pytest.mark.asyncio
-async def test_load_and_run_multi_agent_a2a(
-    agent_framework: AgentFramework,
+async def test_load_and_run_multi_agent(
```
njbrake: Is this now the same test name as a test in the `test_load_and_run_multi_agent` file? I'm worried about a naming overlap

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
daavoo: And smolagents?

javiermtorres: Sorry to insist, but

> SmolAgents does not support async functions as tools, apparently. I will remove it from the A2A integration test.

😅

tests/conftest.py
javiermtorres: No, it's done in the fixtures. I'll remove the comment.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: AFAICT the suite name is considered to name the test, so it would be `tests/integration/test_load_and_run_multi_agent_a2a_tool.py::test_load_and_run_multi_agent`. But if you prefer I can make it use the `_a2a` prefix again.

njbrake: I filter tests using the `-k 'test_name'` parameter, and this will cause unintended behavior, so I'd prefer if it could be a unique name to make my life easier 😆

javiermtorres: Ok, I'll change the name. However, for future changes please note that `pytest -k "multi_agent and a2a"` will select the appropriate tests (since it's a condition on test names, not only literal test names).

javiermtorres:
>  Drop the logging for debugging (I would suggest the test to run with an actual debugger)

The tests were run with a debugger. In the case of future changes, logs offer insight at possible failures; for successful cases, the logs won't show. I'd retain the one liners, plus the final spans organized in a tree and the response.
 
>  Drop the `build_tree` or convert it to an actual assertion

The `KeyError` exception will be caught and turned into a specific failure. The Google framework will then be skipped.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
daavoo: For the record, `test_load_and_run_multi_agent` has been gone since #219

javiermtorres:
The specific integration tests are working ok, only unrelated integration tests are failing:
https://github.com/mozilla-ai/any-agent/actions/runs/15436542451/job/43444149516

```
=========================== short test summary info ============================
FAILED tests/integration/test_serve_agent.py::test_agent_serving_and_communication_async - SystemExit: 1
FAILED tests/integration/test_uninstrument.py::test_uninstrument[OPENAI] - assert not True
 +  where True = any(<generator object test_uninstrument.<locals>.<genexpr> at 0x7f8f281935e0>)
FAILED tests/integration/test_uninstrument.py::test_uninstrument[AGNO] - assert not True
 +  where True = any(<generator object test_uninstrument.<locals>.<genexpr> at 0x7f8f19afbed0>)
FAILED tests/integration/test_uninstrument.py::test_uninstrument[TINYAGENT] - assert not True
 +  where True = any(<generator object test_uninstrument.<locals>.<genexpr> at 0x7f8f28611be0>)
=========== 4 failed, 29 passed, 3 skipped, 1008 warnings in 42.56s ============
```

PR #392, opened 2025-05-30
Use an Agent as a tool via A2A Cookbook

njbrake:
Hopefully this makes it clearer to help people to start building some cool multiagent systems. 

https://github.com/mozilla-ai/any-agent/blob/brake/a2a_as_tool_cookbook/docs/cookbook/a2a_as_tool.ipynb


docs/cookbook/a2a_as_tool.ipynb
daavoo, line 1: Can we show an use case like the original serving docs where we create 2 agents that the main agent can route to?
Like `google-adk expert` and `openai-expert`?

docs/cookbook/a2a_as_tool.ipynb
```diff
@@ -0,0 +1,636 @@
...
+    "Note: because this tutorial relies upon advanced stdio/stderr communication using the MCP Server, it cannot be run on Google Colab.\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Install Dependencies\n",
+    "\n",
+    "any-agent uses the python asyncio module to support async functionality. When running in Jupyter notebooks, this means we need to enable the use of nested event loops. We'll install any-agent and enable this below using nest_asyncio."
```
javiermtorres, line 23: We'll eventually use Elixir 🤣

Issue #385, opened 2025-05-29
Add Uninstrument as destructor of tracer


njbrake:
Right now, if I understand right, uninstrument() of the tracer is only called when someone calls `agent.exit()`. In the old design, the tracer was automatically uninstrumented as a part of the destructor of the tracer. But I don't think this is now the case.

Should we add a call to uninstrument as a part of the __del__ of all the tracers for each framework?

PR #399, opened 2025-05-30
fix(tests): Make sure tinyagent eval traces are not included in the c…

daavoo:
…onsole outputs.

### DIFF
--- a/tests/integration/test_load_and_run_agent.py
+++ b/tests/integration/test_load_and_run_agent.py
@@ -209,7 +209,6 @@ def test_load_and_run_agent(
         assert_duration(agent_trace, (end_ns - start_ns) / 1_000_000_000)
         assert_cost(agent_trace)
         assert_tokens(agent_trace)
-        assert_eval(agent_trace)
 
         if update_trace:
             trace_path = Path(__file__).parent.parent / "assets" / agent_framework.name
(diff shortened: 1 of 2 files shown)


daavoo:
not sure if this is related to https://github.com/mozilla-ai/any-agent/issues/385#issuecomment-2922330435

tests/integration/test_load_and_run_agent.py
```diff
@@ -219,5 +218,8 @@ def write_file(text: str) -> None:
             html_output = agent._exporter.console.export_html(inline_styles=True)  # type: ignore[union-attr]
             with open(f"{trace_path}_trace.html", "w", encoding="utf-8") as f:
                 f.write(html_output.replace("<!DOCTYPE html>", ""))
+
+        agent.exit()
+        assert_eval(agent_trace)
```
njbrake, line 223: Is there a way that we can make it so that exit() is automatically called on garbage collection? I'm still wishing that we didn't need to manually call agent.exit().

njbrake: Yeah but if it's not called at all, what happens when the tracer is garbage collected (whenever that might be). Does the uninstrument happen? I just want to make sure the cleanup does happen and we don't accidentally fail to fix some sort of global state. I have no idea if this is possible in the tracing design.

Issue #368, opened 2025-05-28
Dump partial agent_trace even if agent.run() fails


HareeshBahuleyan:
For example, it failed in the middle of the run because it could not complete the task within the set max_turns , but the user would still like to check the trace.

By handling the interruption more gracefully and still return a trace.

javiermtorres:
Closed via #432

PR #400, opened 2025-05-30
feat(tinyagent): Refactor and return final_output.

daavoo:


src/any_agent/frameworks/tinyagent.py
```diff
@@ -209,199 +181,47 @@ async def _run_async(self, prompt: str, **kwargs: Any) -> str:
...
-                combined_results.append(str(tool_message["content"]))
-
-            # If an exit tool was called, return early with the combined results
-            if exit_tool_called:
-                return "\n".join(combined_results)
+                return str(current_last["content"])
 
-            return "\n".join(combined_results)
-        return str(message.content)
+        return "Max turns reached"
```
njbrake, line 224: Hmm. Can this throw an exception instead of returning? I think that's what the other frameworks would do. Might be weird to have it successfully complete but have this error. 

We have the ticket https://github.com/mozilla-ai/any-agent/issues/368 for a refactor of how we handle failures.

njbrake (commented):
TinyAgent becomes tinier, nice. I would want to be confident that all integration tests still pass before merging, but if those pass then I trust this is good. One comment about how to handle max turns exceeded.

src/any_agent/frameworks/tinyagent.py
daavoo: > Might be weird to have it successfully complete but have this error.

Afaik, it is the same behavior as before the refactor:

```py
            if current_last.get("role") != "tool" and num_of_turns > max_turns:
                logger.debug("Exiting because max turns (%s) reached", max_turns)
                return final_response or "Max turns reached"
```

daavoo:
> TinyAgent becomes tinier, nice. I would want to be confident that all integration tests still pass before merging, but if those pass then I trust this is good. One comment about how to handle max turns exceeded.

https://github.com/mozilla-ai/any-agent/actions/runs/15348976588

src/any_agent/frameworks/tinyagent.py
njbrake: Yeah. I guess what I'm proposing is scope creep, but since this was a complete overhaul of the design anyways it seemed like the ask would fit within this PR ok enough

PR #401, opened 2025-05-30
feat(mcp): Enhance client session timeout handling across multiple fr…

njbrake:
…ameworks

This update introduces a consistent approach to passing `client_session_timeout_seconds` to various MCP connection classes, including Agno, Langchain, LlamaIndex, and OpenAI. The timeout is now properly integrated into the respective client sessions for both STDIO and SSE connections, ensuring better control over session timeouts. Unit tests have been added to verify the correct behavior across these frameworks.


daavoo (approved):
I guess I will trust all those `mock` 😅

PR #403, opened 2025-05-30
feat(tracing): Introduce console exporter to prevent duplicate console writing.

njbrake:
The problem is that there is no way to remove a span processor from the trace provider, and even if you call `shutdown()` on the exporter, it still gets export calls because you can't actually destroy the exporter (because the global trace provider holds a reference to it). So this is the best way I could think of so that the orphaned exporters turn into no-ops.


njbrake:
Ok. This change does not totally fix it. This solves the problem of duplicate output, but then it introduces the problem that as soon as any agent in the python process turns on console printing, all of the agents start printing to the console 🤦

daavoo:
@njbrake I am going to refactor the exporter/instrumentation stuff to try to tackle this and the other related issues. Will send a PR later today

njbrake:
I'm going to close out this PR in favor of David's refactor of export/instrument

PR #420, opened 2025-06-03
feat: All frameworks support structured outputs, and new configuration parameter output_type enables custom Pydantic agent output schemas

njbrake:
Some of the frameworks have direct support for structured outputs, others do not. For each framework I try to use the built in functionality, but when not present (Google ADK, Llama_Index, TinyAgent), I add the functionality in a way that keeps it inside of the agent run/trace as best I can.


njbrake:
@daavoo  before I go too crazy adding implementations for each framework, would you mind taking a look to see if you agree with the direction?

To start it off, I only implemented structured output support in Google ADK, and wanted to get your thoughts.

I tried implementing the "structured inputs as another tool" idea, but ran into roadblocks because we are relying on the agent to listen to our prompt to call the structured output tool, which means we have to validate the final output at the end anyways, and if it fails after the agent returns, then we would need to re-call the agent which would generate a new trace.

src/any_agent/frameworks/google.py
```diff
@@ -98,4 +118,44 @@ async def _run_async(  # type: ignore[no-untyped-def]
...
+            except ValidationError as e:
+                if attempt >= DEFAULT_MAX_SCHEMA_VALIDATION_ATTEMPTS:
+                    msg = f"Failed to generate appropriate final answer schema after {DEFAULT_MAX_SCHEMA_VALIDATION_ATTEMPTS} attempts"
+                    raise ValueError(msg) from e
+                output_type_schema = self.config.output_type.model_json_schema()
+                fix_prompt = f"""
+                The response is invalid. Please fix it. The error is: {e}
+                The output type schema is: {output_type_schema}
+                The response was: {response}.
+                Do not return anything else other than the fixed JSON object.
```
javiermtorres: Could you check if this can be ensured via GCD (instead of through the prompt) using the parameters in each framework? In the case of OpenAI, for example: https://platform.openai.com/docs/guides/structured-outputs?api-mode=chat

njbrake: Yep, this is framework specific. Google does not support structured output the way that openai does. Some implementations will be easier, some harder 😅

javiermtorres: T_T

daavoo (commented):
I am not strongly against this, but I would like to push back a little bit / open to debate.

> I tried implementing the "structured inputs as another tool" idea, but ran into roadblocks because we are relying on the agent to listen to our prompt to call the structured output tool, which means we have to validate the final output at the end anyways, and if it fails after the agent returns, then we would need to re-call the agent which would generate a new trace.

As far as I can tell (please correct me if I am misunderstanding something), all the roadblocks are still present in these changes:

- we are relying on the agent to follow the prompt and return the structured output format
- we have to validate the agent output at the end
- if it fails, we need to re-call the agent

So, I don't see the real difference or added value of these changes vs structured output as a tool.

Having the same issues, what I think are the benefits of implementing as a tool is:

- No need to add framework-specific changes in any-agent.
- Ensure "consistent" behavior (if the agent uses the tool) across frameworks.

I would try to actually compare these changes against a tool-based implementation, where we it has a good description, good error handling, and we inject some instructions (`you must use this TOOL to provide your final answer and retry if it returns an error`)

njbrake:
@daavoo  thanks for the review! I'll try to clarify my reasoning for my approach:

>     * we are relying on the agent to follow the prompt and return the structured output format

This is not quite true:
1. Agno, Langchain, OpenAI, Smolagents, and TinyAgent all have features to support structured outputs. Assumedly each agent is adding some sort of prompt to the end of the agent run to create the structured output but we leave it up to the framework implementation.
2. Google has no support for structured outputs (when also using tools), but the framework lets me add the structured output checking inside of the same agent run session, so that the attempt to fix the output show up inside the same trace
3. Llama Index has no support for structured outputs or iterating inside of the same trace (as far as I could find), so this is the one that is true we rely on the agent to follow the prompt and return the structured output format.

>     * we have to validate the agent output at the end

We only need to validate the agent output at the end for Google, TinyAgent, and Llama Index, the rest of the frameworks handle it internally.

>     * if it fails, we need to re-call the agent

Only the case for Google, and the re-call of the agent happens inside of the same agent trace. For Llama Index we just fail because I would rather fail than re-run the agent, create a new trace, and figure out how to do this without breaking all of our tracing. For all the other frameworks, the re-calling to fix the structured output happens internally.

> So, I don't see the real difference or added value of these changes vs structured output as a tool.

The difference is that we allow each framework to follow their design for how they implement structured outputs. It's one of those things that might work better/worse depending on the framework, and is probably one of the things that may make you choose one framework over the other (or choose our tinyagent implementation if it works the best)

> 
> Having the same issues, what I think are the benefits of implementing as a tool is:
> 
>     * No need to add framework-specific changes in any-agent.

It would make it easier to implement, but I guess since I've already implemented the framework specific changes to make this happen and added a test to make sure it continues to work, I'm not concerned about this.

> 
>     * Ensure "consistent" behavior (if the agent uses the tool) across frameworks.

True, but I'm worried that the consistent behavior we may be introducing is that it will be consistently flaky 😅 

> 
> 
> I would try to actually compare these changes against a tool-based implementation, where we it has a good description, good error handling, and we inject some instructions (`you must use this TOOL to provide your final answer and retry if it returns an error`)

I don't prefer this approach because it's so model and prompt dependent, and then we don't have a way to force a re-try to fix any errors without re-running the agents (we would be unnecessarily putting every framework into the same situation that Llama-Index is in).

Hopefully this explanation helps!

njbrake:
Following a call and chat with @daavoo, I did some extra digging to de-mystify what the frameworks do under the hood. We want to answer the question "do the frameworks actually do anything different that makes it worth using them instead of our own approach with no re-try logic (like we do with llama_index)".

OpenAI: No retry https://github.com/openai/openai-agents-python/blob/main/src/agents/_run_impl.py#L332

Agno: tries out a few different ways of parsing the output https://github.com/agno-agi/agno/blob/main/libs/agno/agno/utils/string.py#L65-L115, it utilizes the structured output setting of the model if supported, but it is not supported in the agno litellm model

Langchain: Adds a node in the computation graph to invoke the model asking for structured response: https://github.com/langchain-ai/langgraph/blob/aedf974dfd7e4297ac6edb5a9a81624249abd822/libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py#L690-L696  Since it’s a directed graph I believe this means that the structured output part is only added and called at the end of the graph.

Smolagents: the FInalAnswerTool is a special class and tool that must be called to complete the agent run https://github.com/huggingface/smolagents/blob/main/src/smolagents/agents.py#L1282-L1302 

Llama_Index: no handling so we have a prompt addition to try to do our best to support (it would be akin to how OpenAI supports it: no retry logic or special handling for output schema).

src/any_agent/evaluation/evaluate.py
```diff
@@ -19,8 +21,12 @@ def evaluate(
...
 
     if evaluation_case.ground_truth and trace.final_output:
+        if isinstance(trace.final_output, BaseModel):
+            final_output = trace.final_output.model_dump_json()
+        else:
+            final_output = trace.final_output
         ground_truth_result = evaluate_final_output(
-            final_output=trace.final_output,
+            final_output=final_output,
             ground_truth_answer=evaluation_case.ground_truth,
```
njbrake, line 30: We will need to touch this again when we support https://github.com/mozilla-ai/any-agent/issues/427

src/any_agent/evaluation/evaluators.py
```diff
@@ -40,14 +40,7 @@ def evaluate_checkpoints(
...
             evaluation = checking_agent.run(prompt=checkpoint.criteria)
-            # strip out the ```json and ``` from the final output if they exist
-            if not evaluation.final_output:
-                msg = "The evaluation result is empty"
-                raise ValueError(msg)
-            final_output = evaluation.final_output.replace("```json", "").replace(
-                "```", ""
-            )
-            eval_output = AgentOutput.model_validate_json(final_output)
+            eval_output = evaluation.final_output  # type: ignore[assignment]
```
njbrake, line 43: Our evaluation agent uses structured outputs yay!

njbrake:
@daavoo  Ok this is ready for you to take a look. To keep these changes from exploding a single PR into a massive set of changes, I plan to do this in stages.

Step 1 (this PR): support structured_outputs as an optional feature of each framework
Step 2: support our custom 'augmentation' on some frameworks structured output implementation, when possible
Step 3: explore enforcing structured outputs at all times (where we supply a default pydantic model if the developer doesn't have one they want to use).

How does this sound?

Trying to eat this elephant one bite at a time 😆

tests/integration/test_load_and_run_agent.py
```diff
@@ -223,3 +225,22 @@ def write_file(text: str) -> None:
         assert_eval(agent_trace)
     finally:
         agent.exit()
+
+
+def test_output_type(agent_framework: AgentFramework) -> None:
```
daavoo: 💅 I would put this in a separate file

docs/agents/index.md
```diff
@@ -38,54 +38,55 @@ Multi-agent can be implemented today using the A2A protocol (see [A2A docs](http
...
 
 agent = AnyAgent.create(
     framework,
     AgentConfig(
         model_id="gpt-4.1-mini",
-        instructions="Extract webpage title and summary from url",
+        instructions="Check if the output contains any math",
         agent_args={
-            "planning_interval": 1,
-            "grammar": WebPageInfo
```
daavoo, line 88: @njbrake does this `grammar` think not work in smolagents?

daavoo (changes requested):
thanks for taking a step back and divide into stages 🙏 

llama-index test is failing:

https://github.com/mozilla-ai/any-agent/actions/runs/15488440780/job/43608329753

docs/agents/index.md
njbrake: Screenshot 2025-06-06 at 9 41 18 AM
The deprecated it

njbrake:
Ok, after a long afternoon of debugging 😓  I found that the repeatedly failing LLAMA_INDEX test is really not a llama index problem at all, but rather a tracing problem. I should have inspected the stack trace much more closely: the problem is the tiny agent instrumentor.

```              ^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.13/site-packages/llama_index/llms/litellm/utils.py:87: in _completion_with_retry
    return await acompletion(**kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
src/any_agent/tracing/instrumentation/tinyagent.py:97: in model_call_wrap
    _set_llm_output(response, span)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
response = <litellm.litellm_core_utils.streaming_handler.CustomStreamWrapper object at 0x7f7378298910>
span = _Span(name="call_llm gpt-4.1-mini", context=SpanContext(trace_id=0x001e00fd0d3d1a62d325b58d7b4c22da, span_id=0x6ee00068f70bc1c3, trace_flags=0x01, trace_state=[], is_remote=False))
    def _set_llm_output(response: ModelResponse, span: Span) -> None:
>       if not response.choices:
               ^^^^^^^^^^^^^^^^
E       AttributeError: 'CustomStreamWrapper' object has no attribute 'choices'
src/any_agent/tracing/instrumentation/tinyagent.py:25: AttributeError
The above exception was the direct cause of the following exception:
agent_framework = <AgentFramework.LLAMA_INDEX: 'llama_index'>
```

Because of this, I will merge this PR (with the understanding that the test is still going to show as failing).  And I have confidence that it will be resolved by https://github.com/mozilla-ai/any-agent/pull/435

Issue #313, opened 2025-05-22
Extract a test support lib, specially for integration tests


javiermtorres:
Does this code exist elsewhere and can we use that instead of copying here?

_Originally posted by @njbrake in https://github.com/mozilla-ai/any-agent/pull/258#discussion_r2103128900_

daavoo:
Do we need a lib? Can we use conftest?

javiermtorres:
Fine for me. The actual task would be gathering some tools lying around in a single place, I don't have any particular choice for it 😅

javiermtorres:
Closed since this is already being done organically.

PR #424, opened 2025-06-03
Add sync version of a2a tool

### LINKED ISSUE #343: Add sync version of `a2a_tool`
So, the `a2a_query` is currently only usable in async context. Can we create a wrapper to make it also usable in sync context?

_Originally posted by @daavoo in https://github.com/mozilla-ai/any-agent/pull/336#discussion_r2108600658_

javiermtorres:
Provides a sync version of a2a tool. Note that the returned tool is still async. This may require further changes, since some frameworks cannot handle async functions as tools.

Closes #343

### DIFF
--- a/src/any_agent/tools/a2a.py
+++ b/src/any_agent/tools/a2a.py
@@ -1,5 +1,6 @@
 # adapted from https://github.com/google/a2a-python/blob/main/examples/helloworld/test_client.py
 
+import asyncio
 import re
 from collections.abc import Callable, Coroutine
 from contextlib import suppress
@@ -28,7 +29,7 @@ with suppress(ImportError):
 async def a2a_tool(
     url: str, toolname: str | None = None, http_kwargs: dict[str, Any] | None = None
 ) -> Callable[[str], Coroutine[Any, Any, str]]:
-    """Perform a query using A2A to another agent.
+    """Perform a query using A2A to another agent (async function).
 
     Args:
         url (str): The url in which the A2A agent is located.
@@ -85,3 +86,24 @@ async def a2a_tool(
             The result from the A2A agent, encoded in json.
     """
     return _send_query
+
+
+def a2a_tool_sync(
+    url: str, toolname: str | None = None, http_kwargs: dict[str, Any] | None = None
+) -> Callable[[str], Coroutine[Any, Any, str]]:
+    """Perform a query using A2A to another agent (sync function).
+    Note that the result is still an async function.
+
+    Args:
+        url (str): The url in which the A2A agent is located.
+        toolname (str): The name for the created tool. Defaults to `call_{agent name in card}`.
+            Leading and trailing whitespace are removed. Whitespace in the middle is replaced by `_`.
+        http_kwargs (dict): Additional kwargs to pass to the httpx client.
+
+    Returns:
+        An async `Callable` that takes a query and returns the agent response.
+
+    """
+    return asyncio.get_event_loop().run_until_complete(
+        a2a_tool(url, toolname, http_kwargs)
+    )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,6 +6,7 @@ from textwrap import dedent
 from typing import Any
 from unittest.mock import AsyncMock, patch
 
+import httpx
 import pytest
 from litellm.types.utils import ModelResponse
 
@@ -180,3 +181,15 @@ def build_tree(items: list[AgentSpan]) -> AgentSpan:
             else:
                 traces[None] = trace
     return traces[None]
+
+
+def probe(url: str) -> bool:
+    """Check that the URL returns something but doesn't time out"""
+    try:
+        with httpx.Client() as client:
+            client.get(url, timeout=1)
+            return True
+    except httpx.TimeoutException:
+        return False
+    except httpx.ConnectError:
+        return False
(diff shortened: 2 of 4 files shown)


src/any_agent/tools/a2a.py
```diff
@@ -85,3 +86,24 @@ async def _send_query(query: str) -> str:
             The result from the A2A agent, encoded in json.
     """
     return _send_query
+
+
+def a2a_tool_sync(
```
njbrake, line 91: To be consistent with the rest of our api, this should be `a2a_tool` and the async implementation should be `a2a_tool_async`

tests/conftest.py
```diff
@@ -180,3 +181,15 @@ def build_tree(items: list[AgentSpan]) -> AgentSpan:
...
+def probe(url: str) -> bool:
+    """Check that the URL returns something but doesn't time out"""
+    try:
+        with httpx.Client() as client:
+            client.get(url, timeout=1)
+            return True
+    except httpx.TimeoutException:
+        return False
+    except httpx.ConnectError:
+        return False
```
njbrake, line 209: If this isn't a fixture, should it go into a "utils.py" file or something like that?

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -156,3 +180,121 @@ def get_datetime() -> str:
...
+    if agent_framework in [
+        AgentFramework.TINYAGENT,
+        AgentFramework.SMOLAGENTS,
+        AgentFramework.AGNO,
+        AgentFramework.GOOGLE,
+        AgentFramework.LLAMA_INDEX,
+    ]:
+        pytest.skip(
+            "https://github.com/mozilla-ai/any-agent/issues/357 tracks fixing so these tests can be re-enabled"
+        )
```
njbrake, line 209: Is this needed?

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -156,3 +180,121 @@ def get_datetime() -> str:
...
+    agent_model = kwargs["model_id"]
+    env_check = validate_environment(kwargs["model_id"])
+    if not env_check["keys_in_environment"]:
+        pytest.skip(f"{env_check['missing_keys']} needed for {agent_framework}")
+
+    model_args = (
+        {"parallel_tool_calls": False}
+        if agent_framework not in [AgentFramework.AGNO, AgentFramework.LLAMA_INDEX]
+        else None
+    )
```
njbrake, line 222: What is this needed for?

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -156,3 +180,121 @@ def get_datetime() -> str:
...
+    proc = multiprocessing.Process(
+        target=run_agent,
+        args=(agent_framework, date_agent_cfg, tool_agent_port, tool_agent_endpoint),
+        daemon=True,
+    )
+    proc.start()
+    while not probe(tool_agent_url):
+        time.sleep(1)
+        probe_log = f"probing <{tool_agent_url}> ..."
+        logger.info(probe_log)
```
njbrake, line 250: Is there a reason why not to serve the agent with async? I don't think you need to serve the agent sync in order to test the a2a_tool sync call, right?

njbrake (commented):
Thanks for the work here! A few comments. Could the test be simplified by using the async syntax for creating and serving the agent, and using the sync call instead of the async for getting that served agent? Idk if that would work or not.

tests/conftest.py
javiermtorres: Can be added directly to conftest, to avoid having a lot of libs lying around, but either way is ok for me. @daavoo wdyt?

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: It's currently part of the main branch :-/ It's fixed in #357

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: Hum, possibly a remnant from other tests. If it's not used anymore, I'll remove it.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: Hum. AFAICT, If you start an event loop, then you cannot use it to run to completion another async function. So, if you start by serving async'ly, then everything you do needs to be done async'ly. That's why I'm doing multiproc here; I thought about using to_thread, but probably using multiproc is easier (and already in the tests). Since we want to try a sync function, that needs to do IO, then every other IO needs to be done sync'ly in some other thread/process.

njbrake: No problem. If it needs to be done this way, no worries. Just double checking.

tests/conftest.py
javiermtorres: Let's use https://github.com/mozilla-ai/any-agent/issues/313 for this discussion.

javiermtorres:
Dropped for #431

PR #428, opened 2025-06-03
Add workflow step to check jupyter notebooks

### LINKED ISSUE #413: Github Action CI should run Jupyter cookbook examples as a part of CI checks
We should periodically run all the cookbook examples to make sure they don't fall out of date with the API. Came out of a convo with @javiermtorres

javiermtorres:
Closes #413


javiermtorres:
Notebooks running ok: https://github.com/mozilla-ai/any-agent/actions/runs/15439216300/job/43452783814

docs/cookbook/mcp_agent.ipynb
```diff
@@ -128,7 +131,7 @@
     "    return input(message + \" \")\n",
     "\n",
     "\n",
-    "agent = AnyAgent.create(\n",
```
njbrake: If you make this async then I think you should switch the rest of the calls (aka agent.run) to also use the async syntax

.github/workflows/tests.yaml
```diff
@@ -43,3 +43,12 @@ jobs:
...
         run: pytest -v tests/integration -d --tx '4*popen//python=python'
+
+      - name: Run Jupyter Notebook tests
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main' || github.event_name == 'workflow_dispatch'
+        working-directory: docs/cookbook
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        run: |
+          python -m ipykernel install --user --name=anyagent
+          for nb in `ls -1 *.ipynb`; do jupyter execute --kernel_name=anyagent $nb; done
```
njbrake: Should this be moved into the tests for docs workflow instead of tests? This check should be run whenever docs change in addition to whenever a merge to main happens

docs/cookbook/mcp_agent.ipynb
javiermtorres: AFAICT this currently runs ok, but I'll check. Depending on how we use the event loop, this may not be strictly the case.

njbrake: Even if it runs ok, we shouldn't mix the async and sync commands int the cookbok, it is confusing to people looking at any-agent for the first time.

.github/workflows/tests.yaml
javiermtorres: My opinions here:

* This is as complex as an integration test, since the cases involve serving or an e2e scenario, for example. The same event conditions would apply even in the docs tests, that is, would only run manually triggered, for example.
* I'd even restrict the trigger to changes in the doc/cookbook, plus source code, as you have mentioned.

Wdyt?

njbrake: Ah, I see what you mean. Is it possible to adjust it then so that it runs after a merge to main even if only the docs changed? I think currently it won't run if a merge to main happens and there are no changes to the source code.

 So then maybe lets adjust it so that it only runs after a merge to main (like you currently have it), but make it so that it also runs on main if docs/cookbook is changed even if the source code didn't change?

docs/cookbook/mcp_agent.ipynb
javiermtorres: Funny. It doesn't complain when running sync now :-/ It could have been that I misinterpreted the error due to the timeout previously. I'll try to revert any unnecessary asyncs.

.github/workflows/tests.yaml
javiermtorres: > Is it possible to adjust it then so that it runs after a merge to main even if only the docs changed? 

I'll check that it happens. I never remember the and/or rules in the paths, but I'll manage XD
Ok, I'll try this out.

njbrake: If it turns out not to be possible, not the end of the world 😄  This PR is a nice step forwards regardless.

javiermtorres:
Please check the conditions here: https://github.com/mozilla-ai/any-agent/blob/7ab04dddae93f1ee45884d17d0d519a0231ee6ee/.github/workflows/tests-cookbook.yaml#L3
I _think_ I got it right, but let me know if this doesn't match your expectations 😅

docs/cookbook/serve_a2a.ipynb
```diff
@@ -92,7 +92,7 @@
     "    ],\n",
     ")\n",
     "\n",
-    "time = AnyAgent.create(\n",
+    "time = await AnyAgent.create_async(\n",
```
njbrake, line 95: Good find!

PR #429, opened 2025-06-03
408 A2A agentskills should be configurable

njbrake:
Now the skills are configurable and I also tweak the API so that it clears the path to allow for future integration of other serving methods like Serving over MCP or ACP.

A2AServingConfig now has an a2a specific type called "AgentSkill" which is no longer appropriate to exist in the main ServingConfig class. This is why I made a new config and put it inside of the `serving` module.

Other than the re-org, the actual code change is in the _get_agent_card function. I also created a unit test.

### DIFF
--- a/src/any_agent/config.py
+++ b/src/any_agent/config.py
@@ -83,39 +83,6 @@ class MCPSse(BaseModel):
     model_config = ConfigDict(frozen=True)
 
 
-class ServingConfig(BaseModel):
-    """Base configuration for serving an agent.
-
-    This abstract base class defines the common interface and validation
-    for serving agents over different protocols. It cannot be instantiated
-    directly - use concrete implementations like A2AServingConfig.
-
-    Note:
-        This is an abstract base class. Use A2AServingConfig or other
-        concrete implementations instead of instantiating this directly.
-
-    """
-
-    model_config = ConfigDict(extra="forbid")
-
-    host: str = "localhost"
-    """Will be passed as argument to `uvicorn.run`."""
-
-    port: int = 5000
-    """Will be passed as argument to `uvicorn.run`."""
-
-    endpoint: str = "/"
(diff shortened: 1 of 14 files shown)


src/any_agent/config.py
```diff
@@ -84,9 +84,16 @@ class MCPSse(BaseModel):
 
 
 class ServingConfig(BaseModel):
```
daavoo: We should probably move this to `any_agent/serving/config.py`

src/any_agent/config.py
```diff
@@ -103,6 +110,9 @@ class ServingConfig(BaseModel):
     log_level: str = "warning"
     """Will be passed as argument to the `uvicorn` server."""
 
+    type: str
+    """The type of serving config (e.g. "a2a")"""
+
```
daavoo: Unless I am misunderstanding the expected usage (pass protocol-specific config object), this would not be needed:

```suggestion
```

src/any_agent/frameworks/any_agent.py
```diff
@@ -172,19 +172,33 @@ async def run_async(self, prompt: str, **kwargs: Any) -> AgentTrace:
         return trace
 
     def serve(self, serving_config: ServingConfig | None = None) -> None:
```
daavoo: We should change to `A2AServingConfig` and later to `A2AServingConfig | MCPServingConfig`

src/any_agent/config.py
njbrake: Yep this was sort of future-proofing and I used it when printing out the error if someone serves with an unsupported config. But since you're ok with moving the ServingConfig into the any_agent.serving module, this makes my changes much simpler and I'll clean up the code

src/any_agent/frameworks/any_agent.py
njbrake: works for me 👍

PR #431, opened 2025-06-04
Sync A2A Tool

njbrake:
I think we need to refactor our use of asyncio in our sync calls so that it appropriately attaches to or runs a new loop. Once that gets fixed, adding a sync call for getting the A2A Tool gets simpler (just requires an additional call to grab the agent name and doc right away).

### DIFF
--- a/tests/integration/test_serve_agent.py
+++ b/tests/integration/test_serve_agent.py
@@ -41,6 +41,7 @@ async def test_agent_serving_and_communication(test_port):
     proc = multiprocessing.Process(target=run_agent, args=(test_port,), daemon=True)
     proc.start()
     server_url = f"http://localhost:{test_port}"
+    await asyncio.sleep(3)
 
     try:
         async with httpx.AsyncClient() as httpx_client:


docs/agents/tools.md
```diff
@@ -77,14 +77,14 @@ MCP can either be run locally ([MCPStdio][any_agent.config.MCPStdio]) or you can
 
     More information about serving agents over the A2A protocol can be found [here](../serving.md)
 
-`any-agent` provides a tool to wrap a connection to another another agent served over the A2A protocol, by invoking the `any_agent.tools.a2a_tool` function, as shown in the following example:
+`any-agent` provides a tool to wrap a connection to another another agent served over the A2A protocol, by invoking the `any_agent.tools.a2a_tool` or `any_agent.tools.a2a_tool_async` function, as shown in the following example:
```
javiermtorres: Should we make a detailed account of the conditions and expectations? Like,

> When using the async functions, the user is responsible for managing the event loop. However, when using the sync functions, no event loop must have been created or started in the application.

src/any_agent/frameworks/any_agent.py
```diff
@@ -104,13 +104,23 @@ def create(
...
-                tracing=tracing,
+        try:
+            return asyncio.get_event_loop().run_until_complete(
+                cls.create_async(
+                    agent_framework=agent_framework,
+                    agent_config=agent_config,
+                    tracing=tracing,
+                )
+            )
+        except RuntimeError:
```
javiermtorres: Is there a more specific error we can check?
Also, can we factor this out, like....

```
def exec(callable):
        try:
            return asyncio.get_event_loop().run_until_complete(
                callable()
            )
        except RuntimeError:
            # No running event loop, we can use asyncio.run()
            return asyncio.run(
                callable()
            )
```

_Also_, is there some intermediate condition that would not be covered? I can't think of any, just in case.

src/any_agent/tools/a2a.py
```diff
@@ -85,3 +86,61 @@ async def _send_query(query: str) -> str:
             The result from the A2A agent, encoded in json.
     """
     return _send_query
+
+
+def _run_async(coro: Coroutine[Any, Any, Any]) -> Any:
```
javiermtorres: Ah, here's the factoring 😁 I assume we want a single approach so we'd reuse this in the previous snippets, right?

src/any_agent/tools/a2a.py
```diff
@@ -85,3 +86,61 @@ async def _send_query(query: str) -> str:
...
+        # If we get here, there IS a loop running, so we can't use asyncio.run()
+        # directly because it would try to create a second loop in the same thread
+        import concurrent.futures
+
+        def run_in_thread() -> Any:
+            # This runs in a fresh thread with no event loop, so asyncio.run() is safe
+            return asyncio.run(coro)
+
+        # Execute the async code in a separate thread to avoid loop conflicts
+        with concurrent.futures.ThreadPoolExecutor() as executor:
```
javiermtorres: Do you think it's worth it to define a package-level executor instead of creating a new one for each invocation?

tests/conftest.py
```diff
@@ -13,8 +14,17 @@
 from any_agent.logging import setup_logger
 from any_agent.tracing.agent_trace import AgentSpan, AgentTrace
 
-BASE_PORT = 5800
-PORT_PER_FRAMEWORK = {fw: BASE_PORT + index for index, fw in enumerate(AgentFramework)}
+
+def find_free_port() -> int:
```
javiermtorres: I'd prefer a fixed assignment :-/ If other stuff goes on in the background, ports may appear free but get occupied while being written down; remember that we're doing parallel testing. A fixed, predictable, _boring_ assignment ensures that net fails will not happen unpredictably.

...or at least until we have more suites with servers, so we'll need to define several ranges and/or make test dependencies. But we aren't there yet.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -160,3 +165,150 @@ def get_datetime() -> str:
...
+    kwargs = {}
+
+    kwargs["model_id"] = "gpt-4.1-nano"
+    agent_model = kwargs["model_id"]
+    env_check = validate_environment(kwargs["model_id"])
+    if not env_check["keys_in_environment"]:
+        pytest.skip(f"{env_check['missing_keys']} needed for {agent_framework}")
+
+    server_process = None
+    sync_port = tool_agent_port + 1  # Use different port to avoid conflicts
```
javiermtorres: Fixed assignment. We'll run into flaky tests otherwise.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -160,3 +165,150 @@ def get_datetime() -> str:
...
+        )
+        server_process.start()
+
+        # Poll until server is ready (max 5 seconds)
+        server_url = f"http://localhost:{sync_port}/{tool_agent_endpoint}"
+        max_attempts = 10
+        poll_interval = 0.5
+        attempts = 0
+
+        while True:
```
javiermtorres: Factor out. Also, don't be so specific about timeouts, status codes, etc (next time, no need to remove it now). This is just a health probe.

javiermtorres: Leave a bit more than 5 seconds, imho.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -160,3 +165,150 @@ def get_datetime() -> str:
...
+        )
+        assert any(
+            span.is_tool_execution()
+            and span.attributes.get("gen_ai.tool.name", None) == "call_date_agent"
+            for span in agent_trace.spans
+        )
+
+    finally:
+        if server_process and server_process.is_alive():
+            # Send SIGTERM for graceful shutdown
```
javiermtorres, line 300: IMHO it's not necessary, we're closing the test anyway

docs/agents/tools.md
njbrake: Great catch, I forgot to update the rest of this readme, now that we have a sync call we don't need to do this wrapping of the async call. I'll update the docs to clean this up. I think it's better to suggest people to use the async calls so I'll keep this as await syntax.

src/any_agent/frameworks/any_agent.py
njbrake: Good call, I will factor this wrapper out so that everybody shares it

src/any_agent/tools/a2a.py
njbrake: For simplicity of avoiding a shared state, I'd like to create new ones for each invocation at the moment. This is not the most performant solution: however, in a production server (aka putting it into FastAPI) everyone would be using the async calls so this code wouldn't be used.

tests/conftest.py
njbrake: No problem. You caught me on scope creep for this PR 😆  I'll undo this part, and let's continue the discussion or brainstorm alternate approaches in a different place.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -23,6 +28,48 @@
 logger.setLevel(logging.DEBUG)
 
 
+# These need to be two separate ranges because when running pytest in parallel, the two tests may run at the same time
+# and try to bind to the same ports if we use the same range.
+BASE_PORT = 5800
```
javiermtorres: The dynamic port allocation was definitely a good idea. Let's address it in a separate PR 👍🏼

tests/conftest.py
njbrake: I used the fixed assignment but then when I was factoring in the extra tests in serve_agent, I used Claude for an assist. And it was like "the best way to do this would be in a dynamic but deterministic way, here's the code". I took a look at it and I'm happy with that approach so I'll implement it in this PR and we can iterate to use a different method if you don't like it's use of hashing (which is how it makes it deterministic)

PR #432, opened 2025-06-05
Return spans in exception if the invocation fails

javiermtorres:
If the invocation fails due to some error that is not caught by the underlying platform, an `AgentRunException` will be thrown, having as `__cause__` the original exception, and keeping the created spans so far in the `spans` property. Note that many platforms will catch a large amount of errors in tools and wrap them in their own spans, with varying status codes.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,7 +7,7 @@ from typing import Any
 from unittest.mock import AsyncMock, patch
 
 import pytest
-from litellm.types.utils import ModelResponse
+from litellm.types.utils import ModelResponse as LLMModelResponse
 
 from any_agent.config import AgentFramework
 from any_agent.logging import setup_logger
@@ -15,6 +15,15 @@ from any_agent.tracing.agent_trace import AgentSpan, AgentTrace
 
 BASE_PORT = 5800
 PORT_PER_FRAMEWORK = {fw: BASE_PORT + index for index, fw in enumerate(AgentFramework)}
+PATCH_PER_FRAMEWORK = {
+    AgentFramework.AGNO: "agno.tools.function.FunctionCall.execute",
+    AgentFramework.GOOGLE: "google.adk.tools.function_tool.FunctionTool.run_async",
+    AgentFramework.LANGCHAIN: "langchain_core.tools.structured.StructuredTool._run",
+    AgentFramework.LLAMA_INDEX: "llama_index.core.tools.function_tool.sync_to_async",
+    AgentFramework.OPENAI: "langchain_core.tools.structured.StructuredTool._run",
+    AgentFramework.SMOLAGENTS: "smolagents.agents.ToolCallingAgent.execute_tool_call",
+    AgentFramework.TINYAGENT: "any_agent.frameworks.tinyagent.ToolExecutor.call_tool",
+}
 
 
 @pytest.fixture(params=list(AgentFramework), ids=lambda x: x.name)
@@ -25,6 +34,11 @@ def agent_framework(request: pytest.FixtureRequest) -> AgentFramework:
 @pytest.fixture
 def tool_agent_port(agent_framework):
     return PORT_PER_FRAMEWORK[agent_framework]
+
+
+@pytest.fixture
+def patched_function(agent_framework):
+    return PATCH_PER_FRAMEWORK[agent_framework]
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -100,9 +114,9 @@ def configure_logging(pytestconfig: pytest.Config) -> None:
 
 
 @pytest.fixture
-def mock_litellm_response() -> ModelResponse:
+def mock_litellm_response() -> LLMModelResponse:
```
javiermtorres: Hum,, this shouldn't be here. Let me push again.

javiermtorres:
Ok, but there might be missing information. I'll make a first try at it.

javiermtorres:
The integration tests added in this PR work ok:

```
tests/integration/test_load_and_run_agent.py::test_exception_trace[LANGCHAIN] 
[gw1] [  1%] SKIPPED tests/integration/test_load_and_run_agent.py::test_exception_trace[LANGCHAIN] 
tests/integration/test_load_and_run_agent.py::test_exception_trace[SMOLAGENTS] 
[gw3] [  2%] PASSED tests/integration/test_load_and_run_agent.py::test_exception_trace[SMOLAGENTS] 
tests/integration/test_load_and_run_agent.py::test_exception_trace[LLAMA_INDEX] 
[gw1] [  4%] PASSED tests/integration/test_load_and_run_agent.py::test_exception_trace[LLAMA_INDEX] 
tests/integration/test_load_and_run_agent.py::test_exception_trace[TINYAGENT] 
[gw3] [  5%] PASSED tests/integration/test_load_and_run_agent.py::test_exception_trace[TINYAGENT] 
tests/integration/test_load_and_run_agent.py::test_exception_trace[OPENAI] 
[gw1] [  8%] PASSED tests/integration/test_load_and_run_agent.py::test_exception_trace[OPENAI] 
tests/integration/test_load_and_run_agent.py::test_exception_trace[AGNO] 
[gw1] [ 13%] PASSED tests/integration/test_load_and_run_agent.py::test_exception_trace[AGNO] 
ts/integration/test_load_and_run_agent.py::test_exception_trace[GOOGLE] 
[gw2] [ 76%] PASSED tests/integration/test_load_and_run_agent.py::test_exception_trace[GOOGLE] 
```

javiermtorres:
> is there a way that the functionality can be tested via a unit test or an existing integration test? We're starting to have a growing number of integration tests, and this functionality seems like it could be tested via a unit test instead of integration?

Let's move this to a separate PR. I think using a unit test is a good idea, but let's not block this functinoality until we have it.

tests/integration/test_load_and_run_agent.py
```diff
@@ -223,3 +231,88 @@ def write_file(text: str) -> None:
         assert_eval(agent_trace)
     finally:
         agent.exit()
+
+
+@pytest.mark.skipif(
+    os.environ.get("ANY_AGENT_INTEGRATION_TESTS", "FALSE").upper() != "TRUE",
+    reason="Integration tests require `ANY_AGENT_INTEGRATION_TESTS=TRUE` env var",
+)
+def test_exception_trace(
```
daavoo, line 240: I would put this in a separate file

tests/integration/test_load_and_run_agent.py
```diff
@@ -223,3 +231,88 @@ def write_file(text: str) -> None:
...
+    agent_config = AgentConfig(
+        tools=tools,  # type: ignore[arg-type]
+        instructions="Answer the query.",
+        model_args=model_args,
+        **kwargs,  # type: ignore[arg-type]
+    )
+    agent = AnyAgent.create(agent_framework, agent_config, tracing=TracingConfig())
+    update_trace = request.config.getoption("--update-trace-assets")
+    if update_trace:
+        agent._exporter.console.record = True  # type: ignore[union-attr]
```
daavoo, line 300: This is only needed if we are actually updating the traces as part of the test

tests/integration/test_load_and_run_agent.py
```diff
@@ -223,3 +231,88 @@ def write_file(text: str) -> None:
...
+    try:
+        with patch(patched_function) as fw_agent_runtool:
+            fw_agent_runtool.side_effect = RuntimeError(exc_reason)
+            spans = []
+            try:
+                agent.run(
+                    "Write a four-line poem and use the tools to write it to a file.",
+                )
+            except AgentRunError as are:
+                spans = are.trace.spans
```
daavoo, line 311: We should add this expected usage to the user docs

tests/integration/test_load_and_run_agent.py
javiermtorres: Ok, separate issue 👍

tests/integration/test_load_and_run_agent.py
javiermtorres: Opening a new issue for this 👍

PR #435, opened 2025-06-06
refactor(tracing): Instrument at the instance level.

daavoo:
- Decouple `AgentTrace` generation from exporter.
- Create a single global exporter. No config.
- Drop `AnyAgent.exit`. instrument/uninstrument happens within `AnyAgent.run_async`.
- Add `instrument` argument allowing to force skip on `run` calls (use that in eval).

---

@njbrake @javiermtorres I intentionally went a little aggressive on changing/removing things. Happy to hear any push backs


src/any_agent/frameworks/any_agent.py
```diff
@@ -135,40 +123,46 @@ async def _load_tools(
...
+        self, prompt: str, instrument: bool = True, **kwargs: Any
+    ) -> AgentTrace:
         """Run the agent asynchronously with the given prompt."""
-        run_id = str(uuid4())
         with self._tracer.start_as_current_span(
             f"invoke_agent [{self.config.name}]"
         ) as invoke_span:
+            if instrument and self._instrumentor:
+                trace_id = invoke_span.get_span_context().trace_id
+                async with self._lock:
```
daavoo: This is to handle the scenario like:

https://github.com/mozilla-ai/any-agent/blob/ec97de5dbdfbc0c08aaf105344d57e7efc5c4244/tests/integration/test_load_and_run_agent_twice.py#L30-L33

src/any_agent/tracing/__init__.py
```diff
@@ -0,0 +1,19 @@
+from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+
+from .exporter import _ConsoleExporter
+
+TRACE_PROVIDER = TracerProvider()
+trace.set_tracer_provider(TRACE_PROVIDER)
+TRACE_PROVIDER.add_span_processor(SimpleSpanProcessor(_ConsoleExporter()))
```
daavoo: Decided to just add a single global console exporter. Adding the helper function below for people to be able to disable console outputs

javiermtorres:
I don't have the use case totally clear. Is this instrumentation then active just within the scope of run? Is the pipeline configurable, or are we fixing it to the single exporter? Are the spans received always with the result of the run?

src/any_agent/tracing/exporter.py
```diff
@@ -36,32 +33,21 @@ def _get_output_panel(span: AgentSpan) -> Panel | None:
...
-        self.run_trace_mapping: dict[str, int] = {}
+class _ConsoleExporter(SpanExporter):
+    def __init__(self) -> None:
+        self.console: Console = Console()
 
-        if self.tracing_config.console:
-            self.console = Console()
+    def print_to_console(self, span: ReadableSpan) -> None:
+        if scope := span.instrumentation_scope:
+            if scope.name != "any_agent":
```
njbrake: 💅  but might be good to elevate this string variable so that its clear what piece of the code is creating this string that this function relies on.

src/any_agent/tracing/instrumentation/agno.py
```diff
@@ -76,18 +77,24 @@ def __init__(self) -> None:
...
 
-    def instrument(self, tracer: Tracer) -> None:
-        async def wrap_aprocess_model_response(  # type: ignore[no-untyped-def]
-            wrapped,
-            instance: Model,
-            args,
-            kwargs,
+    def instrument(self, agent: AgnoAgent) -> None:
+        if len(agent._running_traces) > 1:
+            return
```
njbrake, line 82: Take it or leave it but it may be worth explaining why this if statement is important and how it relies on the current any_agent locking design. Or, correct my comment if I misunderstand why this is important 😆 
```suggestion
        if len(agent._running_traces) > 1:
            # If the agent is already instrumented, no need to re-instrument
            # This relies upon the lock in any-agent to enforce that this happens one at a time in a running agent
            return
```

tests/integration/test_load_and_run_agent_twice.py
```diff
@@ -11,7 +11,7 @@
 @pytest.mark.asyncio
 async def test_run_agent_twice(agent_framework: AgentFramework) -> None:
     """When an agent is run twice, state from the first run shouldn't bleed into the second run"""
-    model_id = "gpt-4.1-nano"
+    model_id = "gpt-4.1-mini"
```
njbrake: 💅  why not keep as nano?

src/any_agent/frameworks/any_agent.py
```diff
@@ -135,40 +123,46 @@ async def _load_tools(
...
             final_output = await self._run_async(prompt, **kwargs)
-        trace = self._exporter.pop_trace(run_id)  # type: ignore[union-attr]
+
+            trace = AgentTrace()
+            if instrument and self._instrumentor:
+                async with self._lock:
+                    self._instrumentor.uninstrument(self)  # type: ignore[arg-type]
+                    trace = self._running_traces.pop(trace_id)
+
+        trace.add_span(invoke_span)
```
njbrake: So, if I understand this right: if you set instrument=False, you only get the single span and not the full trace? This logic is unclear and could benefit from a few comments maybe.

docs/tracing.md
njbrake, line 1: This is missing documentation about how to turn the console output on vs off, I think?

src/any_agent/frameworks/any_agent.py
daavoo: I have explained the logic as part of the `instrument` arg section in the docstring. Do you think that is enough?

daavoo:
@njbrake 

> I would be curious about how much of a performance hit this is, but I feel like that's probably more a "it is what it is" situation

It is mostly calling `setattr` so I would expect it to not be much. I have ran it with https://github.com/joerick/pyinstrument and seems negligible.

> In terms of whether the implementation works, if our integration and cookbook tests pass, that's good enough for me. This PR drops several features of the tracing module like setting colors and such, but I never used or cared about those settings anyways, the only thing I care about is being able to turn the console outputs on and off.

Integration: https://github.com/mozilla-ai/any-agent/actions/runs/15536621721
Cookbook: https://github.com/mozilla-ai/any-agent/actions/runs/15533228256

In addition, I ran the notebooks locally and don't see duplicated spans (but I am not sure how test that programmatically)

daavoo:
@javiermtorres 

> I don't have the use case totally clear. Is this instrumentation then active just within the scope of run?

That is correct. This is the cleanest way I found to avoid the linked issues.

>  Is the pipeline configurable, or are we fixing it to the single exporter? 

We are adding a single exporter by default. People can import `from any_agent.tracing import TRACE_PROVIDER` and add/remove any exporters.

> Are the spans received always with the result of the run?

Not sure what you mean here. The internal span generation is now independent from the exporter

src/any_agent/tracing/instrumentation/agno.py
```diff
@@ -79,6 +79,8 @@ def __init__(self) -> None:
 
     def instrument(self, agent: AgnoAgent) -> None:
         if len(agent._running_traces) > 1:
+            # If the agent is already instrumented, no need to re-instrument
+            # This relies upon the lock in any-agent to enforce that this happens one at a time in a running agent
             return
```
njbrake: I neglected to mentionin my original commend that this comment would need to be propogated to all the frameworks, or elevated to a base class comment or something

daavoo: Right! Moved the comment up and rephrased it a bit

src/any_agent/frameworks/tinyagent.py
```diff
@@ -167,7 +166,7 @@ async def _load_agent(self) -> None:
             logger.debug("Registered tool: %s", tool_name)
 
     async def _run_async(self, prompt: str, **kwargs: Any) -> str | BaseModel:
-        self.messages = [
```
njbrake, line 178: Good catch. I assume this is fixing an issue with running two agents in parallel

src/any_agent/tracing/__init__.py
```diff
@@ -1,11 +1,14 @@
...
 from opentelemetry.sdk.trace.export import SimpleSpanProcessor
 
 from .exporter import _ConsoleExporter
 
-TRACE_PROVIDER = TracerProvider()
-trace.set_tracer_provider(TRACE_PROVIDER)
+TRACE_PROVIDER = trace.get_tracer_provider()
+if isinstance(TRACE_PROVIDER, trace.ProxyTracerProvider):
+    TRACE_PROVIDER = TracerProvider()
+    trace.set_tracer_provider(TRACE_PROVIDER)
```
njbrake, line 11: A comment explaining what and why this is would be helpful for those that don't know OTEL. Unless this is an obvious feature in OTEL that I should know about, in which case disregard and no comment required

tests/integration/test_load_and_run_agent_twice.py
```diff
@@ -6,13 +6,28 @@
 
 from any_agent import AgentConfig, AgentFramework, AnyAgent
 
-from .helpers import mock_search_web
```
njbrake, line 9: In that case probably delete the mock_search_web from the helpers since no one else is using that

src/any_agent/frameworks/tinyagent.py
daavoo: Correct

src/any_agent/tracing/__init__.py
daavoo: > obvious feature in OTEL 

Wouldn't call it obvious but it is how I understood things should be done to either respect existing trace provider or create a default one

PR #442, opened 2025-06-06
Dynamic port assignment in async tests

javiermtorres:
The port can be left to 0, so the OS will assign any to the server. However, the agentcard needs to be updated since its url will have port=0.

### DIFF
--- a/src/any_agent/serving/server.py
+++ b/src/any_agent/serving/server.py
@@ -61,6 +61,9 @@ async def serve_a2a_async(
     task = asyncio.create_task(uv_server.serve())
     while not uv_server.started:  # noqa: ASYNC110
         await asyncio.sleep(0.1)
+    if port == 0:
+        server_port = uv_server.servers[0].sockets[0].getsockname()[1]
+        server.agent_card.url = f"http://{host}:{server_port}/{endpoint.lstrip("/")}"
     return (task, uv_server)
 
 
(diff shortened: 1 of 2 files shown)


njbrake (commented):
I think this direction is fine, only comment would be that if you use the dynamic port allocation feature, lets use it everywhere in the tests and get rid of the test_port fixture. I know this PR is in draft and you are probably planning to do that anyways :) 💯

njbrake (approved):
Thanks for your work on this. I'll offer one other design option for you to consider: what if you bound to a socket in the any_agent serve function and then passed that socket in to the underlying serve functions? That way the AnyAgent.serve_async handles all the logic around picking the right port, and the underlying functions aren't responsible for creating and parsing the port.

This would also make it so that when we add support for MCP/ACP etc, you won't need to re-implement the port selection and parsing logic (even though it's only a few lines of code so no big deal).

Feel free to take this or leave it, since what you have here I think is also good.

Screenshot 2025-06-09 at 12 02 19 PM

njbrake:
@javiermtorres  oh one more comment, you can probably delete the test_port fixture altogether before you merge this.

javiermtorres:
We'd need to set the sockets in the part where the uvicorn server is created:
image
But, otoh, it's better than setting the port to 0 and retrieving the port from the uvicorn internals, given that uvicorn allows us to provide the serving sockets. It may even give me the possibility to arrange the port info for the agentcard beforehand instead of retrofitting it. I'll give it a try tomorrow. I'll also try to arrange things so that Hypercorn can be easily tested later on, since we're isolating our code more and more from the underlying ASGI server.

javiermtorres:
> Thanks for your work on this. I'll offer one other design option for you to consider: 

Let's move this work to a follow up PR. The port assignment works fine at the moment, and moving from port tweaking to socket tweaking will not be a big issue.

PR #458, opened 2025-06-11
A2A Server passes back a task instead of a Message

njbrake:
Functionally, the user sees no change: the API does not change, and the new client code (which handles both tasks and messages) is compatible with the old server code (that only handles messages).

By passing tasks instead of messages, we have advanced flexibility about content and can expressly provide task status back to the client so they can know when more input is required.

This is a step towards supporting true muliturn functionality in any-agent.

In order to make this work, I leverage structured output by wrapping whatever structured output type is configured into an envelope so that the agent can set the taskstate alongside the original pydantic model when it responds.

For how long it took me to figure out how to do this, I'm pretty satisfied with how little code was needed to implement it 🎉


src/any_agent/serving/server.py
```diff
@@ -26,6 +29,9 @@
 def _get_a2a_app(
     agent: AnyAgent, serving_config: A2AServingConfig
 ) -> A2AStarletteApplication:
+    # Ensure the agent is A2A-ready (task_status+data envelope)
+    agent = prepare_agent_for_a2a(agent)
```
njbrake: This is the sort of sketchy part that I'd be interested in feedback on. In order to handle structured outputs I re-create the agent. I am still thinking about whether there is a better way (like, if a developer wants to serve do we assert that they use an `output_type` that inherits from our class, or something like that, instead of trying to handle it for them on the fly).

src/any_agent/serving/envelope.py
njbrake, line 1: We only need to do this if we aren't forcing the developer to conform their output schema to anything in particular.

src/any_agent/serving/envelope.py
```diff
@@ -0,0 +1,80 @@
...
+    return OutputContainer
+
+
+def prepare_agent_for_a2a(agent: AnyAgent) -> AnyAgent:
+    """Return an agent whose ``config.output_type`` is A2A-ready.
+
+    If *agent* is already envelope-compatible we hand it back untouched.
+    Otherwise we clone its config, wrap the output type, and spin up a
+    *new* agent instance via `AnyAgent.create` so that framework-specific
+    initialisation sees the correct schema right from the start.
```
njbrake, line 68: I collaborated with OpenAI `o3` to generate the comments, interesting to observe how it has a different style from Claude 4 (which I normally use)

javiermtorres: Does it relate to MD? I got some comments about newer releases using MD when generating text even if unrequested, from a friend of mine working in the area.

src/any_agent/serving/agent_executor.py
```diff
@@ -22,10 +48,49 @@ async def execute(  # type: ignore[misc]
...
+        if not isinstance(final_output, BaseModel):
+            msg = f"Expected BaseModel, got {type(final_output)}, {final_output}"
+            raise TypeError(msg)
+
+        # Runtime attributes guaranteed by the dynamically created model.
+        if not (hasattr(final_output, "task_status") and hasattr(final_output, "data")):
+            msg = "Final output must have `task_status` and `data` attributes"
+            raise AttributeError(msg)
+
+        envelope = cast("_OutputEnvelope", final_output)
```
javiermtorres: Note that turn taking, envelopes, etc, could be handled by the executor and not the agent. Specially since the context itself belongs to the executor already. So the agent need not generate envelopes that are specific for this use case.

If, OTOH, we want to make agents that use a state/context right from the start, then let's come up with a generic way of keeping track of a session and session state that is here adapted to the A2A SDK specifics.

Otherwise, it feels like a single consistent responsibility is spread across several pieces of code.

src/any_agent/serving/server.py
javiermtorres: On one hand, re-creating an agent is cheap, and possibly quite stateless. But one might be mislead to think that whatever state was there in past requests is being carried over to the new agent.

If an agent is fairly stateless in itself until we add some context as part of the executor, then an existing agent should be "socketable" into some slot in an agent executor. IMHO the contract about the agent results (free form text vs structured outputs) should be clear from the start.

njbrake:
@daavoo  and I had a good chat today about design options, and we decided to initially pursue the implementation that leaves the API untouched.

In the future we may consider making .serve() a static class method of AnyAgent instead of a member function, but we'll keep it as is for now.

Integration tests passed, ready for merge https://github.com/mozilla-ai/any-agent/actions/runs/15616268506 🫡

njbrake:
> Either here or in a follow-up (if needed):
> 
>     * Let's update docs/cookbooks.

For this PR, I don't think there's anything to update: the change is an internal architecture change but the user won't know any difference, maybe I misunderstand?

>     * Add an integration test or extend the checks for this new behavior.

I believe I have these changes properly tested via the unit tests that I added (which make sure the a2a_tool can handle both servers that respond with tasks (aka our served agents) but still also servers which respond with messages (which other A2A servers may do).

We already have integration tests which exercise the A2A functionality so I think we're ok there too. Let me know if there was something specific you wanted tested that isn't covered, totally possible that I'm forgetting some scenario. 🎉

njbrake:
Integration tests passing https://github.com/mozilla-ai/any-agent/actions/runs/15632412391/job/44039990045

PR #461, opened 2025-06-12
Implementation option 2 for A2A Server passes back a task instead of a Message

njbrake:
See https://github.com/mozilla-ai/any-agent/pull/458 for the other option


src/any_agent/serving/envelope.py
```diff
@@ -0,0 +1,53 @@
...
+            AgentConfig(
+                model_id="gpt-4",
+                output_type=A2AEnvelope[MyAgentOutput]
+            )
+        )
+        ```
+
+    """
+
+    task_status: TaskState
```
njbrake, line 31: This is consistent with how they get an agent to output a task status in the official a2a-samples https://github.com/google-a2a/a2a-samples/blob/main/samples/python/agents/airbnb_planner_multiagent/airbnb_agent/airbnb_agent.py#L210. 

Structure outputs from the agent then are translated into A2A TaskStates and passed back to the client

njbrake:
Closed in favor of option 1

PR #465, opened 2025-06-12
refactor(tests): No logs generated by tests

njbrake:
I'd like to propose the removal of logging in the tests. When running tests I was finding log statements being generated that were distracting from the library logs: LiteLLM debug logging was turned on and littering the logs. If debugging tests is needed, I would prefer to rely upon debugger (debugpy) instead of log statements.

- Removed the `build_tree` function and its associated `CHILD_TAG` constant from `conftest.py`.
- Cleaned up logging setup in `test_load_and_run_multi_agent_a2a_tool.py`, removing unnecessary logger info calls.
- Simplified imports in `conftest.py` by removing `AgentSpan` from imports.


tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -164,16 +146,6 @@ def get_datetime() -> str:
         _assert_contains_current_date_info(agent_trace.final_output)
         _assert_has_date_agent_tool_call(agent_trace)
 
-        try:
```
javiermtorres, line 162: Remove the printouts, but please don't remove the exception check. We need to check that the span tree is being generated correctly.

javiermtorres:
I agree that these specific logs are not too useful (but please don't
remove the check on the generated spans). However, in my experience it is
good to have some context in the logs when tests don't work as expected
during regressions. Pytest will silence any passing tests anyway.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
njbrake: Can you explain more about why we need to check that the span tree is correctly generated? I'm confused about what it is checking that isn't handled by pydantic built-in validation when an AgentTrace object is created.

javiermtorres: Sometimes the parent span id is not correctly set in a child span. A tool inspecting the whole trace will not be able to present a consistent view of the run. That's a property of the whole trace, not of an independent span.

njbrake: Thanks! Do you have more info about when this would happen? Is this something that we would want to validate as a Pydantic validator attached to the Pydantic model? 

If it's something that can go wrong (especially if it doesn't always go wrong), I'm thinking that maybe this check should happen in our application code, not an integration test.

javiermtorres: If I knew why this would happen I wouldn't have written this check :D

Unless a validator would check a list of spans, I don't see this as something to be validated at runtime. It would also be too expensive to check in every run.

We could actually mock the llm (even in other integration tests) to get into the unit test realm and check the span construction specifically 🤔

njbrake: Yes, I'm suggesting we write a validator that checks.the list of spans :) Or we add some sort of check to the `add_span` function.

javiermtorres: Hum. I think that writing a validator to check this every time is an unnecessary burden on the "main track". It's up to us, so no user input will modify this process, and once we get it right it'll be right until we change the code. I still think a regression test is the best way to handle this issue (which we still need to look into, afaict???? or did we already?).

tests/conftest.py
```diff
@@ -178,29 +178,3 @@ def agent_trace(request: pytest.FixtureRequest) -> AgentTrace:
...
-            if parent_k:
-                try:
-                    traces[parent_k].attributes[CHILD_TAG][trace.context.span_id] = (
-                        trace
-                    )
-                except KeyError:
-                    pass
-            else:
-                traces[None] = trace
-    return traces[None]
```
njbrake, line 206: @javiermtorres  I'll pause on our discussion on the other thread in this PR to discuss this function 😆 , since reading through It I'm now thinking maybe it is not checking what we think it's checking? A few things:

1. the variables `traces` and `trace` are misnamed since these are spans, not traces.
2.  Key errors are being caught and ignored, so this isn't going to catch any issues, it's purely for logging, right?

WDYT?

javiermtorres: I've always said traces in 25 years of sw dev in telco equipment, I hope you understand that my head slips a bit XD Yes, they'd be spans.

Hum, I thought I had removed the try part here 🤔 The idea is exactly to get a key error if the parent span cannot be found (there's a key error caught in the main test). So this would do it:

```
def build_tree(items: list[AgentSpan]) -> AgentSpan:
    """Structures the spans into a tree."""
    traces = {}
    for trace in items:
        k = trace.context.span_id
        trace.attributes[CHILD_TAG] = {}
        traces[k] = trace
    for trace in items:
        k = trace.context.span_id
        if trace.parent:
            parent_k = trace.parent.span_id
            if parent_k:
                traces[parent_k].attributes[CHILD_TAG][trace.context.span_id] = (
                    trace
                )
            else:
                traces[None] = trace
    return traces[None]
```

njbrake: No worries 😄  I'll merge this PR but will turn this comment thread into a new issue so that we can an `assert` style function that we can add into integration tests to catch this behavior if it becomes a concern.

javiermtorres: Well, np having it as an issue, but it _is_ already a concern. So we should add it back to check that the Google adapter (at least) works ok.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
daavoo: How the tracing currently works / should work in the a2a as a tool scenario needs to be properly investigated (and explicitly tested).

PR #468, opened 2025-06-12
feat(agent): Add resume_from functionality to agent run methods

njbrake:
I'd like to open this up for discussion before I get too deep.

I see two paths:

1. Leverage each existing frameworks "resume" functionality, if it exists. Pros: maybe there is some caching they do to lower latency or cost of making a big initial LLM call? I would need to dig into each framework to see if that is happening.

2. Write our own wrapper that takes the last AgentTrace, extracts the previous messages, and prepends it to their prompt (This is the approach I went with in this PR)

Generally I like path 2, because this makes it pretty simple to handle across all the frameworks, but it certainly breaks any message caching that frameworks might have used in LiteLLM https://docs.litellm.ai/docs/proxy/caching

So, maybe for right now we could implement this as a sort of "brute force" approach but we could refine it as we go, and probably could do internally so that it wouldn't require any future changes to the API after this initial addition of the `resume_from` parameter.

I'm also adding @HareeshBahuleyan  and @stefanfrench  here as the conclusions we reach here may help the agent-factory design (if we do this PR right 😅 )

### DIFF
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 from abc import ABC, abstractmethod
+import json
 from typing import TYPE_CHECKING, Any, assert_never
 
 from opentelemetry import trace
@@ -140,17 +141,18 @@ class AnyAgent(ABC):
             tools.extend(mcp_server.tools)
         return tools, mcp_servers
 
-    def run(self, prompt: str, **kwargs: Any) -> AgentTrace:
+    def run(self, prompt: str, resume_from: AgentTrace | None = None, **kwargs: Any) -> AgentTrace:
         """Run the agent with the given prompt."""
-        return run_async_in_sync(self.run_async(prompt, **kwargs))
+        return run_async_in_sync(self.run_async(prompt=prompt, resume_from=resume_from, **kwargs))
 
     async def run_async(
-        self, prompt: str, instrument: bool = True, **kwargs: Any
+        self, prompt: str, instrument: bool = True, resume_from: AgentTrace | None = None, **kwargs: Any
     ) -> AgentTrace:
         """Run the agent asynchronously with the given prompt.
 
         Args:
             prompt: The user prompt to be passed to the agent.
+            resume_from: The `AgentTrace` to resume from.
             instrument: Whether to instrument the underlying framework
                 to generate LLM Calls and Tool Execution Spans.
 
@@ -165,6 +167,12 @@ class AnyAgent(ABC):
                 steps taken by the agent.
 
         """
+        if resume_from is not None:
+            # Construct a "history" string that can be prepended to the prompt
+            history = "\n".join([json.dumps(span.attributes) for span in resume_from.spans if span.attributes is not None])
+            prompt = f"History:\n{history}\nPrompt:\n{prompt}"
+            print(f"Resuming from history: {history}")
+
         try:
             trace = AgentTrace()
             with self._tracer.start_as_current_span(
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
 import asyncio
-from abc import ABC, abstractmethod
 import json
+from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any, assert_never
 
 from opentelemetry import trace
@@ -141,12 +141,20 @@ class AnyAgent(ABC):
             tools.extend(mcp_server.tools)
         return tools, mcp_servers
 
-    def run(self, prompt: str, resume_from: AgentTrace | None = None, **kwargs: Any) -> AgentTrace:
+    def run(
+        self, prompt: str, resume_from: AgentTrace | None = None, **kwargs: Any
+    ) -> AgentTrace:
         """Run the agent with the given prompt."""
-        return run_async_in_sync(self.run_async(prompt=prompt, resume_from=resume_from, **kwargs))
+        return run_async_in_sync(
+            self.run_async(prompt=prompt, resume_from=resume_from, **kwargs)
+        )
 
     async def run_async(
-        self, prompt: str, instrument: bool = True, resume_from: AgentTrace | None = None, **kwargs: Any
+        self,
+        prompt: str,
+        instrument: bool = True,
+        resume_from: AgentTrace | None = None,
+        **kwargs: Any,
     ) -> AgentTrace:
         """Run the agent asynchronously with the given prompt.
 
@@ -169,9 +177,14 @@ class AnyAgent(ABC):
         """
         if resume_from is not None:
             # Construct a "history" string that can be prepended to the prompt
-            history = "\n".join([json.dumps(span.attributes) for span in resume_from.spans if span.attributes is not None])
+            history = "\n".join(
+                [
+                    json.dumps(span.attributes)
+                    for span in resume_from.spans
+                    if span.attributes is not None
+                ]
+            )
             prompt = f"History:\n{history}\nPrompt:\n{prompt}"
-            print(f"Resuming from history: {history}")
 
         try:
             trace = AgentTrace()
--- a/tests/integration/test_load_and_run_agent.py
+++ b/tests/integration/test_load_and_run_agent.py
@@ -5,11 +5,11 @@ import time
 from datetime import datetime, timedelta
 from pathlib import Path
 from typing import Any
-from pydantic import BaseModel
 from unittest.mock import patch
 
 import pytest
 from litellm.utils import validate_environment
+from pydantic import BaseModel
 
 from any_agent import (
     AgentConfig,
@@ -314,7 +314,9 @@ def test_exception_trace(
 def test_resume_from_feature(agent_framework: AgentFramework, tmp_path: Path) -> None:
     """Test that the agent can resume from a previous trace using the resume_from argument."""
     if agent_framework is AgentFramework.GOOGLE:
-        pytest.skip("Google infinite recursion bug: https://github.com/mozilla-ai/any-agent/issues/467 prevents this test from passing")
+        pytest.skip(
+            "Google infinite recursion bug: https://github.com/mozilla-ai/any-agent/issues/467 prevents this test from passing"
+        )
     kwargs = {}
     kwargs["model_id"] = "gpt-4.1-mini"
     env_check = validate_environment(kwargs["model_id"])
@@ -335,7 +337,5 @@ def test_resume_from_feature(agent_framework: AgentFramework, tmp_path: Path) ->
     )
     assert trace1.final_output.year == 2000
 
-    trace2 = agent.run(
-        "Now increment the year by 5.", resume_from=trace1
-    )
+    trace2 = agent.run("Now increment the year by 5.", resume_from=trace1)
     assert trace2.final_output.year == 2005
--- a/tests/integration/test_load_and_run_agent.py
+++ b/tests/integration/test_load_and_run_agent.py
@@ -5,6 +5,7 @@ import time
 from datetime import datetime, timedelta
 from pathlib import Path
 from typing import Any
+from pydantic import BaseModel
 from unittest.mock import patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


src/any_agent/frameworks/any_agent.py
```diff
@@ -165,6 +175,17 @@ async def run_async(
...
+        if resume_from is not None:
+            # Construct a "history" string that can be prepended to the prompt
+            history = "\n".join(
+                [
+                    json.dumps(span.attributes)
+                    for span in resume_from.spans
+                    if span.attributes is not None
+                ]
+            )
+            prompt = f"History:\n{history}\nPrompt:\n{prompt}"
```
njbrake: This wouldn't be the final implementation, it's just so you get an idea about basically what would happen

njbrake:
tl;dr, I am a fan of what I understand to be the design of openai/llama-index/langgraph where we can pass in a list of previous messages as context. It avoids the complexity of managing session states and agent state, and will be pretty easy to implement in all the frameworks. 

### Unsolved Problems

How do we handle the trace: if a user is trying to do multi turn, Do we provide some method to link the two trace_ids together? I'm thinking maybe we can edit the function https://github.com/mozilla-ai/any-agent/blob/main/src/any_agent/frameworks/any_agent.py#L170 to optionally set a `links` or `context` parameter of the `start_as_current_span` function to link the two traces together that way, so that anyone monitoring agents in an OTEL collector will be able to see that the new message is a continuation of the old conversation.


## What frameworks do for Multiturn Chat 

This is what happens after an agent.run has finished and a final output has been returned to the user

### OpenAI

They have a method for creating a new input using the past run with the https://openai.github.io/openai-agents-python/ref/result/#agents.result.RunResultBase.to_input_list helper function

See their example https://openai.github.io/openai-agents-python/running_agents/#conversationschat-threads

Their typing for the ‘input’ var of the run function is `input: str | list[TResponseInputItem]`. 

So for openai, if we wanted to resume, we could reconstruct the list of ResponseInputItems.


### Agno

They use the concept of a session https://docs.agno.com/agents/sessions

https://github.com/agno-agi/agno/blob/main/libs/agno/agno/agent/agent.py#L1105 

You set the session and then pass that in again and agno takes care of restoring what happened when it returned the last time. It’s unclear to me how they manage deciding when to clean up sessions.


### Google ADK

Similar to Agno, it’s tracked via ‘session’ https://google.github.io/adk-docs/sessions/session/

### Langgraph

Looks like there are many different ways to do it, but the simplest looks like it would be to construct a list of past messages and pass it in to ainvoke. https://langchain-ai.github.io/langgraph/tutorials/get-started/1-build-basic-chatbot/. 

This is similar to the OpenAI design. 

### Llama-Index

The run method has a ‘chat_history` key where we can pass in the history directly, or manage a Context object which stores the session state. https://docs.llamaindex.ai/en/stable/getting_started/starter_example/#adding-chat-history

### Smolagents

Smolagents looks like it manages chat history as a member variable attached to the class instance. https://github.com/huggingface/smolagents/blob/v1.18.0/src/smolagents/agents.py#L354

Basically this means that one agent instantiation == one state.

Probably for this reason, smolagents doesn’t have an async API (at least, I’m not seeing one, and we don’t use one). If we wanted to be able to set `reset=False` when calling run, we would need to redesign the implementation so that each time run is called, we create a new agent and have some sort of state management to allow for clients to hook into the agent they created last run, if they need it. Could get tricky.

### Tinyagents

Blue sky, we can implement this however we want lol

javiermtorres:
> How do we handle the trace: if a user is trying to do multi turn, Do we provide some method to link the two trace_ids together? 

If this refers to otel, imho there should be a parent span representing the "session". I'd assume there's a session even in single-message interactions, since the user will often slip into multiturn conversations. Links seem to apply to async related processes.

Re: the session handling, I'd take path no. 2 for TinyAgents (but not using the agenttrace, keep this info in a proper place), _but_ at the any-agent level, so one could plug in another agent impl if necessary (or the other way around ^_^).

daavoo (commented):
I am -1 on the `history` but+1 on the span helpers. 

I don't like when frameworks try to be smart about stuff like this and then I have to figure out how it works and how to adapt/disable it for my use case.

I believe this is something users should implement on top of any-agent and let them decide how to inject information in the prompt. 

I am not against adding the helpers and documenting the pattern.

daavoo:
> How do we handle the trace: if a user is trying to do multi turn, Do we provide some method to link the two trace_ids together? 

We are just starting an `invoke_agent` span as current span during `run_async`. 
If the user owning the app manages the multiple agent invocations, they should be able to create a parent span that would contain the multiple `invoke_agent` spans.

javiermtorres:
> they should be able to create a parent span that would contain the multiple `invoke_agent` spans.

Should any-agent take care of this, or should we instruct the user to do it themselves???? I don't have a clear opinion, really :-/

njbrake:
> I am -1 on the `history` but+1 on the span helpers.
> 
@daavoo I'm hesitant to throw away the design of history as a list of messages when a framework supports the use of this.
Forcing everything to be put into the prompt will result in putting a model out of distribution which I would expect to impact performance. 

To my knowledge, when a model is instruction/chat fine-tuned, the pre-processing standardizes the conversation data into a specific format of special tokens etc (aka "[USER\]" vs "<---user--->") . The documentation by huggingface does a great job illustrating this example, they have a tokenizer method called `apply_chat_template` to overcome this very problem. [https://huggingface.co/docs/transformers/main/en/chat_templating](https://huggingface.co/docs/transformers/main/en/chat_templating)

By passing in the list of messages to LiteLLM, we allow it to utilize the prompt templates that it manages for us https://docs.litellm.ai/docs/completion/prompt_formatting, or even to pass the messages into the LLM provider directly, which the provider will take care of formatting correctly based on the specific model being inferenced.

We could say "its up to the user to properly construct the prompt based on which model they're using" but that requires the user to understand this technical detail about models, and also introduces more risk that they may not understand this and will try to do something strange which will negatively impact their results.

I would probably lean towards saying that **all** frameworks should be able to pass a list of messages, and remove my manual prompt string formulation. Let me know if my explanation makes sense or if I'm completely misunderstanding you, like I was when we were talking about structured outputs last week 😆

daavoo:
> @daavoo I'm hesitant to throw away the design of history as a list of messages when a framework supports the use of this. Forcing everything to be put into the prompt will result in putting a model out of distribution which I would expect to impact performance.

I am not saying we force anything, I am saying we let the user decide and document how to do it.

> To my knowledge, when a model is instruction/chat fine-tuned, the pre-processing standardizes the conversation data into a specific format of special tokens etc (aka "[USER]" vs "<---user--->") . The documentation by huggingface does a great job illustrating this example, they have a tokenizer method called `apply_chat_template` to overcome this very problem. https://huggingface.co/docs/transformers/main/en/chat_templating

> By passing in the list of messages to LiteLLM, we allow it to utilize the prompt templates that it manages for us https://docs.litellm.ai/docs/completion/prompt_formatting, or even to pass the messages into the LLM provider directly, which the provider will take care of formatting correctly based on the specific model being inferenced.
> 
> We could say "its up to the user to properly construct the prompt based on which model they're using" but that requires the user to understand this technical detail about models, and also introduces more risk that they may not understand this and will try to do something strange which will negatively impact their results.

As far as I understand, chat templating is just an implementation detail used to implement chatbots on top of LLMs (mainly to be able to post-train on conversations with consistent format and have a consistent "stop generating" token for the "assistant").

I seriously doubt that including those chat templating tokens when building the `history` has an actual impact on the agent performance (happy to be pointed to proves of this being wrong) compared to formatting a new prompt in whatever format you please like:

```
prompt +="""
# Previous context

Hey this is some previous context you might found useful: {history}
"""
```

What I am sure matters is the content of the `history`, and that is my reason to push back on this PR: we are deciding what information goes into and how it is formatted, in a way that is not transparent to the users.

That is why I would prefer to document helper `span` methods and different patterns to preserve context across different `run` calls. 

I wouldn't mind to document how to pass formatted messages as input to `run`, but still I would prefer someone to prove that this matters.

njbrake:
Thanks! Maybe we are talking about two different but similar design goals.

### 1 Multi-turn conversation
In this scenario a person is looking to have a continued conversation like a chatbot back and forth. The content matches the implementation and post-training of the model to use the chat template tokens to indicate the back and forth between user and agent. 

I don't think we can assume that diverging from this format has no actual impact to performance: we are then moving away from how it was post-trained. I'm not saying that it is going to start saying crazy things or go off the rails, but its hasty to assume that the behavior won't change at all. 


> I wouldn't mind to document how to pass formatted messages as input to run, but still I would prefer someone to prove that this matters.

I would prefer someone to prove that this does not matter 😆 

Note that here I am only talking about un-summarized conversations: if we're talking about summarizing or excluding certain past messages, then we are talking about the next scenario.

-----


### 2 Continued conversation with an agent that is aware of what happened in the past

Its sort of multiturn, but the burden lies on the developer to do this right. If they do a good job summarizing and/or including the relevant content, they'll get good performance and probably use less tokens than multiturn. But, if they mess something up and add or remove certain content, they may end up with unexpected results. 

I think that this is absolutely a valid scenario and one that we should even recommend or draw peoples attention to. Like, if they have an agent that's making webpage calls, their context is going to be littered with tons of tokens that they don't really need. They should summarize and filter their history before continuing, to save money by reducing token usage,  if nothing else.

-----

From my read, I am talking about supporting 1, where as I understand you are saying that we should support 2, and that 1 is not necessary. 

I'm not against dropping support for 1. However, I think if we decide to drop support for 1 in favor of 2, we should provide some strong documentation explaining this and our reasoning, as you said. By dropping support for 1, we force the developer to design their history implementation, they're not given the option to wholly rely on LLM post-trained multiturn conversation performance.
 
I'm also thinking about this in terms of a2a: If we drop 1 in favor of 2,  now my A2A server has to make some design decisions about how to format the history before running the agent. I don't think it will be a problem performance wise but then it does bring back the problem of doing formatting in a way that isn't transparent to users. We could add some sort of prompt_template formatting to our ServingConfig to alleviate that problem.

> What I am sure matters is the content of the history, and that is my reason to push back on this PR: we are deciding what information goes into and how it is formatted, in a way that is not transparent to the users.

On this point it may be my fault for my draft implementation (unless I misunderstand you). I have some code to extract the span info and turn it into the prompt (`span_to_string`) before passing it to the model. I 100% agree this is bad and we should not merge support for my draft `span_to_string` method and history implementation for langchain, google, and llama_index that are using that design.

njbrake:
Ok, I've reduced the scope of code changes and condensed our discussion into a descriptive and prescriptive docs page. Let me know what you think.

docs/agents/index.md
```diff
@@ -124,3 +124,91 @@ async def main():
 if __name__ == "__main__":
     asyncio.run(main())
 ```
+
+### Multi-Turn Conversations
```
daavoo, line 128: 💅 can be in a follow-up: This feels worth a separate page. Not sure where. Maybe we should add a new "How-to" section where we cover specific scenarios

docs/agents/index.md
```diff
@@ -124,3 +124,91 @@ async def main():
...
+Please ask me some questions to understand my preferences, then provide recommendations.
+"""
+
+agent_trace = agent.run(prompt)
+print(f"Final recommendations: {agent_trace.final_output}")
+```
+
+This approach is demonstrated in our [MCP Agent cookbook example](../cookbook/mcp_agent.ipynb), where an agent uses user interaction tools to gather trip planning information dynamically. The agent can ask clarifying questions, get user preferences, and provide personalized recommendations all within a single `run()` call.
+
+!!! tip "When to Use Each Approach"
```
daavoo: 💅 this should probably be at the top before describing each option in depth

src/any_agent/tracing/agent_trace.py
```diff
@@ -82,6 +83,69 @@ def compute_cost_info(
     return CostInfo.model_validate(new_info)
 
 
+def spans_to_messages(spans: list[AgentSpan]) -> list[dict[str, str]]:
```
daavoo: Maybe worth adding a pydantic `Message` type?

src/any_agent/tracing/agent_trace.py
```diff
@@ -82,6 +83,69 @@ def compute_cost_info(
...
+        List of message dicts with 'role' and 'content' keys.
+
+    """
+    messages: list[dict[str, str]] = []
+
+    # Process spans in chronological order (excluding the final invoke_agent span)
+    # Filter out any non-AgentSpan objects and agent invocation spans
+    filtered_spans: list[AgentSpan] = []
+    for span in spans:
+        if hasattr(span, "is_agent_invocation") and not span.is_agent_invocation():
```
daavoo: How can there be an span that is not an `AgentSpan`?

src/any_agent/tracing/agent_trace.py
daavoo: I think I would also like this as a method of `AgentTrace`

src/any_agent/tracing/agent_trace.py
njbrake: Just in case someone was calling the method with a badly constructed list. But, if we move this to be a method of AgentTrace, the corner case goes away 👍

docs/agents/index.md
njbrake: Totally. I think we should have a cookbook example for multiturn that expands and implements these recommendations, at which point we can re-org this section into a new page.

PR #469, opened 2025-06-13
MCP serving

### LINKED ISSUE #273: Add option to serve agent via `MCP`

javiermtorres:
In a similar way to serving via A2A, the MCP SDK is used to provide an MCP server that can serve an existing agent. It is currently handled only in async mode.

Closes #273

### DIFF
--- a/src/any_agent/config.py
+++ b/src/any_agent/config.py
@@ -56,8 +56,6 @@ class MCPStdio(BaseModel):
     If none is specified, the default behavior is that the agent will have access to all tools under that MCP server.
     """
 
-    cwd: str | None = None
-
     client_session_timeout_seconds: float | None = 5
     """the read timeout passed to the MCP ClientSession."""
 
--- a/src/any_agent/config.py
+++ b/src/any_agent/config.py
@@ -56,6 +56,8 @@ class MCPStdio(BaseModel):
     If none is specified, the default behavior is that the agent will have access to all tools under that MCP server.
     """
 
+    cwd: str | None = None
+
     client_session_timeout_seconds: float | None = 5
     """the read timeout passed to the MCP ClientSession."""
 
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -27,7 +27,7 @@ if TYPE_CHECKING:
     from opentelemetry.trace import Tracer
     from pydantic import BaseModel
 
-    from any_agent.serving.config import A2AServingConfig, MCPServingConfig
+    from any_agent.serving import A2AServingConfig, MCPServingConfig
     from any_agent.tools.mcp.mcp_server import _MCPServerBase
 
 
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -272,14 +272,12 @@ class AnyAgent(ABC):
             ```
 
         """
-        from any_agent.serving import A2AServingConfig, MCPServingConfig
+        from any_agent.serving import MCPServingConfig
 
         if isinstance(serving_config, MCPServingConfig):
             self._serve_mcp(serving_config)
-        elif isinstance(serving_config, A2AServingConfig):
-            self._serve_a2a(serving_config)
         else:
-            self._serve_a2a(None)
+            self._serve_a2a(serving_config)
 
     async def _serve_a2a_async(
         self, serving_config: A2AServingConfig | None
@@ -352,13 +350,11 @@ class AnyAgent(ABC):
             ```
 
         """
-        from any_agent.serving import A2AServingConfig, MCPServingConfig
+        from any_agent.serving import MCPServingConfig
 
         if isinstance(serving_config, MCPServingConfig):
             return await self._serve_mcp_async(serving_config)
-        if isinstance(serving_config, A2AServingConfig):
-            return await self._serve_a2a_async(serving_config)
-        return await self._serve_a2a_async(None)
+        return await self._serve_a2a_async(serving_config)
 
     def _recreate_with_config(self, new_config: AgentConfig) -> AnyAgent:
         """Create a new agent instance with the given config, preserving MCP servers and tools.
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -262,7 +262,7 @@ class AnyAgent(ABC):
                           Must be an instance of A2AServingConfig or MCPServingConfig.
 
         Raises:
-            ImportError: If the `a2a` dependencies are not installed.
+            ImportError: If the `a2a` dependencies are not installed and an `A2AServingConfig` is used.
 
         Example:
             ```
@@ -293,12 +293,6 @@ class AnyAgent(ABC):
         if serving_config is None:
             serving_config = A2AServingConfig()
 
-        if not isinstance(serving_config, A2AServingConfig):
-            msg = (
-                f"serving_config must be an instance of A2AServingConfig, "
-                f"got {serving_config.type}. "
-            )
-            raise ValueError(msg)
         app = await _get_a2a_app_async(self, serving_config=serving_config)
 
         return await serve_a2a_async(
@@ -312,14 +306,7 @@ class AnyAgent(ABC):
     async def _serve_mcp_async(
         self, serving_config: MCPServingConfig
     ) -> tuple[asyncio.Task[Any], uvicorn.Server]:
-        from any_agent.serving import MCPServingConfig, serve_mcp_async
-
-        if not isinstance(serving_config, MCPServingConfig):
-            msg = (
-                f"serving_config must be an instance of MCPServingConfig, "
-                f"got {serving_config.type}. "
-            )
-            raise ValueError(msg)
+        from any_agent.serving import serve_mcp_async
 
         return await serve_mcp_async(
             self,
@@ -349,7 +336,7 @@ class AnyAgent(ABC):
                           Must be an instance of A2AServingConfig or MCPServingConfig.
 
         Raises:
-            ImportError: If the `a2a` dependencies are not installed.
+            ImportError: If the `a2a` dependencies are not installed and an `A2AServingConfig` is used.
 
         Example:
             ```
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import asyncio
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Any, assert_never
+from typing import TYPE_CHECKING, Any, assert_never, overload
+from uuid import uuid4
 
 from opentelemetry import trace as otel_trace
 
@@ -12,6 +13,8 @@ from any_agent.config import (
     Tool,
 )
 from any_agent.logging import logger
+from any_agent.serving import serve_a2a, serve_a2a_async, serve_mcp, serve_mcp_async
+from any_agent.serving.config import A2AServingConfig, MCPServingConfig
 from any_agent.tools.wrappers import _wrap_tools
 from any_agent.tracing.agent_trace import AgentTrace
 from any_agent.tracing.exporter import SCOPE_NAME
@@ -27,7 +30,6 @@ if TYPE_CHECKING:
     from opentelemetry.trace import Tracer
     from pydantic import BaseModel
 
-    from any_agent.serving.config import A2AServingConfig
     from any_agent.tools.mcp.mcp_server import _MCPServerBase
 
 
@@ -219,7 +221,7 @@ class AnyAgent(ABC):
 
     # Subclass: ServedAnyAgent?
 
-    def serve(self, serving_config: A2AServingConfig | None = None) -> None:
+    def serve_a2a(self, serving_config: A2AServingConfig) -> None:
         """Serve this agent using the protocol defined in the serving_config.
 
         Args:
@@ -237,23 +239,33 @@ class AnyAgent(ABC):
             ```
 
         """
-        from any_agent.serving import A2AServingConfig, _get_a2a_app, serve_a2a
+        from any_agent.serving import serve_a2a
 
-        if serving_config is None:
-            serving_config = A2AServingConfig()
+        serve_a2a(self, serving_config)
 
-        app = _get_a2a_app(self, serving_config=serving_config)
+    def serve_mcp(self, serving_config: MCPServingConfig) -> None:
+        from any_agent.serving import serve_mcp
 
-        serve_a2a(
-            app,
-            host=serving_config.host,
-            port=serving_config.port,
-            endpoint=serving_config.endpoint,
-            log_level=serving_config.log_level,
-        )
+        serve_mcp(self, serving_config)
 
-    async def serve_async(
-        self, serving_config: A2AServingConfig | None = None
+    @overload
+    async def serve(self, serving_config: MCPServingConfig) -> None: ...
+
+    @overload
+    async def serve(self, serving_config: A2AServingConfig | None = None) -> None: ...
+
+    async def serve(
+        self, serving_config: MCPServingConfig | A2AServingConfig | None
+    ) -> None:
+        if isinstance(serving_config, MCPServingConfig):
+            serve_mcp(self, serving_config)
+        elif isinstance(serving_config, A2AServingConfig):
+            serve_a2a(self, serving_config)
+        else:
+            serve_a2a(self, A2AServingConfig())
+
+    async def serve_a2a_async(
+        self, serving_config: A2AServingConfig
     ) -> tuple[asyncio.Task[Any], uvicorn.Server]:
         """Serve this agent asynchronously using the protocol defined in the serving_config.
 
@@ -283,11 +295,7 @@ class AnyAgent(ABC):
             ```
 
         """
-        from any_agent.serving import (
-            A2AServingConfig,
-            _get_a2a_app_async,
-            serve_a2a_async,
-        )
+        from any_agent.serving import A2AServingConfig, serve_a2a_async
 
         if serving_config is None:
             serving_config = A2AServingConfig()
@@ -299,14 +307,10 @@ class AnyAgent(ABC):
                 f"Currently only A2A serving is supported."
             )
             raise ValueError(msg)
-        app = await _get_a2a_app_async(self, serving_config=serving_config)
 
         return await serve_a2a_async(
-            app,
-            host=serving_config.host,
-            port=serving_config.port,
-            endpoint=serving_config.endpoint,
-            log_level=serving_config.log_level,
+            self,
+            serving_config=serving_config,
         )
 
     def _recreate_with_config(self, new_config: AgentConfig) -> AnyAgent:
@@ -367,17 +371,30 @@ class AnyAgent(ABC):
     # by default: {query: str}
 
     async def serve_mcp_async(
-        self, serving_config: A2AServingConfig | None = None
+        self, serving_config: MCPServingConfig
     ) -> tuple[asyncio.Task[Any], uvicorn.Server]:
         from any_agent.serving import serve_mcp_async
 
-        return await serve_mcp_async(
-            self,
-            host=serving_config.host,
-            port=serving_config.port,
-            endpoint=serving_config.endpoint,
-            log_level=serving_config.log_level,
-        )
+        return await serve_mcp_async(self, serving_config=serving_config)
+
+    @overload
+    async def serve_async(
+        self, serving_config: MCPServingConfig
+    ) -> tuple[asyncio.Task[Any], uvicorn.Server]: ...
+
+    @overload
+    async def serve_async(
+        self, serving_config: A2AServingConfig | None = None
+    ) -> tuple[asyncio.Task[Any], uvicorn.Server]: ...
+
+    async def serve_async(
+        self, serving_config: MCPServingConfig | A2AServingConfig | None
+    ) -> tuple[asyncio.Task[Any], uvicorn.Server]:
+        if isinstance(serving_config, MCPServingConfig):
+            return await serve_mcp_async(self, serving_config)
+        if isinstance(serving_config, A2AServingConfig):
+            return await serve_a2a_async(self, serving_config)
+        return await serve_a2a_async(self, A2AServingConfig())
 
     @abstractmethod
     async def _load_agent(self) -> None:
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -217,6 +217,8 @@ class AnyAgent(ABC):
         trace.final_output = final_output
         return trace
 
+    # Subclass: ServedAnyAgent?
+
     def serve(self, serving_config: A2AServingConfig | None = None) -> None:
         """Serve this agent using the protocol defined in the serving_config.
 
@@ -360,6 +362,22 @@ class AnyAgent(ABC):
         await new_agent._load_agent()
 
         return new_agent
+    # FIXME: get a pydantic model, build a query
+    # def query(arguments: type[BaseModel]) -> str:
+    # by default: {query: str}
+
+    async def serve_mcp_async(
+        self, serving_config: A2AServingConfig | None = None
+    ) -> tuple[asyncio.Task[Any], uvicorn.Server]:
+        from any_agent.serving import serve_mcp_async
+
+        return await serve_mcp_async(
+            self,
+            host=serving_config.host,
+            port=serving_config.port,
+            endpoint=serving_config.endpoint,
+            log_level=serving_config.log_level,
+        )
 
     @abstractmethod
     async def _load_agent(self) -> None:
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import asyncio
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any, assert_never, overload
-from uuid import uuid4
 
 from opentelemetry import trace as otel_trace
 
@@ -13,8 +12,6 @@ from any_agent.config import (
     Tool,
 )
 from any_agent.logging import logger
-from any_agent.serving import serve_a2a, serve_a2a_async, serve_mcp, serve_mcp_async
-from any_agent.serving.config import A2AServingConfig, MCPServingConfig
 from any_agent.tools.wrappers import _wrap_tools
 from any_agent.tracing.agent_trace import AgentTrace
 from any_agent.tracing.exporter import SCOPE_NAME
@@ -30,6 +27,7 @@ if TYPE_CHECKING:
     from opentelemetry.trace import Tracer
     from pydantic import BaseModel
 
+    from any_agent.serving.config import A2AServingConfig, MCPServingConfig
     from any_agent.tools.mcp.mcp_server import _MCPServerBase
 
 
@@ -219,17 +217,52 @@ class AnyAgent(ABC):
         trace.final_output = final_output
         return trace
 
-    # Subclass: ServedAnyAgent?
+    def _serve_a2a(self, serving_config: A2AServingConfig | None) -> None:
+        from any_agent.serving import A2AServingConfig, _get_a2a_app, serve_a2a
 
-    def serve_a2a(self, serving_config: A2AServingConfig) -> None:
+        if serving_config is None:
+            serving_config = A2AServingConfig()
+
+        app = _get_a2a_app(self, serving_config=serving_config)
+
+        serve_a2a(
+            app,
+            host=serving_config.host,
+            port=serving_config.port,
+            endpoint=serving_config.endpoint,
+            log_level=serving_config.log_level,
+        )
+
+    def _serve_mcp(self, serving_config: MCPServingConfig) -> None:
+        from any_agent.serving import (
+            serve_mcp,
+        )
+
+        serve_mcp(
+            self,
+            host=serving_config.host,
+            port=serving_config.port,
+            endpoint=serving_config.endpoint,
+            log_level=serving_config.log_level,
+        )
+
+    @overload
+    def serve(self, serving_config: MCPServingConfig) -> None: ...
+
+    @overload
+    def serve(self, serving_config: A2AServingConfig | None = None) -> None: ...
+
+    def serve(
+        self, serving_config: MCPServingConfig | A2AServingConfig | None = None
+    ) -> None:
         """Serve this agent using the protocol defined in the serving_config.
 
         Args:
             serving_config: Configuration for serving the agent. If None, uses default A2AServingConfig.
-                          Must be an instance of A2AServingConfig.
+                          Must be an instance of A2AServingConfig or MCPServingConfig.
 
         Raises:
-            ImportError: If the `serving` dependencies are not installed.
+            ImportError: If the `a2a` dependencies are not installed.
 
         Example:
             ```
@@ -239,52 +272,89 @@ class AnyAgent(ABC):
             ```
 
         """
-        from any_agent.serving import serve_a2a
+        from any_agent.serving import A2AServingConfig, MCPServingConfig
 
-        serve_a2a(self, serving_config)
-
-    def serve_mcp(self, serving_config: MCPServingConfig) -> None:
-        from any_agent.serving import serve_mcp
-
-        serve_mcp(self, serving_config)
-
-    @overload
-    async def serve(self, serving_config: MCPServingConfig) -> None: ...
-
-    @overload
-    async def serve(self, serving_config: A2AServingConfig | None = None) -> None: ...
-
-    async def serve(
-        self, serving_config: MCPServingConfig | A2AServingConfig | None
-    ) -> None:
         if isinstance(serving_config, MCPServingConfig):
-            serve_mcp(self, serving_config)
+            self._serve_mcp(serving_config)
         elif isinstance(serving_config, A2AServingConfig):
-            serve_a2a(self, serving_config)
+            self._serve_a2a(serving_config)
         else:
-            serve_a2a(self, A2AServingConfig())
+            self._serve_a2a(None)
 
-    async def serve_a2a_async(
-        self, serving_config: A2AServingConfig
+    async def _serve_a2a_async(
+        self, serving_config: A2AServingConfig | None
+    ) -> tuple[asyncio.Task[Any], uvicorn.Server]:
+        from any_agent.serving import (
+            A2AServingConfig,
+            _get_a2a_app_async,
+            serve_a2a_async,
+        )
+
+        if serving_config is None:
+            serving_config = A2AServingConfig()
+
+        if not isinstance(serving_config, A2AServingConfig):
+            msg = (
+                f"serving_config must be an instance of A2AServingConfig, "
+                f"got {serving_config.type}. "
+            )
+            raise ValueError(msg)
+        app = await _get_a2a_app_async(self, serving_config=serving_config)
+
+        return await serve_a2a_async(
+            app,
+            host=serving_config.host,
+            port=serving_config.port,
+            endpoint=serving_config.endpoint,
+            log_level=serving_config.log_level,
+        )
+
+    async def _serve_mcp_async(
+        self, serving_config: MCPServingConfig
+    ) -> tuple[asyncio.Task[Any], uvicorn.Server]:
+        from any_agent.serving import MCPServingConfig, serve_mcp_async
+
+        if not isinstance(serving_config, MCPServingConfig):
+            msg = (
+                f"serving_config must be an instance of MCPServingConfig, "
+                f"got {serving_config.type}. "
+            )
+            raise ValueError(msg)
+
+        return await serve_mcp_async(
+            self,
+            host=serving_config.host,
+            port=serving_config.port,
+            endpoint=serving_config.endpoint,
+            log_level=serving_config.log_level,
+        )
+
+    @overload
+    async def serve_async(
+        self, serving_config: MCPServingConfig
+    ) -> tuple[asyncio.Task[Any], uvicorn.Server]: ...
+
+    @overload
+    async def serve_async(
+        self, serving_config: A2AServingConfig | None = None
+    ) -> tuple[asyncio.Task[Any], uvicorn.Server]: ...
+
+    async def serve_async(
+        self, serving_config: MCPServingConfig | A2AServingConfig | None = None
     ) -> tuple[asyncio.Task[Any], uvicorn.Server]:
         """Serve this agent asynchronously using the protocol defined in the serving_config.
 
         Args:
             serving_config: Configuration for serving the agent. If None, uses default A2AServingConfig.
-                          Must be an instance of A2AServingConfig.
-
-        Returns:
-            A tuple containing:
-            - asyncio.Task: The server task (keep a reference to prevent garbage collection)
-            - uvicorn.Server: The server instance for controlling the server lifecycle
+                          Must be an instance of A2AServingConfig or MCPServingConfig.
 
         Raises:
-            ImportError: If the `serving` dependencies are not installed.
+            ImportError: If the `a2a` dependencies are not installed.
 
         Example:
             ```
             agent = await AnyAgent.create_async("tinyagent", AgentConfig(...))
-            config = A2AServingConfig(port=8080)
+            config = MCPServingConfig(port=8080)
             task, server = await agent.serve_async(config)
             try:
                 # Server is running
@@ -295,23 +365,13 @@ class AnyAgent(ABC):
             ```
 
         """
-        from any_agent.serving import A2AServingConfig, serve_a2a_async
+        from any_agent.serving import A2AServingConfig, MCPServingConfig
 
-        if serving_config is None:
-            serving_config = A2AServingConfig()
-
-        if not isinstance(serving_config, A2AServingConfig):
-            msg = (
-                f"serving_config must be an instance of A2AServingConfig, "
-                f"got {serving_config.type}. "
-                f"Currently only A2A serving is supported."
-            )
-            raise ValueError(msg)
-
-        return await serve_a2a_async(
-            self,
-            serving_config=serving_config,
-        )
+        if isinstance(serving_config, MCPServingConfig):
+            return await self._serve_mcp_async(serving_config)
+        if isinstance(serving_config, A2AServingConfig):
+            return await self._serve_a2a_async(serving_config)
+        return await self._serve_a2a_async(None)
 
     def _recreate_with_config(self, new_config: AgentConfig) -> AnyAgent:
         """Create a new agent instance with the given config, preserving MCP servers and tools.
@@ -366,35 +426,6 @@ class AnyAgent(ABC):
         await new_agent._load_agent()
 
         return new_agent
-    # FIXME: get a pydantic model, build a query
-    # def query(arguments: type[BaseModel]) -> str:
-    # by default: {query: str}
-
-    async def serve_mcp_async(
-        self, serving_config: MCPServingConfig
-    ) -> tuple[asyncio.Task[Any], uvicorn.Server]:
-        from any_agent.serving import serve_mcp_async
-
-        return await serve_mcp_async(self, serving_config=serving_config)
-
-    @overload
-    async def serve_async(
-        self, serving_config: MCPServingConfig
-    ) -> tuple[asyncio.Task[Any], uvicorn.Server]: ...
-
-    @overload
-    async def serve_async(
-        self, serving_config: A2AServingConfig | None = None
-    ) -> tuple[asyncio.Task[Any], uvicorn.Server]: ...
-
-    async def serve_async(
-        self, serving_config: MCPServingConfig | A2AServingConfig | None
-    ) -> tuple[asyncio.Task[Any], uvicorn.Server]:
-        if isinstance(serving_config, MCPServingConfig):
-            return await serve_mcp_async(self, serving_config)
-        if isinstance(serving_config, A2AServingConfig):
-            return await serve_a2a_async(self, serving_config)
-        return await serve_a2a_async(self, A2AServingConfig())
 
     @abstractmethod
     async def _load_agent(self) -> None:
--- a/src/any_agent/serving/__init__.py
+++ b/src/any_agent/serving/__init__.py
@@ -1,6 +1,13 @@
 try:
     from .config import A2AServingConfig, MCPServingConfig
-    from .server import serve_a2a, serve_a2a_async, serve_mcp, serve_mcp_async
+    from .server import (
+        _get_a2a_app,
+        _get_a2a_app_async,
+        serve_a2a,
+        serve_a2a_async,
+        serve_mcp,
+        serve_mcp_async,
+    )
 except ImportError as e:
     msg = "You need to `pip install 'any-agent[a2a]'` to use this method."
     raise ImportError(msg) from e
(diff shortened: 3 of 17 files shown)


src/any_agent/frameworks/any_agent.py
```diff
@@ -252,6 +254,23 @@ async def serve_async(
             log_level=serving_config.log_level,
         )
 
+    # FIXME: get a pydantic model, build a query
+    # def query(arguments: type[BaseModel]) -> str:
+    # by default: {query: str}
+
+    async def serve_mcp_async(
```
daavoo: Let's group this under a single `serve` and decide what to serve based on whether `A2AServingConfig` or `MCPServingConfig` type was passed

src/any_agent/serving/__init__.py
```diff
@@ -5,5 +5,16 @@
     msg = "You need to `pip install 'any-agent[a2a]'` to use this method."
     raise ImportError(msg) from e
 
+try:
+    from .server import serve_mcp_async
+except ImportError as e:
+    msg = "You need to `pip install 'any-agent[mcp]'` to use this method."
```
daavoo: afaik, `mcp` is part of the mandatory dependencies so this is not needed

src/any_agent/serving/server.py
```diff
@@ -49,6 +55,87 @@ def _create_server(
...
+                },
+            )
+        ]
+
+    @server.call_tool()
+    async def handle_call_tool(
+        name: str, arguments: dict | None
+    ) -> list[mcptypes.TextContent | mcptypes.ImageContent | mcptypes.EmbeddedResource]:
+        result = await agent.run_async(arguments["query"])
+        return [mcptypes.TextContent(type="text", text=result.final_output)]
```
daavoo: final_output might be an structured output so it won't be `text`. You need to research how people handle pydnatic objects as return type

src/any_agent/serving/server.py
```diff
@@ -49,6 +55,87 @@ def _create_server(
...
+
+    @server.list_resource_templates()
+    async def handle_list_resource_templates() -> list[mcptypes.ResourceTemplate]:
```
daavoo: Do you need to define this even if you return an empty list?

src/any_agent/serving/server.py
```diff
@@ -49,6 +55,87 @@ def _create_server(
...
+    server = _create_mcp_server_instance(agent)
+    init_options = server.create_initialization_options()
+
+    async def _handle_sse(request: Request):
+        async with sse.connect_sse(
+            request.scope, request.receive, request._send
+        ) as streams:
+            await server.run(streams[0], streams[1], init_options)
+        # Return empty response to avoid NoneType error
+        return Response()
```
daavoo: Why is this needed? Is there some docs/example somewhere?

src/any_agent/serving/server.py
daavoo, line 1: Let's divide the code into `a2a` `mcp` dir/files

tests/integration/helpers.py
```diff
@@ -44,3 +46,25 @@ def wait_for_a2a_server(server_url: str):
         if attempts >= max_attempts:
             msg = f"Could not connect to {server_url}. Tried {max_attempts} times with {poll_interval} second interval."
             raise ConnectionError(msg)
+
+async def wait_for_a2a_server_async(server_url: str):
```
daavoo:
```suggestion
async def wait_for_server_async(server_url: str):
```

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
daavoo, line 1: Let's rename the file or split into `a2a` / `mcp` files

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -304,3 +315,124 @@ def test_load_and_run_multi_agent_a2a_sync(
                 # Force kill if graceful shutdown failed
                 server_process.kill()
                 server_process.join()
+
+
+@pytest.mark.skipif(
+    os.environ.get("ANY_AGENT_INTEGRATION_TESTS", "FALSE").upper() != "TRUE",
+    reason="Integration tests require `ANY_AGENT_INTEGRATION_TESTS=TRUE` env var",
+)
```
daavoo: Per #466 
```suggestion
```

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -304,3 +315,124 @@ def test_load_and_run_multi_agent_a2a_sync(
...
+@pytest.mark.asyncio
+async def test_load_and_run_multi_agent_mcp(
+    agent_framework: AgentFramework, test_port: int
+) -> None:
+    """Tests that an agent contacts another using A2A using the adapter tool.
```
daavoo: Update docstring

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -304,3 +315,124 @@ def test_load_and_run_multi_agent_a2a_sync(
...
+    """
+    if agent_framework in [
+        # async a2a is not supported
+        AgentFramework.SMOLAGENTS,
+        # spans are not built correctly
+        AgentFramework.LLAMA_INDEX,
+        # AgentFramework.GOOGLE,
+    ]:
+        pytest.skip(
+            "https://github.com/mozilla-ai/any-agent/issues/357 tracks fixing so these tests can be re-enabled"
```
daavoo: The issue is closed. Are these skips still needed?

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -304,3 +315,124 @@ def test_load_and_run_multi_agent_a2a_sync(
...
+        # We cannot use the SSE stream, as it will not be closed in the request
+        await wait_for_a2a_server_async(ping_url)
+
+        # Search agent is ready for card resolution
+
+        main_agent_cfg = AgentConfig(
+            instructions="Use the available tools to obtain additional information to answer the query.",
+            description="The orchestrator that can use other agents via tools using the A2A protocol.",
+            tools=[
+                MCPSse(url=server_url, client_session_timeout_seconds=300),
```
daavoo: Should we also support Stdio?

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -304,3 +315,124 @@ def test_load_and_run_multi_agent_a2a_sync(
...
+
+        try:
+            build_tree(agent_trace.spans).model_dump_json(indent=2)
+        except KeyError as e:
+            pytest.fail(f"The span tree was not built successfully: {e}")
+
+    finally:
+        if main_agent:
+            await main_agent._mcp_servers[0].server.cleanup()
+            main_agent.exit()
```
daavoo: There is no `exit`

src/any_agent/frameworks/any_agent.py
javiermtorres: Ok, I didn't want to do it yet until I checked how much machinery we have in common, but I agree.

src/any_agent/serving/__init__.py
javiermtorres: Yes, I noticed this later on but I didn't remember about this piece :(

src/any_agent/serving/server.py
javiermtorres: Ok, np. I can fix that later. But MCP may not allow structured content, in this sense; let me check it out.

src/any_agent/serving/server.py
javiermtorres: Possibly not.... There were some other calls I did remove, I probably forgot about this one.

src/any_agent/serving/server.py
javiermtorres: Yes. I'll point out to the relevant pieces in the MCP reference implementation, which are basically code sample comments :-/

tests/integration/helpers.py
javiermtorres: Ah true :D

src/any_agent/serving/server.py
javiermtorres: If you're ok with the direction of the PR, I'll go further and restructure it in the next push, so we can discuss what things can be done in common.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: Ok, I'll rebase.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: I know that the A2A cases did have issues for smolagents and llama_index, I'll re-check them just in case.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: Hum.... Since the issue talks about _serving_, I was assuming SSE. But, in any case, I don't see stdio as useful since you could already make a fun around an any-agent run_async.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: Should we do specific cleanup? SSE tools will need cleanup, at least. And exit _did_ exist, right????

njbrake:
@javiermtorres  looks like there are a few things that David has comments out about: if you could, you can re-add me as a reviewer when it's a little closer to ready, I'll wait for the dust to settle from those comments before I take a deep dive

src/any_agent/serving/server.py
javiermtorres: In the spec, not yet in the SDK: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/371/
I'll prepare for this, but basically I'll output a serialization and be ready to generate the output schema.

tests/integration/helpers.py
javiermtorres: I'll also rename `wait_for_a2a_server` to `wait_for_server` since there's nothing a2a specific here.

src/any_agent/serving/server.py
javiermtorres: There's a reference in https://github.com/modelcontextprotocol/python-sdk/blob/1eb1bba83c70c3121bce7fc0263e5fac2c3f0520/src/mcp/server/sse.py#L33. I'll include that as a comment.

daavoo: Is there a problem wit using `fastmcp`? It seems that we are using the "lowlevel" API but I am not sure there is a strong reason against the more high-level fastmcp?

javiermtorres: Actually it's included in the `mcp.server` package: https://github.com/modelcontextprotocol/python-sdk/blob/1eb1bba83c70c3121bce7fc0263e5fac2c3f0520/src/mcp/server/__init__.py
So even if it's in the `lowlevel` package, at least it's exposed (and documented) at a top level package: https://github.com/modelcontextprotocol/python-sdk/blob/1eb1bba83c70c3121bce7fc0263e5fac2c3f0520/src/mcp/server/lowlevel/server.py
It actually works fine, until one wants to control the web server lifecycle.

You can easily see that the core of the server functionality is the same for the fastmcp: https://github.com/modelcontextprotocol/python-sdk/blob/1eb1bba83c70c3121bce7fc0263e5fac2c3f0520/src/mcp/server/fastmcp/server.py#L646 only with more machinery for additional auth endpoints, converting tools, etc, which we don't really need for tests. So I'm not sure that it will behave in a different way than the current implementation. It is also bundled with the uvicorn setup, which we separate in our tests.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: Apparently we're storing some context managers along the way @daavoo but not closing them at the end. This:

```
await main_agent._mcp_servers[0].mcp_connection._exit_stack.aclose()
```

frees up the MCP SSE connections in most cases; I'd say we should add this as some sort of cleanup procedure.

javiermtorres:
> @javiermtorres looks like there are a few things that David has comments out about: if you could, you can re-add me as a reviewer when it's a little closer to ready, I'll wait for the dust to settle from those comments before I take a deep dive

I think it would be good to get your opinion now, if you have the time to spare. TL;DR: should we care about agent/server cleanup (exemplified mostly in the integration tests); fastmcp (more complicated solution, harder to manage) vs standalone server (simpler component) in serving; should we move asap to streamable http which will likely provide less headaches in cleanup.

njbrake:
> > @javiermtorres looks like there are a few things that David has comments out about: if you could, you can re-add me as a reviewer when it's a little closer to ready, I'll wait for the dust to settle from those comments before I take a deep dive
> 
> I think it would be good to get your opinion now, if you have the time to spare. TL;DR: should we care about agent/server cleanup (exemplified mostly in the integration tests); fastmcp (more complicated solution, harder to manage) vs standalone server (simpler component) in serving; should we move asap to streamable http which will likely provide less headaches in cleanup.

@javiermtorres sounds good! Before I take a look, could you update the branch to resolve the conflicts from main? I would like to be sure that I'm looking at changes that are up to date with the design in main.

javiermtorres:
Sure! I'll rebase to main and ping you again.

javiermtorres:
@njbrake done, this should do the trick:
```
pytest -s tests/integration/test_load_and_run_multi_agent_a2a_tool.py::test_load_and_run_multi_agent_mcp
```

src/any_agent/config.py
```diff
@@ -56,6 +56,8 @@ class MCPStdio(BaseModel):
     If none is specified, the default behavior is that the agent will have access to all tools under that MCP server.
     """
 
+    cwd: str | None = None
```
njbrake: Please add a description for this parameter, similar to how it's done with the others.

src/any_agent/frameworks/any_agent.py
```diff
@@ -217,7 +219,9 @@ async def run_async(
         trace.final_output = final_output
         return trace
 
-    def serve(self, serving_config: A2AServingConfig | None = None) -> None:
+    # Subclass: ServedAnyAgent?
```
njbrake: Please remove or add context to this comment.

src/any_agent/frameworks/any_agent.py
```diff
@@ -217,7 +219,9 @@ async def run_async(
...
+
+    def serve_a2a(self, serving_config: A2AServingConfig) -> None:
```
njbrake: Please add context to why the default was removed. Docstring is now also out of date.

src/any_agent/frameworks/any_agent.py
```diff
@@ -235,23 +239,33 @@ def serve(self, serving_config: A2AServingConfig | None = None) -> None:
             ```
 
         """
-        from any_agent.serving import A2AServingConfig, _get_a2a_app, serve_a2a
+        from any_agent.serving import serve_a2a
 
-        if serving_config is None:
-            serving_config = A2AServingConfig()
+        serve_a2a(self, serving_config)
```
njbrake: Would it be possible to remove the changes you made around a2a and consolidate this PR to only focus on serving over MCP? I'm happy to accept a re-org of the a2a changes but this PR is about MCP, and I'd rather take care of a2a related changes somewhere else.

src/any_agent/frameworks/any_agent.py
```diff
@@ -235,23 +239,33 @@ def serve(self, serving_config: A2AServingConfig | None = None) -> None:
...
 
-        app = _get_a2a_app(self, serving_config=serving_config)
+    def serve_mcp(self, serving_config: MCPServingConfig) -> None:
+        from any_agent.serving import serve_mcp
```
njbrake: Please fix missing docstring.

src/any_agent/frameworks/any_agent.py
```diff
@@ -235,23 +239,33 @@ def serve(self, serving_config: A2AServingConfig | None = None) -> None:
...
         """
-        from any_agent.serving import A2AServingConfig, _get_a2a_app, serve_a2a
+        from any_agent.serving import serve_a2a
 
-        if serving_config is None:
-            serving_config = A2AServingConfig()
+        serve_a2a(self, serving_config)
 
-        app = _get_a2a_app(self, serving_config=serving_config)
+    def serve_mcp(self, serving_config: MCPServingConfig) -> None:
```
njbrake: Should there be a default value if no serving_config is provided?

src/any_agent/frameworks/any_agent.py
```diff
@@ -235,23 +239,33 @@ def serve(self, serving_config: A2AServingConfig | None = None) -> None:
...
-        )
+        serve_mcp(self, serving_config)
 
-    async def serve_async(
-        self, serving_config: A2AServingConfig | None = None
+    @overload
+    async def serve(self, serving_config: MCPServingConfig) -> None: ...
+
+    @overload
+    async def serve(self, serving_config: A2AServingConfig | None = None) -> None: ...
```
njbrake: I'm not super familiar with the use of '@overload`: can you explain what these are doing and why they are needed? Thanks!

src/any_agent/frameworks/any_agent.py
```diff
@@ -235,23 +239,33 @@ def serve(self, serving_config: A2AServingConfig | None = None) -> None:
...
+
+    async def serve(
+        self, serving_config: MCPServingConfig | A2AServingConfig | None
+    ) -> None:
```
njbrake: Missing docstrings. Hmm, maybe serve_mcp and serve_a2a should be converted to private member functions if they're designed to be used underneath this wrapper?

src/any_agent/frameworks/any_agent.py
```diff
@@ -235,23 +239,33 @@ def serve(self, serving_config: A2AServingConfig | None = None) -> None:
...
+        if isinstance(serving_config, MCPServingConfig):
+            serve_mcp(self, serving_config)
+        elif isinstance(serving_config, A2AServingConfig):
+            serve_a2a(self, serving_config)
+        else:
+            serve_a2a(self, A2AServingConfig())
```
njbrake: Maybe we should add a deprecation warning here to say that a person should provide a serving config when calling serve, and that we will remove this default in the future? WDYT?

src/any_agent/frameworks/any_agent.py
```diff
@@ -235,23 +239,33 @@ def serve(self, serving_config: A2AServingConfig | None = None) -> None:
...
 
-    async def serve_async(
-        self, serving_config: A2AServingConfig | None = None
+    @overload
+    async def serve(self, serving_config: MCPServingConfig) -> None: ...
+
+    @overload
+    async def serve(self, serving_config: A2AServingConfig | None = None) -> None: ...
+
+    async def serve(
```
njbrake: wasn't serve a sync function, and serve_async the async version of it?

src/any_agent/frameworks/any_agent.py
```diff
@@ -360,6 +366,35 @@ async def _recreate_with_config_async(self, new_config: AgentConfig) -> AnyAgent
         await new_agent._load_agent()
 
         return new_agent
+    # FIXME: get a pydantic model, build a query
+    # def query(arguments: type[BaseModel]) -> str:
+    # by default: {query: str}
```
njbrake: Please fix or make a ticket to track this fix and remove this comment. Thanks!

src/any_agent/frameworks/any_agent.py
```diff
@@ -360,6 +366,35 @@ async def _recreate_with_config_async(self, new_config: AgentConfig) -> AnyAgent
...
+    ) -> tuple[asyncio.Task[Any], uvicorn.Server]: ...
+
+    async def serve_async(
+        self, serving_config: MCPServingConfig | A2AServingConfig | None
+    ) -> tuple[asyncio.Task[Any], uvicorn.Server]:
+        if isinstance(serving_config, MCPServingConfig):
+            return await serve_mcp_async(self, serving_config)
+        if isinstance(serving_config, A2AServingConfig):
+            return await serve_a2a_async(self, serving_config)
+        return await serve_a2a_async(self, A2AServingConfig())
```
njbrake: same as the sync version question, I'm wondering if we should add some sort of deprecation warning.

src/any_agent/serving/agent_card.py
```diff
@@ -12,7 +12,7 @@
     from any_agent.serving.config import A2AServingConfig
 
 
-def _get_agent_card(agent: AnyAgent, serving_config: A2AServingConfig) -> AgentCard:
+def _build_agent_card(agent: AnyAgent, serving_config: A2AServingConfig) -> AgentCard:
```
njbrake: Why the rename?

src/any_agent/serving/server.py
```diff
@@ -3,111 +3,167 @@
...
 
+async def serve_a2a_async(
+    agent: AnyAgent,
+    serving_config: A2AServingConfig,
+) -> tuple[asyncio.Task[Any], uvicorn.Server]:
+    """Provide an A2A server to be used in an event loop."""
+    agent_card = _build_agent_card(agent, serving_config)
     request_handler = DefaultRequestHandler(
-        agent_executor=AnyAgentExecutor(agent, task_manager),
+        agent_executor=AnyAgentExecutor(agent),
```
njbrake: Bad merge, please remove the edits you made to a2a code wherever possible so that the changes focus on MCP.

tests/integration/helpers.py
```diff
@@ -5,6 +5,27 @@
 import requests
 
 
+def mock_search_web(query: str) -> str:
```
njbrake: Sorry if I missed this 😅  this function looks familar. Did it come from somewhere else that can have it removed from that location?

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -2,15 +2,34 @@
...
+    level=logging.DEBUG,
+    format=FORMAT,
+    datefmt="[%X]",
+    handlers=[RichHandler(rich_tracebacks=True)],
+)
+
+logger = logging.getLogger("any_agent_test")
+logger.setLevel(logging.ERROR)
+mcp_logger = logging.getLogger("mcp.server.sse")
+mcp_logger.setLevel(logging.ERROR)
```
njbrake: Please remove logging from test code

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -279,3 +307,112 @@ def test_load_and_run_multi_agent_a2a_sync(agent_framework: AgentFramework) -> N
...
+        # AgentFramework.TINYAGENT,
+        # AgentFramework.GOOGLE,
+        # async a2a is not supported
+        AgentFramework.SMOLAGENTS,
+        # spans are not built correctly
+        AgentFramework.LLAMA_INDEX,
+    ]:
+        pytest.skip(
+            "https://github.com/mozilla-ai/any-agent/issues/357 tracks fixing so these tests can be re-enabled"
+        )
```
njbrake: are these skip reasons correct?

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -279,3 +307,112 @@ def test_load_and_run_multi_agent_a2a_sync(agent_framework: AgentFramework) -> N
...
+        agent_trace = await main_agent.run_async(DATE_PROMPT)
+
+        _assert_valid_agent_trace(agent_trace)
+        _assert_contains_current_date_info(agent_trace.final_output)
+        _assert_has_tool_date_agent_call(agent_trace)
+
+    finally:
+        if main_agent:
+            await main_agent._mcp_servers[0].mcp_connection._exit_stack.aclose()
+        # FIXME incorporate into the cleanup
```
njbrake: Please fix or create an issue to fix and remove this comment.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -279,3 +307,112 @@ def test_load_and_run_multi_agent_a2a_sync(agent_framework: AgentFramework) -> N
...
+        if AppStatus.should_exit_event is not None:
+            AppStatus.should_exit_event.set()
+            AppStatus.should_exit_event = None
+        if served_server:
+            served_server.should_exit = True
+            await served_task
+        await asyncio.sleep(0.1)
```
njbrake: why the sleep?

njbrake (changes requested):
Thanks for your work to get MCP serving up and running for any-agent ✨ ! I'm excited to see this coming into focus and to have another option for serving an agent. Please address the comments I've provided, there is a bit more cleanup required to reduce the scope of changes in the PR as well as clean up the docs.

src/any_agent/config.py
javiermtorres: I'll remove this parameter. It is a leftover from some tests that didn't work out. Sorry to have left it in the code at this stage.

src/any_agent/frameworks/any_agent.py
javiermtorres: The latest commit includes changes in line with @daavoo 's comment on unifying the A2A and MCP servings: https://github.com/mozilla-ai/any-agent/pull/469#discussion_r2144599584
Unless I got it wrong, I understood that the serving should be unified as much as possible (which turns out to be quite feasible since both just provide a Starlette app plugged in to a uvicorn).

javiermtorres: This arch aspect (single serving infra vs separate methods) was one of the items I wanted to agree on.

tests/integration/helpers.py
javiermtorres: I'll check for rebase issues. Some code may have been added by the rebase :-/

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: I'll remove it. I thought there was a task being starved in the event loop, but it was a red herring.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: This is actually the second arch aspect I wanted to discuss: whether the agent should care about the cleanup. Right now it isn't, and it'll stay as part of the test cleanup for the time being. But I'll open an issue to track this if we think it's an important aspect.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: For SmolAgents it is, afaict. For LlamaIndex, we need to add the check in https://github.com/mozilla-ai/any-agent/issues/472 back to find out.

src/any_agent/frameworks/any_agent.py
javiermtorres: To keep backwards compatibility, a None condition is assigned to A2A serving. But maybe it could now be a necessary parameter instead of being optional.

src/any_agent/frameworks/any_agent.py
javiermtorres: Basically, it's a way of mimicking overloaded functions and adding appropriate type checking. The implementation will fall back to the first found non-overloaded entry (IIRC) but the type checker will try to find the most restrictive definition and apply it. But in this case it's not overwhelmingly useful since there is not much typechecking to do once the single input param is checked :-/ The None case is the really only ambiguous case, and it could be just removed by forcing the user to choose an instance (e.g. a default).

I added it as part of the "unifying serving" approach, but if the A2A and MCP servings are going to be separate, of course it's not necessary.

src/any_agent/frameworks/any_agent.py
javiermtorres: Depends on how we structure the serving, but yes, if there's a single serve frontend, then it makes sense to hide the _serve_a2a and _serve_mcp.

src/any_agent/frameworks/any_agent.py
javiermtorres: regardless of the direction in the serving structure, I'll make sure that the async public methods are marked _async 👍

src/any_agent/frameworks/any_agent.py
javiermtorres: This is done to be backwards compatible with the existing serve. Depending on the direction we want to take, yes, we will mark some uses as deprecated as needed.

src/any_agent/frameworks/any_agent.py
javiermtorres: I'll probably make a new ticket 👍

src/any_agent/serving/agent_card.py
javiermtorres: The issue I see with the name is that it can be confused with GETting the agent card from the server. This method actually makes a new card available to A2A clients.

src/any_agent/serving/server.py
javiermtorres: There is some code rehashing to make infra more modular (same handling of uvicorn instances, same handling of Starlette aps, etc). But the rebase already left some outdated code, so I'll check that's not the cause.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
njbrake: Sorry, this is my fault I should have been more specific 😆 . This skip reason is incorrect, the issue it references is already closed.

src/any_agent/frameworks/any_agent.py
njbrake: I am all for unifying the serve logic to toggle between a2a vs mcp based on which configuration is passed in 👍 . However, what I would like to have removed in this PR is the changes like renaming get_agent_card to build_agent _card, changing and removing calls to _get_a2a_app, etc. I agree that the changes are good, but I would like that to be addressed in a separate PR because those changes go beyond the unifying the top level serve logic.

javiermtorres: Ok. I'll then make a single API but postpone the serving restructuring. I would nonetheless defend the get->build change since it's an internal call and the change is not related to functionality.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: Ah ok! I'll try to add some more meaningful text for the skip, sorry.

njbrake (commented):
There's some weird bug in github that keeps saying it's waiting for my review even though I have reviewed this 😆

javiermtorres:
I'll try removing you from the reviewers for the time being

javiermtorres:
Integration running ok: https://github.com/mozilla-ai/any-agent/actions/runs/15777858607/job/44476257625

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: All skipped frameworks, and their reason, should be there now. I'll try to check them periodically and/or open issues accordingly.

src/any_agent/frameworks/any_agent.py
```diff
@@ -217,15 +217,52 @@ async def run_async(
...
         """Serve this agent using the protocol defined in the serving_config.
 
         Args:
             serving_config: Configuration for serving the agent. If None, uses default A2AServingConfig.
-                          Must be an instance of A2AServingConfig.
+                          Must be an instance of A2AServingConfig or MCPServingConfig.
 
         Raises:
-            ImportError: If the `serving` dependencies are not installed.
+            ImportError: If the `a2a` dependencies are not installed.
```
njbrake: This command should still work if a2a dependencies are not installed. You shouldn't need to have a2a dependencies installed to serve via mcp

src/any_agent/frameworks/any_agent.py
```diff
@@ -235,42 +272,89 @@ def serve(self, serving_config: A2AServingConfig | None = None) -> None:
...
+        self, serving_config: MCPServingConfig
+    ) -> tuple[asyncio.Task[Any], uvicorn.Server]:
+        from any_agent.serving import MCPServingConfig, serve_mcp_async
+
+        if not isinstance(serving_config, MCPServingConfig):
+            msg = (
+                f"serving_config must be an instance of MCPServingConfig, "
+                f"got {serving_config.type}. "
+            )
+            raise ValueError(msg)
```
njbrake: Imo this value check should be removed. The typing already specifies mcpserverving config

src/any_agent/frameworks/any_agent.py
```diff
@@ -235,42 +272,89 @@ def serve(self, serving_config: A2AServingConfig | None = None) -> None:
...
         if serving_config is None:
             serving_config = A2AServingConfig()
 
-        app = _get_a2a_app(self, serving_config=serving_config)
+        if not isinstance(serving_config, A2AServingConfig):
+            msg = (
+                f"serving_config must be an instance of A2AServingConfig, "
+                f"got {serving_config.type}. "
+            )
+            raise ValueError(msg)
```
njbrake: imo this should be removed as typing already specifies taht it will be that type

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -52,15 +52,14 @@ async def test_load_and_run_multi_agent_a2a(agent_framework: AgentFramework) ->
...
-        AgentFramework.SMOLAGENTS,
-        # spans are not built correctly
-        AgentFramework.LLAMA_INDEX,
-        # AgentFramework.GOOGLE,
-    ]:
+    skip_reason = {
+        AgentFramework.SMOLAGENTS: "async a2a is not supported",
+        AgentFramework.GOOGLE: "Default value is not supported in function declaration schema for Google AI",
+        # AgentFramework.LLAMA_INDEX: "spans are not built correctly",
+    }
```
njbrake: Why did you start skipping google and stop skipping llama index?

src/any_agent/frameworks/any_agent.py
javiermtorres: Ok, I was actually modifying the outdated `serving` dependency, which was still lingering in some places.

src/any_agent/frameworks/any_agent.py
javiermtorres: Ok, I was following the existing code for the A2A case.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: Since we don't check that the span tree is built correctly, we don't have a way of checking that condition (it may be that we need to skip it later on, but the case is currently passing).

Re: Google, I'll try to add some more context. IIRC it was due to how the MCP is handled from the framework, but I'll try to add a comment with some more information.

src/any_agent/frameworks/any_agent.py
javiermtorres: Ah, wait, this is the common code that gets called when MCP _or_ A2A is called, so the importerror may happen at this point. I can remove the overloading if you'd prefer, but then we'd have separate serve_a2a and serve_mcp, which I assumed we didn't want to have.

src/any_agent/frameworks/any_agent.py
javiermtorres: Ref: https://github.com/mozilla-ai/any-agent/blob/21178b5ee58168d593c35d11217c85876b788e17/src/any_agent/frameworks/any_agent.py#L293

src/any_agent/frameworks/any_agent.py
javiermtorres: It was already present in the code. I can remove it if you insist.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: It seems that setting the temp to 0 avoids issues for the Google framework and doesn't affect the other frameworks.

src/any_agent/serving/config.py
```diff
@@ -89,3 +89,38 @@ class A2AServingConfig(BaseModel):
...
+            endpoint="/my-agent",
+            skills=[
+                AgentSkill(
+                    id="search",
+                    name="web_search",
+                    description="Search the web for information"
+                )
+            ]
+        )
+
```
daavoo: Need to update this

src/any_agent/frameworks/any_agent.py
njbrake: I think there should be a way to accomplish this without removing the overloading. Let me know if you're stuck on this and I can give a try at looking into it as well.

src/any_agent/frameworks/any_agent.py
njbrake: Yes lets remove it now that this is called by the serve method which is already checking the types.

src/any_agent/frameworks/any_agent.py
javiermtorres: AFAICT the overloading definitions should not include docstrings: https://docs.astral.sh/ruff/rules/overload-with-docstring/
I'm changing the `Raises` definition slightly, at least:

```suggestion
            ImportError: If the `a2a` dependencies are not installed and an `A2AServingConfig` is used.
```

src/any_agent/frameworks/any_agent.py
javiermtorres: Removing as in the following case

src/any_agent/serving/config.py
javiermtorres: Updated

src/any_agent/frameworks/any_agent.py
```diff
@@ -217,15 +217,52 @@ async def run_async(
...
         """Serve this agent using the protocol defined in the serving_config.
 
         Args:
             serving_config: Configuration for serving the agent. If None, uses default A2AServingConfig.
-                          Must be an instance of A2AServingConfig.
+                          Must be an instance of A2AServingConfig or MCPServingConfig.
 
         Raises:
-            ImportError: If the `serving` dependencies are not installed.
+            ImportError: If the `a2a` dependencies are not installed and an `A2AServingConfig` is used.
```
njbrake, line 265: Have you tried and verified this to be the case? I would think that if the a2a dependencies are not installed, it would throw an exception on the line `from any_agent.serving import A2AServingConfig`.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
```diff
@@ -203,15 +203,13 @@ def test_load_and_run_multi_agent_a2a_sync(agent_framework: AgentFramework) -> N
...
-    if agent_framework in [
-        # async a2a is not supported
-        AgentFramework.SMOLAGENTS,
-        # spans are not built correctly
-        AgentFramework.LLAMA_INDEX,
-        # AgentFramework.GOOGLE,
-    ]:
+    skip_reason = {
+        AgentFramework.SMOLAGENTS: "async a2a is not supported; run_async_in_sync fails",
+        # AgentFramework.LLAMA_INDEX: "spans are not built correctly",
```
njbrake, line 208: If this is commented out then can it be removed?

tests/integration/test_load_and_run_multi_agent_mcp_tool.py
```diff
@@ -0,0 +1,149 @@
...
+async def test_load_and_run_multi_agent_mcp(
+    agent_framework: AgentFramework, test_port: int
+) -> None:
+    """Tests that an agent contacts another as a tool using MCP.
+
+    Note that there is an issue when using Google ADK: https://github.com/google/adk-python/pull/566
+    """
+    skip_reason = {
+        AgentFramework.SMOLAGENTS: "async a2a is not supported",
+        # AgentFramework.LLAMA_INDEX: "spans are not built correctly",
```
njbrake: If this is commented out can it be removed?

njbrake (changes requested):
The tests contain some commented out code that maybe should be removed?

There are almost 100 comments on this PR, yet when I review the PR I don't see any of the comments. Why do all the comments dissappear, is it related to the merge strategy? It's difficult to continually re-review a PR and not to be able to see my previous comments to understand whether they've been resolved or not.

src/any_agent/frameworks/any_agent.py
javiermtorres: IIRC the imports are needed because of the instance checks, which are needed to route the overloaded calls. I can use an else directly instead of the elif, which should avoid the need for the import until _serve_a2a. Let me check it out.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: As I mentioned, I'd rather leave it until we set the trace tree check back in place and we check if this actually fails.

javiermtorres:
> There are almost 100 comments on this PR, yet when I review the PR I don't see any of the comments. Why do all the comments dissappear, is it related to the merge strategy? It's difficult to continually re-review a PR and not to be able to see my previous comments to understand whether they've been resolved or not.

Some of the comments refer to code to be removed; I suppose that removes a few of the comments. Take into account that I was asked to remove the unified uvicorn handling during this PR; this may account for some as well. I don't really know why GitHub would show or not the comments, beyond my resolving issues when I change them in the code. Also note that there is not a 1:1 correspondence between comments and changes. I am open to suggestions as to how to improve the reviewing process.

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
njbrake: Why is this change related to MCP? If it's not related to your MCP changes, the place to change whether a test should be skipped or not is not correct for this PR.

njbrake: If the comment is out-of-date etc, I would rather it be left out of date and no changes made to this file, and address the change in a future PR. Trying to keep this PR as focused as possible on only MCP changes.

src/any_agent/frameworks/any_agent.py
```diff
@@ -235,42 +272,76 @@ def serve(self, serving_config: A2AServingConfig | None = None) -> None:
...
         """
-        from any_agent.serving import A2AServingConfig, _get_a2a_app, serve_a2a
+        from any_agent.serving import A2AServingConfig, MCPServingConfig
+
+        if isinstance(serving_config, MCPServingConfig):
+            self._serve_mcp(serving_config)
+        elif isinstance(serving_config, A2AServingConfig):
+            self._serve_a2a(serving_config)
+        else:
+            self._serve_a2a(None)
```
njbrake:
```suggestion
        from any_agent.serving import MCPServingConfig

        if isinstance(serving_config, MCPServingConfig):
            self._serve_mcp(serving_config)
            return
         
        from any_agent.serving import A2AServingConfig
        if isinstance(serving_config, A2AServingConfig):
            self._serve_a2a(serving_config)
        else:
            self._serve_a2a(None)
```

Something like this. This change should also be applied to the async version

javiermtorres: I was actually doing now

```suggestion
        if isinstance(serving_config, MCPServingConfig):
            self._serve_mcp(serving_config)
        else:
            self._serve_a2a(serving_config)

```

which actually delays the import until _serve_a2a (it's already there). Would that be ok to you?

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
javiermtorres: I'm reflecting in the MCP tests the current state of the A2A tests: https://github.com/mozilla-ai/any-agent/blob/f1cdca2355993319a0ecfa6cddca856e43d9799a/tests/integration/test_load_and_run_multi_agent_a2a_tool.py#L58
There is already an issue for it: https://github.com/mozilla-ai/any-agent/issues/472
I'll remove it from this PR since you insist on it.

src/any_agent/frameworks/any_agent.py
njbrake: yep, good with me 👍

src/any_agent/serving/__init__.py
```diff
@@ -1,15 +1,68 @@
...
+        "A2AServingConfig",
+        "MCPServingConfig",
+        "_get_a2a_app",
+        "_get_a2a_app_async",
+        "serve_a2a",
+        "serve_a2a_async",
+        "serve_mcp",
+        "serve_mcp_async",
+    ]
+except ImportError:
```
javiermtorres, line 29: @njbrake This works fine when the a2a dep is not installed. However, it will cause the mkdocs to use the failing definitions and not the real ones. I don't know why, since removing the try will locate the right packages, so an import error at this stage does not seem likely. Would you know why mkdocs would be behaving in this way?

tests/unit/serving/test_agent_card.py
```diff
@@ -11,6 +11,7 @@
 # Skip entire module if a2a dependencies are not available
 pytest.importorskip("a2a.types")
 pytest.importorskip("any_agent.serving.agent_card")
+pytest.importorskip("typing.override")
```
javiermtorres, line 14: @njbrake note that override is not available on 3.11. It was causing importerrors in the previous commit. It was possibly masked by importing `any_agent.serving`, which now needs to be available since the MCP option is always available. It's quite probably related to the mkdocs issue; I could separate the imports into `serving.a2a` and `serving.mcp`, but I guess this beats the whole "same API" point?

src/any_agent/serving/__init__.py
javiermtorres: Removing the "failing definitions" would mean that the user would not see these definitions, causing a usual ImportError (possibly ModuleNotFound) instead of the customized error we want them to see.

javiermtorres:
@njbrake I have the impression the linter needs the A2A dependencies to check the typing. See this: https://github.com/mozilla-ai/any-agent/actions/runs/15855972847/job/44701273127?pr=469#step:4:1
The "Any" return is quite possibly due to the ignore type annotations in the import fail definitions. Wdyt?

src/any_agent/frameworks/any_agent.py
```diff
@@ -272,14 +272,12 @@ def serve(
...
         if isinstance(serving_config, MCPServingConfig):
             self._serve_mcp(serving_config)
-        elif isinstance(serving_config, A2AServingConfig):
-            self._serve_a2a(serving_config)
         else:
-            self._serve_a2a(None)
+            self._serve_a2a(serving_config)
 
     async def _serve_a2a_async(
         self, serving_config: A2AServingConfig | None
```
njbrake, line 283:
```suggestion
        self, serving_config: "A2AServingConfig" | None
```

Import A2AServingConfig at the top of the file in a TYPE_CHECKING block

njbrake: Then, in this function, import A2AServingConfig

src/any_agent/serving/__init__.py
njbrake: You can delete all of this code to load the failing definitions. Please see my comment and code suggestion in AnyAgent which I believe resolves the issue

src/any_agent/serving/__init__.py
```diff
@@ -1,15 +1,34 @@
...
+
+    __all__ = [
+        "A2AServingConfig",
+        "MCPServingConfig",
+        "_get_a2a_app",
+        "_get_a2a_app_async",
+        "serve_a2a",
+        "serve_a2a_async",
+        "serve_mcp",
+        "serve_mcp_async",
```
njbrake: Why not append to __all__ instead of resetting it?

tests/integration/test_load_and_run_multi_agent_a2a_tool.py
njbrake: Yes, please remove it from this PR

src/any_agent/serving/__init__.py
javiermtorres: Appending was my first choice, actually :-/ It's a bit less error prone, but I was afraid that the tooling messed up somehow when building docs or similar.

javiermtorres: ...just saying :-/ https://github.com/mozilla-ai/any-agent/actions/runs/15873473329/job/44755411580?pr=469#step:5:116

PR #473, opened 2025-06-13
refactor(tracing): Rename cache invalidation methods for clarity

njbrake:
There was a mismatch in the cache name when it was changed from 'usage' to 'tokens', resulting in the cache never being cleared

Updated the cache invalidation methods in `AgentTrace` to reflect the change from usage and cost to tokens and cost. Adjusted related docstrings and method calls to maintain consistency. This enhances code readability and aligns with the new terminology.


daavoo (commented):
We should add a test for the cache invaliadtion.

PR #474, opened 2025-06-13
refactor(agent): Simplify  run_async and fix linting errors

njbrake:
Some misc fixes:

1. trace was redefining the `trace` imported by OTEL (shadow imports)
2. I would only like to try catch issues with the agent run, I don't want to try catch issues with OTEL, I want that exception passed directly back to the caller EDIT: I had to undo this part because it was making things extra tricky to properly close the last span. I decided that I wasn't so determined not to catch issues outside of the agent run that it was worth making the code extra complicated.
3. linting was unhappy with trace_id not being defined so I added that explicit definition


src/any_agent/frameworks/any_agent.py
```diff
@@ -165,43 +165,52 @@ async def run_async(
...
+        trace = AgentTrace()
+        trace_id: int
+        instrumentation_enabled = instrument and self._instrumentor
+
+        with self._tracer.start_as_current_span(
+            f"invoke_agent [{self.config.name}]"
+        ) as invoke_span:
+            if instrumentation_enabled:
+                trace_id = invoke_span.get_span_context().trace_id
+                async with self._lock:
```
javiermtorres: Do we need a lock here? Is there any async operation within the instrumenting code?

src/any_agent/frameworks/any_agent.py
```diff
@@ -165,43 +165,52 @@ async def run_async(
...
-        except Exception as e:
-            if instrument:
+                        # Get the instrumented trace if available, otherwise use the original trace
+                        instrumented_trace = self._running_traces.pop(trace_id)
+                        if instrumented_trace is not None:
+                            trace = instrumented_trace
+                trace.add_span(invoke_span)
+                raise AgentRunError(trace) from e
+
+            if instrumentation_enabled:
```
javiermtorres, line 200: Shouldn't instrumentation remain enabled when the agent is being served, as opposed to being run in a standalone way?

daavoo (approved):
LGTM.

Why are the linting errors showing for you but not caught in the CI / default config?
Can we update the config to be all on the same page?

src/any_agent/frameworks/any_agent.py
daavoo: Parallel `run_async` calls would enter race conditions because we modify the same object (`_running_traces`).

Per the comment:

```
                    # We check the locked `_running_traces` inside `instrument`.
                    # If there is more than 1 entry in `running_traces`, it means that the agent has
                    # already being instrumented so we won't instrument it again.
```

src/any_agent/frameworks/any_agent.py
daavoo: > Shouldn't instrumentation remain enabled when the agent is being served, as opposed to being run in a standalone way?

What would be the difference?

src/any_agent/frameworks/any_agent.py
javiermtorres: Async calls are not really parallel; they're "yielded" afaict to the same OS thread running the current event loop when a file desc would be waited on. Usually a lock is not a big burden, but if we _know_ that there won't be async tasks around, I was wondering if we could avoid it. But this depends on the conditions under which we want to run this; if we are serving several agents at the same time then it's definitely needed.

njbrake:
> LGTM.
> 
> Why are the linting errors showing for you but not caught in the CI / default config? Can we update the config to be all on the same page?

I run pylint on the code in my IDE, but in our pre-commit we only run ruff and mypy. Ruff and pylint have some overlap but I guess not completely.

src/any_agent/frameworks/any_agent.py
javiermtorres: Because this involves a non-trivial amount of effort being done for each call :-/ And adding this overhead to a running server with a huge amount of calls could be noticeable.

daavoo:
> I run pylint on the code in my IDE, but in our pre-commit we only run ruff and mypy. Ruff and pylint have some overlap but I guess not completely.

It would be good to check the specific rules that you see in pylint and see if we are ignoring them in ruff (and proceed to enable those)

njbrake:
> > I run pylint on the code in my IDE, but in our pre-commit we only run ruff and mypy. Ruff and pylint have some overlap but I guess not completely.
> 
> It would be good to check the specific rules that you see in pylint and see if we are ignoring them in ruff (and proceed to enable those)

I will take a look. I checked for the exact code and it isn't being ignored by ruff explicitly, but there are a whole bunch of ruff codes that I don't understand that are being ignored as specified in the pyproject.toml. I'll take a look as a side project while I'm working on other stuff.

src/any_agent/frameworks/any_agent.py
daavoo: It is needed for cases like the following, which we explicitly check in our integration tests:

https://github.com/mozilla-ai/any-agent/blob/bf37f60760ffacdd386aa9d61bc8b169ba2cc065/tests/integration/test_load_and_run_agent_twice.py#L51-L54

javiermtorres: Yes, that's the kind of cases I was looking for.

PR #476, opened 2025-06-13
fix(instrumentation/google): Properly uninstrument when original call…

daavoo:
…backs were `None`.


daavoo:
Integration Tests: https://github.com/mozilla-ai/any-agent/actions/runs/15634158739

PR #480, opened 2025-06-13
feat(agent): Add async config recreation methods and enhance agent pr…

njbrake:
This fixes the notebook broken test: the problem was with the prepare_for_serving_a2a: when recreating the agent, we were throwing away the existing MCP servers, which then caused an issue when trying to turn off and restart the MCP servers.

When making the changes I also realized we were missing some async versions of a few functions, so I added those as well.

- Introduced `recreate_with_config` and `recreate_with_config_async` methods to allow agents to be recreated with new configurations while preserving MCP servers and tools.
- Updated the A2A serving logic to use the new async methods for better performance and resource management.
- Enhanced the `prepare_agent_for_a2a` functions to utilize the new recreation methods, ensuring smoother transitions and avoiding connection timeouts.


njbrake:
Cookbooks tests: https://github.com/mozilla-ai/any-agent/actions/runs/15638694737

Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/15638691356/job/44060448065

PR #481, opened 2025-06-13
tests: Use `output_type` in `test_load_and_run_agent.py`.

daavoo:
@njbrake no strong opinion but since we already test 900 different things in `test_load_and_run_agent` I would prefer to reduce the number of integration tests, either by merging like this case or relying on unit tests if it is possible (I feel like structured outputs we actually want to run a real thing)


daavoo:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/15638996069

PR #463, opened 2025-06-12
feat(tools): Wrap tool calls to catch exceptions

daavoo:
Frameworks handle tool exceptions too inconsistently.
We use a wrapper to consistently return a string formatted whenever a exception occurs.

This allow us to also set an appropriate OTEL Status for the tool span but also allows the Agent to try to recover from the tool failure (hoping that the exception provides some clear guidance on what went wrong).

PR #482, opened 2025-06-13
Unit test for LLM tool exceptions

javiermtorres:
The LLM calls are patched (fortunately most frameworks use LiteLLM) so the test can be run as a unit test instead of an integration test. Since the calls can be streamed or "final", the patching needs to be done per framework (and possibly adapted with framework bumps).


tests/conftest.py
```diff
@@ -14,6 +14,16 @@
 from any_agent.tracing.agent_trace import AgentTrace
 from tests.integration.helpers import wait_for_server_async
 
+PATCH_PER_FRAMEWORK = {
```
njbrake: Should this constant be moved to a conftest.py inside of the unit/tools section?

Also, it may be useful to give this a more descriptive constant name, PATCH_PER_FRAMEWORK isn't very descriptive.

tests/unit/tools/test_exception_run.py
```diff
@@ -0,0 +1,146 @@
...
+        # `('Recursion limit of 25 reached without hitting a stop condition. You can increase the ...oubleshooting/errors/GRAPH_RECURSION_LIMIT',)`
+        AgentFramework.LANGCHAIN,
+    ):
+        pytest.skip(f"Framework {agent_framework.value} not currently passing the test")
+
+    kwargs = {}
+
+    tmp_file = "tmp.txt"
+    # FIXME patch some call so an exception is triggered
+    # An exception within a tool will be handled by the framework
```
njbrake: Can this comment be removed and either the fixme addressed, or a Github issue created to track this, if it's an issue that should be solved

njbrake (commented):
Thanks for the work on this! Can the integration test be removed now that this is a unit test?

tests/conftest.py
daavoo: I don't think we should be patching these internal functions.

See the current integration tests for simulation of an unexpected RuntimeError at some point inside the framework execution:

https://github.com/mozilla-ai/any-agent/blob/32759ecef8033ede286f2ef0b25e3a56da644f83/tests/integration/test_error_handling.py#L35-L39

And an actual "tool execution" error:

https://github.com/mozilla-ai/any-agent/blob/32759ecef8033ede286f2ef0b25e3a56da644f83/tests/integration/test_error_handling.py#L63-L77

This patching lies somewhere in the middle where it is a failure that feels hard to understand happening in the "real world".

tests/unit/tools/test_exception_run.py
javiermtorres: I'll remove the comment.

tests/conftest.py
javiermtorres: Ok

javiermtorres:
The original idea, before #463 was merged, was to check that an exception actually caused an `AgentRunError` to be thrown. As you mention in the PR, the handling in the different tools is inconsistent, to the point where just throwing an error in the tool or patching just the completions was not enough to cause a real exception to be thrown, hence the specific patching points. Sometimes an error in the tool would just be caught by the underlying framework and turned into a trace in itself, so it would just get captured.

I'll patch the LLM calls then in the integration test added in #463, instead of the previous incarnation, to move it to unit test realm.

Now that I'm inspecting #463 more closely, using a plain string match like in https://github.com/mozilla-ai/any-agent/blob/eb27b5d447716edb99e969410c0f13bbcbc8c870/src/any_agent/tracing/instrumentation/common.py#L24 can be prone to errors. Now that we have structured output in place, it feels much better to wrap the tool output in a real json/dict envelope and act on it. I'll open an issue.

daavoo:
> Now that we have structured output in place, it feels much better to wrap the tool output in a real json/dict envelope and act on it.

Note that the reason for using a plain string is that frameworks can / would mess up with whatever object you return. It was the only way I found to get a consistent behavior across frameworks.

tests/unit/tools/test_exception_run.py
```diff
@@ -0,0 +1,117 @@
...
+    agent_framework: AgentFramework,
+) -> None:
+    """An exception raised inside a tool will be caught by us.
+
+    We make sure an appropriate Status is set to the tool execution span.
+    We allow the Agent to try to recover from the tool calling failure.
+    """
+
+    skip_reason = {
+        AgentFramework.AGNO: "Does not decide that the loop should end, and calls the tool the maximum amount of times before giving up",
```
javiermtorres: Agno does not seem to recognize that the tool didn't end properly, so it retries for a long time.

tests/unit/tools/test_exception_run.py
```diff
@@ -0,0 +1,117 @@
...
+        AgentFramework.GOOGLE: "When mocking the LLM tool call, the tool calling does not happen and the loop is ended",
```
javiermtorres: When using the Google fw, somehow the loop silenty ends without indication of a tool failure (I'll try to debug this thoroughly)

tests/conftest.py
javiermtorres: The idea for this would be exactly to cause issues where each framework could not catch them, and see how each framework dealt with it. But now that the approach towards errors has changed, it does not make sense.

njbrake:
@javiermtorres is there an integration test that can/should be removed if this unit test is merged?

njbrake (changes requested):
Before this is merged I would like it to work with all frameworks.

tests/unit/tools/test_exception_run.py
daavoo: The integration test is not skipped and works for AGNO: https://github.com/mozilla-ai/any-agent/blob/22d2b750fde03768c658c5c67fe676fd8682a5a9/tests/integration/test_error_handling.py#L78 would be good to understand what is different here

tests/unit/tools/test_exception_run.py
daavoo: The integration test is not skipped and works for GOOGLE: https://github.com/mozilla-ai/any-agent/blob/22d2b750fde03768c658c5c67fe676fd8682a5a9/tests/integration/test_error_handling.py#L78 would be good to understand what is different here

tests/unit/tools/test_exception_run.py
javiermtorres: The real LLM will give up and just give an answer after a couple of rounds. The mocks have been updated to reflect that.

tests/unit/tools/test_exception_run.py
javiermtorres: Apparently it's the same case for the Google framework.

javiermtorres:
> @javiermtorres is there an integration test that can/should be removed if this unit test is merged?

The corresponding integration test would be https://github.com/mozilla-ai/any-agent/blob/94974ff70d91a3cd06f955cf881d73d25e87df52/tests/integration/test_error_handling.py#L78
AFAICT they'd be equivalent. The frameworks are a bit quirky but the details can be dug out.

> Before this is merged I would like it to work with all frameworks.

The test runs now in all frameworks.

njbrake:
> > @javiermtorres is there an integration test that can/should be removed if this unit test is merged?
> 
> The corresponding integration test would be
> 
> https://github.com/mozilla-ai/any-agent/blob/94974ff70d91a3cd06f955cf881d73d25e87df52/tests/integration/test_error_handling.py#L78
> 
> AFAICT they'd be equivalent. The frameworks are a bit quirky but the details can be dug out.

If this is the case then can you please remove that integration test as a part of this PR?

javiermtorres:
> If this is the case then can you please remove that integration test as a part of this PR?

@daavoo would you agree to this? I'd say we could start moving the more "boring" tests dealing with deterministic behaviour into unit territory, while keeping integration for more complete e2e, plus a sample of the unit tests that could signal changes in the LLM or framework behaviours.

daavoo:
> @daavoo would you agree to this? I'd say we could start moving the more "boring" tests dealing with deterministic behaviour into unit territory, while keeping integration for more complete e2e, plus a sample of the unit tests that could signal changes in the LLM or framework behaviours.

I am ok with that.

I would keep the integration test for runtime errors.

TBH, I am not sure this has been a good tradeoff of effort / added value. The integration test was failing (afaik) and here we ended up with a bunch of complex mocks.

javiermtorres:
> I would keep the integration test for runtime errors.

I definitely agree on this, specially in the face of api changes.

> TBH, I am not sure this has been a good tradeoff of effort / added value. The integration test was failing (afaik) and here we ended up with a bunch of complex mocks.

At least for me, this PR has shed some light on how the different frameworks handle the LLM responses, not only at the protocol level, but also when building the context and deciding when to stop the dialogue. Also, it can be useful when testing the instrumenters, since we don't need to depend on the LLM real behaviour (even with temp=0, the version may have been modified). This is also a reason to keep the integration test, for me: even though we have accurate mocks of at least one LLM for tests, without latency or cost issues, we will need to align with the real model at times.

Ok, I'm merging this as it is; please feel free to open an issue for further modifications.

PR #487, opened 2025-06-16
feat(agent): Enhance final output handling with retry logic and JSON validation

njbrake:
Two things here: one fix, and one "infinite loop til max steps" fix.

The problem: Google wasn't getting the hint that the problem wasn't with the info provided, but the fact that it was providing a direct string and not a json string. 
Screenshot 2025-06-16 at 11 23 02 AM

The fix: Use a two stage validation approach. Step one, make sure they passed a valid json string, if not, return an error saying that it's invalid JSON. Step two, once it's valid JSON, then we can try to parse it with pydantic.
Now:

Screenshot 2025-06-16 at 11 24 38 AM

### DIFF
--- a/src/any_agent/frameworks/google.py
+++ b/src/any_agent/frameworks/google.py
@@ -99,6 +99,9 @@ class GoogleAgent(AnyAgent):
 
         if self.config.output_type:
             final_output = None
+            final_output_attempts = 0
+            max_attempts = 3
+
             async for event in runner.run_async(
                 user_id=user_id,
                 session_id=session_id,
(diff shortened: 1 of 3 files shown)


njbrake:
It's an intermittent failure so I can't say with absolute certainty that it's gone, however, the integration is passing on the github action, and I've re-run the test locally a bunch of times and it no longer fails (it used to fail about 1 in every 5 attempts)

https://github.com/mozilla-ai/any-agent/actions/runs/15686124747

daavoo (approved):
❤️

PR #489, opened 2025-06-17
Include exception info on docs

### LINKED ISSUE #471: Add info on exception and traces to docs
We should add this expected usage to the user docs

_Originally posted by @daavoo in https://github.com/mozilla-ai/any-agent/pull/432#discussion_r2137283575_

javiermtorres:
This PR includes some info on the exceptions thrown by the run method, plus a pointer to the AgentRunError class.

Closes #471


docs/agents/index.md
```diff
@@ -95,12 +95,17 @@ agent = AnyAgent.create(
 ## Running Agents
 
 ```python
-agent_trace = agent.run("Which Agent Framework is the best??")
-print(agent_trace.final_output)
+try:
+    agent_trace = agent.run("Which Agent Framework is the best??")
+    print(agent_trace.final_output)
+except AgentRunError as are:
+    spans = are.trace.spans
```
daavoo: No strong opinion, just `e` is more common convention and `agent_trace` to match the previous block💅 
```suggestion
except AgentRunError as e:
    agent_trace= e.trace
```

src/any_agent/frameworks/any_agent.py
```diff
@@ -228,9 +228,11 @@ def serve(self, serving_config: A2AServingConfig | None = None) -> None:
             ImportError: If the `serving` dependencies are not installed.
 
         Example:
+            ```
```
daavoo, line 231: 👍

docs/agents/index.md
javiermtorres: I tend to use the acronym since the single 'e' is easily abused. I'll change it anyway, np.

PR #491, opened 2025-06-17
Make any_agent.__version__ attribute available

HareeshBahuleyan:
Currently it fails when I run `print(any_agent.__version__)`

I tested if it works from this branch by doing the following:

```
# Create a temp env
mamba create -n test-env python=3.11
mamba activate test-env

# Install from this branch
pip install "git+https://github.com/mozilla-ai/any-agent.git@package-version#egg=any-agent[all]"

# Test version displayed correctly
python -c "import any_agent; print(any_agent.__version__)"
```

### DIFF
--- a/src/any_agent/__init__.py
+++ b/src/any_agent/__init__.py
@@ -7,7 +7,9 @@ from .tracing.agent_trace import AgentTrace
 try:
     __version__ = version("any-agent")
 except PackageNotFoundError:
-    __version__ = "unknown"
+    # In the case of local development
+    # i.e., running directly from the source directory without package being installed
+    __version__ = "0.0.0-dev"
 
 __all__ = [
     "AgentConfig",


njbrake:
> Nothing strongly against this, but I would prefer to not add it. Happy to hear other opinions from @mozilla-ai/any-agent-team
> 
> I was going to write the reasons but someone already did it in StackOverflow:
> 
> https://stackoverflow.com/a/72168209
> 
> TLDR: `__version__` is just a convention, that not everyone follows, and you can instead do:
> 
> ```python
> from importlib.metadata import version; print(version("any-agent"))
> ```

I'm not against there being a __version__ attribute by I don't love the dynamic creation of the `_version.py` file. Is there a way to easily implement @daavoo's suggestion so maybe we continue to export __version__ to make it easier for people that like to access it that way, but we use the `from importlib.metadata import version` type syntax and hook it up to __version__ so that we don't need to create the _version.py file?

HareeshBahuleyan:
@daavoo Right, the alternative is fine with me - I was temporarily using that 👍 

I guess some users may attempt to access the version via `__version__` (based on convention) and get an error (like my case 😅 ).

HareeshBahuleyan:
@njbrake Agree, but I not sure how to do this in an automated manner for CI. Let me check.

HareeshBahuleyan:
@njbrake - updated the PR with the import suggested by @daavoo

src/any_agent/__init__.py
```diff
@@ -1,11 +1,19 @@
+from importlib.metadata import PackageNotFoundError, version
+
 from .config import AgentConfig, AgentFramework
 from .frameworks.any_agent import AgentRunError, AnyAgent
 from .tracing.agent_trace import AgentTrace
 
+try:
+    __version__ = version("any-agent")
+except PackageNotFoundError:
+    __version__ = "unknown"
```
njbrake: How about `__version__ = "0.0.0-dev"` or just `dev` or something like that? The package won't be found if we're in development, that's the issue right? Maybe a comment or something to explain why this happens would be useful.

HareeshBahuleyan: Done ✅

njbrake (approved):
lgtm. Thanks!

PR #494, opened 2025-06-17
feat(agent): A2A Served Agent Support for Multi-Turn

njbrake:
To keep this PR manageable, note that this does not include any changes to the a2a-as-a-tool client side implementation, it is purely the server side.

Once we feel good about the server support, I'll make another PR to add support for a multi-turn agent-to-agent conversation using a2a-as-a-tool.

### DIFF
--- a/docs/agents/index.md
+++ b/docs/agents/index.md
@@ -5,7 +5,7 @@
 To define any agent system you will always use the same imports:
 
 ```python
-from any_agent import AgentConfig, AnyAgent
+from any_agent import AgentConfig, AnyAgent, AgentRunError
 # In these examples, the built-in tools will be used
 from any_agent.tools import search_web, visit_webpage
 ```
@@ -95,11 +95,16 @@ agent = AnyAgent.create(
 ## Running Agents
 
 ```python
-agent_trace = agent.run("Which Agent Framework is the best??")
-print(agent_trace.final_output)
+try:
+    agent_trace = agent.run("Which Agent Framework is the best??")
+    print(agent_trace.final_output)
+except AgentRunError as are:
+    agent_trace = are.trace
 ```
 
 Check [`AgentTrace`][any_agent.tracing.agent_trace.AgentTrace] for more info on the return type.
+
+Exceptions are wrapped in an [`AgentRunError`][any_agent.AgentRunError], that carries the original exception in the `__cause__` attribute. Additionally, its `trace` property holds the trace containing the spans collected so far.
 
 ### Async
 
--- a/docs/api/agent.md
+++ b/docs/api/agent.md
@@ -1,3 +1,5 @@
 ## Agent
 
 ::: any_agent.AnyAgent
+
+::: any_agent.AgentRunError
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -228,9 +228,11 @@ class AnyAgent(ABC):
             ImportError: If the `serving` dependencies are not installed.
 
         Example:
+            ```
             agent = AnyAgent.create("tinyagent", AgentConfig(...))
             config = A2AServingConfig(port=8080, endpoint="/my-agent")
             agent.serve(config)
+            ```
 
         """
         from any_agent.serving import A2AServingConfig, _get_a2a_app, serve_a2a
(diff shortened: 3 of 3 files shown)


src/any_agent/serving/config.py
```diff
@@ -1,6 +1,43 @@
...
 from a2a.types import AgentSkill
 from pydantic import BaseModel, ConfigDict
 
+from any_agent.tracing.agent_trace import AgentMessage
+
+# Type alias for history formatting function
+HistoryFormatter = Callable[[list[AgentMessage], str], str]
+
+
+def default_history_formatter(messages: list[AgentMessage], current_query: str) -> str:
```
daavoo, line 12: 💅 it feels a little strange to define this here. Maybe the config can default to `None` and in `TaskManager` we do:

```pyton
self.history_formatter = config.history_formatter or deffault_history_formatter
```

No strong opinion

src/any_agent/serving/task_manager.py
```diff
@@ -0,0 +1,170 @@
...
+
+        Args:
+            config: Serving configuration containing task settings
+
+        """
+        self.config = config
+        self._tasks: dict[str, TaskData] = {}
+        self._last_cleanup: datetime | None = None
+
+    def _should_run_cleanup(self) -> bool:
```
daavoo: Is this really needed? From a quick look it looks like potentially a premature optimization.
Like, how expensive is to always check if there are tasks and are expired vs always running this (which the increased code that we need to have for it)

tests/integration/test_session_management.py
```diff
@@ -0,0 +1,144 @@
...
+from pydantic import BaseModel
+
+from any_agent import AgentConfig, AnyAgent
+from any_agent.serving import A2AServingConfig
+
+from .helpers import wait_for_server_async
+
+
+@pytest.mark.asyncio
+async def test_task_management_multi_turn_conversation():
```
daavoo: Found the test very helpful to understand the behavior ❤️

tests/integration/test_session_management.py
```diff
@@ -0,0 +1,144 @@
...
+            send_message_payload_3 = {
+                "message": {
+                    "role": "user",
+                    "parts": [{"kind": "text", "text": "My age is 30."}],
+                    "messageId": str(uuid4()),
+                    "contextId": response_1.root.result.contextId,  # Same context to continue conversation
+                    "taskId": response_1.root.result.id,
+                },
+            }
+            request_3 = SendMessageRequest(
```
daavoo: Is it worth adding an extra step to test "overriding" information? Like. "ups forget what I said, I turned 31 last week"

src/any_agent/serving/task_manager.py
njbrake: I think we'll want to optimize this at some point since the operation is O(n) runtime where n is the number of sessions. However, point taken that this may be a premature optimization and we can come back to it later. I have no problem with removing 👍

src/any_agent/serving/config.py
njbrake: The reason I put it here was to make it more transparent to the developer what we do to format the history by default. By moving it into the TaskManager then it becomes slightly less transparent what our default behavior is imo. However, I'll take some time to think about before I move forward with the current structure 👍

tests/integration/test_session_management.py
njbrake: Hmm. I don't think so? At that point its more testing out how the model responds to clarifying/changing info about the history, which seems more related to model performance and not as related to whether we are correctly giving it the past history. Let me know if you think the overriding info adds something necessary to testing whether or not the history is provided and the state managed.

daavoo: makes sense

src/any_agent/serving/task_manager.py
daavoo: > where n is the number of sessions

Not sure I understand `session` here.

njbrake: A shoot sorry, I keep reverting to my old syntax. I meant "tasks" not "sessions"

njbrake:
With the test I added, i was able to mock the LLM call (since it's not necessary to verify the behavior I wanted). Now the test takes less than half a second 🎉 

Screenshot 2025-06-18 at 10 48 28 AM

I decided to keep it as an integration test because I'm still actually serving the agent over a port, its just that the agent run call is mocked so we don't make calls to an LLM. We have other integration tests that test that run_async is working as expected so I don't bother re-testing that functionality in the test I added.

PR #499, opened 2025-06-18
Initial otel cookbook

### LINKED ISSUE #449: A2A: Create documentation/cookbook about how to access traces of a served agent
Currently, the A2A Served agent only returns the final output, it throws away the rest of the trace from the agent run. A few options (not exclusive to each other)

1. Add configuration option in A2AServingConfig to allow for agent to return the trace to the client alongside the final_output (aka response with a DataPart instead of TextPart object)
2. Store the trace in a dict (or a configurable database connection) on the server, and the server is responsible for managing/analyzing its own traces
3. Provide documentation to explain that if the developer wants to track the traces of the A2A served agent, they should add an exporter to whatever OTEL compliant log service they would like to use. Long term we probably want to document how to do this anyways, so that people can utilize the traces for general monitoring even if they don't want to use the evaluation framework.

javiermtorres:
As mentioned in the issue:

> Provide documentation to explain that if the developer wants to track the traces of the A2A served agent, they should add an exporter to whatever OTEL compliant log service they would like to use. Long term we probably want to document how to do this anyways, so that people can utilize the traces for general monitoring even if they don't want to use the evaluation framework.

This cookbook provides some information on how to use Grafana Tempo to record the traces, plus the lines needed to export the traces to this service.

Closes #449

### DIFF
--- a/.github/workflows/tests-docs.yaml
+++ b/.github/workflows/tests-docs.yaml
@@ -34,4 +34,6 @@ jobs:
           uv sync --group tests --extra all --extra a2a
 
       - name: Run Documentation tests
-        run: pytest -v tests/docs
+        # The env key is necessary for docs involving OTel
+        # see https://protobuf.dev/news/2022-05-06/#python-updates
+        run: PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python pytest -v tests/docs
--- a/docs/tracing.md
+++ b/docs/tracing.md
@@ -115,7 +115,9 @@ with open("agent_trace.json", "w", encoding="utf-8") as f:
 
 ## Adding an OpenTelemetry exporter
 
-Before starting to use the library, you can add new OpenTelemetry exporters and processors as needed. Note that this does not affect the existing processor that returns the traces to the user, and optionally prints them on `stdout`. The following code will use the OpenTelemetry Python SDK to send the agent traces to an additional endpoint using OTLP over HTTP in the indicated URL:
+Before starting to use the library, you can add new OpenTelemetry exporters and processors as needed. Note that this does not affect the existing processor that returns the traces to the user, and optionally prints them on `stdout`. Note that due to the breaking changes in the [protocol buffer Python implementation](https://protobuf.dev/news/2022-05-06/#python-updates), it may be necessary to use a pure Python implementation by using `PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python` in the environment.
+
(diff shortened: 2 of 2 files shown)


javiermtorres:
I'll keep the cookbook and add the section. I don't think it hurts having some more examples here. We can remove it later on if we really see it unnecessary.

.github/workflows/tests-docs.yaml
```diff
@@ -34,4 +34,6 @@ jobs:
           uv sync --group tests --extra all --extra a2a
 
       - name: Run Documentation tests
-        run: pytest -v tests/docs
+        # The env key is necessary for docs involving OTel
+        # see https://protobuf.dev/news/2022-05-06/#python-updates
+        run: PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python pytest -v tests/docs
```
njbrake, line 39: Can this be set in the docs test py file instead of in the github action?

javiermtorres: I'll check. I tried setting it in the notebook itself, but it was too late in the process :-/

javiermtorres: https://github.com/mozilla-ai/any-agent/pull/513

daavoo: @javiermtorres I don't understand the problem and/or the link added as reference (I didn't find much valuable information there).

Could it be just a version mismatch of the protobuf dependency in a2a? I see they have just released https://github.com/a2aproject/a2a-python/releases/tag/v0.2.9

docs/cookbook/otel_integration.ipynb
daavoo, line 1: I don't see much value in having a separate notebook for this. We don't show any screenshot or include info about what you would see in the UI and/or how to use it.

I still feel that the less cookbooks, the better.

We could just add a note about different UI alternatives to visualize OTEL traces in the docs and perhaps in the existing a2a cookbook.

njbrake: I agree, I think this cookbook should be removed.

.github/workflows/tests-docs.yaml
javiermtorres: The problem is stated in the commented link: https://protobuf.dev/news/2022-05-06/#python-updates
We could check if this would be solved with the a2a updates, but in any case we won't be using the OTLP connectors anyway since you'd prefer to remove the cookbook, so the problem just goes away.

docs/cookbook/otel_integration.ipynb
javiermtorres: I'll remove the cookbook then.

PR #501, opened 2025-06-19
fix(mcp/google): Update imports.

daavoo:
mypy was trying to tell use something but we ignored it.


daavoo:
Integration test pass: https://github.com/mozilla-ai/any-agent/actions/runs/15754818075

Issue #502, opened 2025-06-19
Revisit return content of `a2a_tool`


daavoo:
I think we need to cleanup what we the agent as a tool returns to only give whatever information is necessary.
We are using a lot of tokens for no reason and it might make it harder for the agent calling this to extract the relevant information.

I actually thought about this already when looking at the traces of the a2a cookbook but forgot to open an issue

_Originally posted by @daavoo in https://github.com/mozilla-ai/any-agent/pull/498#discussion_r2156605577_

njbrake:
How about for starters, we get rid of the `history` key? The A2A server returns this value so that a client is aware of the history of the task that they are interfacing with. However, if it's agent to agent communication, the client agent is already aware of the history, because it's the one that is repeatedly calling the task. 

that's the key that uses the most tokens, I propose removing it via https://github.com/mozilla-ai/any-agent/pull/503.

PR #503, opened 2025-06-19
Improve reliability of test, decrease token use of

njbrake:
* make the assert an 'in' instead of '==' because nano was appending `1. ` to the beginning of the prompt which is ok because I guess technically I did have those numbers in the prompt, although mini wasn't including them and nano wasn't including them yesterday 🤷 

I also excluded the `history` key from what the a2a_tool returns to reduce token usage and to focus on only providing back to the client agent the content of the result.

### DIFF
--- a/src/any_agent/tools/a2a.py
+++ b/src/any_agent/tools/a2a.py
@@ -84,7 +84,10 @@ async def a2a_tool_async(
                 )
             elif hasattr(response.root, "result"):
                 result = response.root.result.model_dump_json(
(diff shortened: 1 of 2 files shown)


njbrake:
> I would still try to come up with a more clean format

Agree. I'll refactor to extract specific values and create a nice paragraph. That can be a future PR that will close out https://github.com/mozilla-ai/any-agent/issues/502 which we should keep open until then.

PR #505, opened 2025-06-19
feat(evaluation): Add unique ID validation for checkpoints

njbrake:
This way you can get back the checkpoint result from the result object by keeping track of the id you created.

I know we're going to clean up the eval codebase to turn it back into more primitives so the EvaluationCase may get removed but at least this will still be relevant for the CheckpointCriteria model.


src/any_agent/evaluation/evaluators.py
```diff
@@ -91,4 +92,5 @@ def evaluate_final_output(
         reason=f"Partial Match (F1) score is {round(f1_score, 2)}",
         criteria="Is the answer a direct match?",
         points=1,
+        id="final_output",
```
javiermtorres, line 95: What about using a distinct uuid (like https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_8_(custom)) to mark this kind of condition? Or maybe a boolean flag?

njbrake: I hardcode it because in this version of the code, there will only be a single final_output evaluation result. This code will likely go away with #507 I'll leave it this way for now

Issue #448, opened 2025-06-09
Evaluation API discussion


daavoo:
I wanted to have an open discussion about how the evaluation API should look like.

In my mind, I see 3 different types of "checks" I would like to use to evaluate agent behavior, with increasing level of flexibility and decreasing level of determinism:

- Code checks

  You want to assess something that can be easily written in Python using `trace` helpers.

```python

def check_non_default_tools(trace: AgentTrace) -> AgentOutput:
    if any(
        span.is_tool_execution()
        and span.attributes.get("gen_ai.tool.name")
        and span.attributes.get("gen_ai.tool.name") not in DEFAULT_TOOLS
        for span in trace.spans
    ):
        return AgentOutput(
            passed=True,
            reasoning="Found used tools that are not in the DEFAULT_TOOLS.",
        )
    return AgentOutput(
        passed=False,
        reasoning="Didn't find used tools that are not in the DEFAULT_TOOLS.",
    )
```

- LLM checks

    You need an LLM for assessment, but the check has a clear context that can be extracted with `trace` helpers.

```python
from litellm import completion

def check_final_output_aligns_with_description(trace: AgentTrace, characteristic: str, description: str) -> AgentOutput:
    system_prompt = f"""
    You will be given an user input for which you need to evaluate whether the given criteria is met.

    # Criteria
    
    Does the final output description of what the user did yesterday align with this characteristic: {characteristic}? 
    Here's a description of what that characteristic would look like: {description}.

    # Expected output
    
    Please conform your output to the following schema: {AgentOutput.model_json_schema()}."
    """

    response = completion(
        model=model,
        messages=[
            {"role": "system", "content": system_prompt},
            {"role": "user", "content": trace.final_output}
        ],
    )
    return AgentOutput.model_validate_json(response.choices[0].message["content"])
```

- Agent Checks

  You need an LLM for assessment, but the context might vary from trace to trace and/or your criteria description is intentionally "vague".

```python
def check_hallucination(trace: AgentTrace) -> AgentOutput:
    instructions = """
    You are a helpful assistant that will be used to evaluate the correctness of an agent trace. 

    # Expected workflow

    Given a specific question regarding the quality of the something about the agent, utilize the appropriate tools in order to gather the information needed in order to accurately answer the question. 

    # Expected output
    
    Please conform your output to the following schema: {AgentOutput.model_json_schema()}."
    """
    
    trace_tools = get_trace_tools(trace)
    agent_config = AgentConfig(
[...]

PR #507, opened 2025-06-19
feat: Slim down the evaluation API and bring it "back to basics"

njbrake:
As we discussed in #448, the current evaluation API has a lot of what might be considered to be application level logic for running an evaluation. Although in the future we would like to help ease the developers ability to run and maintain an evaluation suite, maybe it make sense for version 1.0.0 to slim this down into a single `AgentAsJudge` class to ease the most burdensome part of evaluating the trace, which is building an agent that can selectively access different tools/pieces of the trace to answer the question.

Thoughts? Have I gone too far and removed too much?

If you like this plan of stripping out a bunch of this code, I do realize that I will need to refactor and clean the docs and tests before this is truly ready for review, but I wanted to get this up here to aid our conversation.


njbrake:
@stefanfrench @mozilla-ai/any-agent-team is there anyone else that should be involved in this decision?

daavoo:
> Thoughts? Have I gone too far and removed too much?

I think the changes are a great step back.

`AgentAsJudge` feels like a really good proposed value of using any-agent and valuable on it's very own.

I still think that we could still want to provide a `LlmAsJudge` helper for simple "nlp assertions".

I don't think we need anything for "code assertions" beyond documentation.

stefanfrench:
@njbrake Slimming it down to `AgentAsJudge` feels good to me from a user perpective

@daavoo if you ran `AgentAsJudge` on simple NLP assertions, I guess it would still work but be very inneficient right? I'm not against  including`LlmAsJudge`, provided in the docs we give some clear example recommendations of the simple NLP assertions its better for.

javiermtorres:
I agree with keeping the functionality as evaluation functions, and removing an evaluation framework of our own, since there are already options out there.

If we acted on the original otel spans, and not our own models, we could import this as is in a PySpark deployment, for example, when doing batch evaluation. Maybe we can think of a way of using both inputs.

EDIT: Note that this would mean our evaluation tools would be usable for any-agent (pun intended) conforming to the otel genai semantics out of the box.

src/any_agent/evaluation/tools.py
```diff
@@ -0,0 +1,101 @@
...
+from typing import Any
+
+from pydantic import BaseModel
+
+from any_agent.tracing.agent_trace import AgentTrace
+
+MAX_EVIDENCE_LENGTH: int = 50
+
+
+class AgentEvaluationTools:
```
javiermtorres: I don't see the functions as being part of a class. The trace is not really a state, but rather an input of each function, afaict, so it won't change across invocations. Putting the functions at the module level is imho simpler and cleaner.

src/any_agent/evaluation/agent.py
```diff
@@ -1,121 +1,85 @@
...
-            str | BaseModel | None: The final output of the agent
+{AgentOutput.model_json_schema()}
+"""
 
-        """
-        return self.trace.final_output
 
-    def get_tokens_used(self) -> int:
-        """Get the number of tokens used by the agent as reported by the trace.
+class AgentAsJudge:
```
javiermtorres: Note the same issue with using a class, as related below. The class attributes will not last, or be modified, across invocations. A higher level function or just adding the current attributes as params is imho again simpler and cleaner.

src/any_agent/evaluation/agent.py
```diff
@@ -1,121 +1,85 @@
...
 from any_agent import AgentConfig, AnyAgent
+from any_agent.config import AgentFramework
 from any_agent.evaluation.schemas import AgentOutput
+from any_agent.evaluation.tools import AgentEvaluationTools
 from any_agent.tracing.agent_trace import AgentTrace
+from any_agent.utils.asyncio_sync import run_async_in_sync
 
-MAX_EVIDENCE_LENGTH: int = 500
-
+AGENT_INSTRUCTIONS = f"""You are a helpful assistant that will be used to evaluate the correctness of an agent trace. Given a specific question regarding the quality of the something about the agent, utilize the appropriate tools in order to gather the answer needed in order to accurately answer the question. If you are asked anything about what the agent did, you should strongly consider using the get_evidence_from_spans tool to get the evidence. However, if the question is about specific details of the agent's actions, you don't necessarily need to use the get_evidence_from_spans tool.
```
javiermtorres: Do you have preliminary results of how this will work in a real invocation? It would be really good to check it out!!!

src/any_agent/evaluation/tools.py
njbrake: Since this is being passed to an agent as a tool, I structure it this way to avoid the agent needing to pass the trace object around. The function I give the agent have the trace "baked in" so all the agent needs to do is call the function, it doesn't have to worry about getting the parameters right 👍

src/any_agent/evaluation/agent.py
njbrake: Totally! Our ongoing work with evaluations should help to give us some good feedback about how this works.

src/any_agent/evaluation/agent.py
javiermtorres: Just to clarify, I'd use a curried or higher order function for this instead of classes, but as far as it works, fine for me.

docs/evaluation.md
```diff
@@ -1,82 +1,185 @@
...
-    ```bash
-    any-agent-evaluate \
-        --evaluation_case_path "docs/examples/evaluation_case.yaml" \
-        --trace_path "tests/assets/OPENAI_trace.json"
-    ```
+- Basic pass/fail assessments
+- Simple criteria checking
+- Text-based evaluations
+
+### Example: Basic LLM Judge
```
daavoo: Not a blocker: we should try to come up with a more real/useful example to showcase

src/any_agent/evaluation/llm_judge.py
```diff
@@ -0,0 +1,132 @@
...
+2. "reasoning": Brief explanation for your decision (2-3 sentences max)
+
+Your response MUST be valid JSON matching this schema:
+{
+  "passed": boolean,
+  "reasoning": "string"
+}"""
+
+
+class LlmJudge:
```
daavoo, line 37: No strong opinion:

In my mind, LlmJudge was a much less opinionated API: just passing (content, assertion) and letting the user decide on those values.

The current implementation is basically AgentJudge with no loop or tools (so just worse?).

I would prefer if LlmJudge could be focused on simple natural language assertions for arbitrary chunks of text, and we document to the users how to get some of those useful chunks (by documenting examples for final_output, and maybe even extracting part of the trace)

docs/evaluation.md
```diff
@@ -1,82 +1,185 @@
 # Agent Evaluation
 
-!!! warning
+The any-agent evaluation module provides two primitives for evaluating agent traces using LLM-as-a-judge techniques:
```
daavoo: Not a blocker, I would like to comment at this level the 3 alternatives I mentioned (we are missing an explicit mention to "code checks" or whatever).

Maybe a simple table of `Method - Pros - Cons` to better showcase when you might want to use each

src/any_agent/evaluation/llm_judge.py
njbrake: Same as you, no strong opinion from me here either. So maybe I update the run method so that instead of passing in the agent_trace and the question, I instead have the user pass in an arbitrary 'context' string instead of the trace, and have an example of some possible use cases. Makes sense to me 👍

njbrake:
docs/integration/cookbook tests passing 
Screenshot 2025-06-24 at 12 06 45 PM

src/any_agent/evaluation/tools.py
```diff
@@ -0,0 +1,78 @@
...
+    def get_tokens_used(self) -> int:
+        """Get the number of tokens used by the agent as reported by the trace.
+
+        Returns:
+            int: The number of tokens used by the agent
+
+        """
+        return self.trace.tokens.total_tokens
+
+    def get_number_of_steps(self) -> int:
```
daavoo, line 47: 💅 feels more consistent with `tokens_used`

```suggestion
    def get_steps_taken(self) -> int:
```

tests/integration/test_load_and_run_agent.py
```diff
@@ -115,60 +112,41 @@ def assert_tokens(agent_trace: AgentTrace) -> None:
...
-            passed=False, reasoning="Didn't find `write_file` tool usage."
-        )
-
-    # Create checkpoints with mix of auto-generated and custom IDs
-    checkpoint_with_auto_id = CheckpointCriteria(
-        criteria=_check_if_agent_called_write_file,
-        points=1,
+    """Test evaluation using the new judge classes."""
+    # Test 1: Check if agent called write_file tool using LlmJudge
+    llm_judge = LlmJudge(model_id="gpt-4.1-nano")
```
daavoo, line 117: I don't think we should use LlmJudge for this

njbrake: I assume you don't like that the LlmJudge is checking something that could easily be checked by a function that looks at the list of messages and looks if any are write_file tool calls, right?

In practice you should write a callable to check it, but the purpose at this point is just to exercise the LlmJudge code. I'll make a separate issue so we can track whether or not you want to remove this code and write a separate integration test for the LlmJudge, or make this a more complex criteria that is more suitable for an LlmJudge.

Issue #506, opened 2025-06-19
Missing `openai-agents` dependency causes runtime ImportError


alexmeckes:
# Bug Report: Missing `openai-agents` dependency causes runtime ImportError

## Summary

The `any-agent` package (v0.21.0) fails at runtime with `ModuleNotFoundError: No module named 'agents'` due to a missing dependency declaration. The package imports from `openai-agents` but doesn't declare it as a dependency. Pretty easy to fix and find the module, but is a bad experience for first time users.

## Environment

- **any-agent version**: 0.21.0
- **Python version**: 3.13
- **OS**: macOS 
- **Installation method**: `pip install any-agent`

### Expected Behavior
After installing `any-agent`, basic imports should work:
```python
from any_agent import AnyAgent  # Should work
```

### Actual Behavior
```python
from any_agent import AnyAgent
# Traceback (most recent call last):
#   File "<string>", line 1, in <module>
#   File ".../any_agent/__init__.py", line 4, in <module>
#     from .frameworks.any_agent import AgentRunError, AnyAgent
#   File ".../any_agent/frameworks/any_agent.py", line 18, in <module>
#     from any_agent.tracing.instrumentation import (
#   File ".../any_agent/tracing/instrumentation/__init__.py", line 11, in <module>
#     from .openai import _OpenAIAgentsInstrumentor
#   File ".../any_agent/tracing/instrumentation/openai.py", line 7, in <module>
#     from agents.tracing import TracingProcessor
# ModuleNotFoundError: No module named 'agents'
```

### 1. Missing Dependency Declaration
```bash
$ pip show any-agent
Requires: duckduckgo_search, fire, litellm, markdownify, mcp, opentelemetry-exporter-otlp,
         opentelemetry-sdk, pydantic, requests, rich, tavily-python
```

**Missing**: `openai-agents` - despite the code importing from it!

### 2. Import Chain Evidence
The failure occurs in `any_agent/tracing/instrumentation/openai.py`:
```python
from agents.tracing import TracingProcessor  # Line 7 - fails without openai-agents
```

The `agents` module comes from the `openai-agents` package, which provides the OpenAI Agents SDK. 

## Reproduction Steps

1. Create a fresh virtual environment:
  ```bash
  python -m venv test_env
  source test_env/bin/activate  # or test_env\Scripts\activate on Windows
  ```

2. Install any-agent:
  ```bash
  pip install any-agent==0.21.0
  ```

3. Try to import:
  ```python
[...]

PR #509, opened 2025-06-20
tests: Run more env variations.

daavoo:
- Split unit and integration workflows.
- For integration add new test using "barebones" installation. This should catch things like #506
- For unit, run in a matrix of supported minor Python versions (3.11, 3.12, 3.13) and OS (macos, ubuntu, windows).


njbrake:
Should we use https://nox.thea.codes/en/stable/index.html

daavoo:
> Should we use https://nox.thea.codes/en/stable/index.html

no strong opinion.

I would prefer to not use it unless we want to run the matrix of Python versions locally (which I personally don't want).

I like when it's done at the GitHub level because:

- It decouples things and doesn't require setting up a new tool (we still run `pytest` in the CI) 
- The workers take care of spinning different Python envs **and** OS variations, which you can't really do out of the box (or maybe you just can't) in nox.

PR #513, opened 2025-06-24
Move protobuf settings for docs tests

javiermtorres:


javiermtorres:
@njbrake wdyt?

njbrake (changes requested):
Please put it in the docs tests, since this only applies for the docs. Also, before this is merged please run the "cookbook"' tests to ensure that the cookbook tests are passing, since I see a failed cookbook test run on main. Thanks!

javiermtorres:
> Please put it in the docs tests, since this only applies for the docs.

Unfortunately, this does not seem possible :-/ The issue lies in https://github.com/mozilla-ai/any-agent/blob/dbad5d5e65704f1c966009f2b429af4e5fb13f23/tests/conftest.py#L18
At that line, the otel sdk will be imported, so we cannot set the env var later than this line (which would be the case if we used a separate conftest). Do you think we can modify the import order for our conftest files?

njbrake:
@javiermtorres can you confirm that this change is required? I'm running the docs test without that env var set and everything is fine. 

Screenshot 2025-06-30 at 5 56 43 AM

njbrake:
I am ok with it staying in the conftest but first please confirm that this env var is truly needed.

javiermtorres:
> I am ok with it staying in the conftest but first please confirm that this env var is truly needed.

Let me take a closer look at this, yes. It could be that:

* The newest A2A SDK changed the protobuf version so it's not colliding with the version in otel
* We removed the otel examples in docs (we removed the cookbook, at least)

I'll keep you posted :)

javiermtorres:
Confirmed, apparently the A2A SDK v0.2.10 has changed the protobuf version. The PR will then remove this env var in the conftests.

njbrake (changes requested):
Please also fix the comment above the removal of the env var, and fix the docs at 
`docs/tracing.md` in the `## Adding an OpenTelemetry exporter` section

javiermtorres:
Re: the docs, I thought about them. I'd rather keep them as they are, since they state "it may be necessary", which is still true and will help the user in case of issues.

daavoo:
> Re: the docs, I thought about them. I'd rather keep them as they are, since they state "it may be necessary", which is still true and will help the user in case of issues.

I would prefer to just drop it from the docs. I don't want to accumulate (potentially) outdated info, specially if it's not related to any-agent but some upstream deps. Someone should be easily able to find the answer if they hit the bug (and if they hit the bug, then there is either something wrong in their env or something we should fix in our pinning)

javiermtorres:
Ok, updating docs, removing comment and merging

PR #515, opened 2025-06-24
Fix protobuf for the otel cookbook

javiermtorres:


javiermtorres:
The OTEL cookbook has been removed.

PR #522, opened 2025-06-25
fix: enhance AgentRunError to be readable and include original exception details 

njbrake:
Updated the AgentRunError class to accept and store the original exception, improving error handling. The string representation now reflects the original exception message for better debugging.

was:
Screenshot 2025-06-25 at 6 05 08 AM

now:
Screenshot 2025-06-25 at 5 57 57 AM

### DIFF
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -35,13 +35,27 @@ class AgentRunError(Exception):
     """Error that wraps underlying framework specific errors and carries spans."""
 
     _trace: AgentTrace
+    _original_exception: Exception
 
-    def __init__(self, trace: AgentTrace):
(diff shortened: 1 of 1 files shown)


javiermtorres (approved):
LGTM. I assume this doesn't mess up with the __cause__ and other exception related mechanisms.

src/any_agent/frameworks/any_agent.py
```diff
@@ -35,14 +35,30 @@ class AgentRunError(Exception):
     """Error that wraps underlying framework specific errors and carries spans."""
 
     _trace: AgentTrace
+    _original_exception: Exception
 
-    def __init__(self, trace: AgentTrace):
+    def __init__(self, trace: AgentTrace, original_exception: Exception):
```
javiermtorres, line 40: Speaking about this... could we use the `__cause__` instead of another param?

PR #538, opened 2025-06-27
feat: Refactor server handling in A2A and MCP serving

njbrake:
Right now, handling an async served agent is tricky, you get back an async task and uvicorn server instance, and need to undderstand what to do with them. My suggestion here is to abstract this out behing a typed pydantic model so that people can have clean ways to shutdown and access information about the server. The task and server items are still there, so power users still have access to them if they want them.

- Introduced `ServerHandle` class for better management of server lifecycle.
- Updated `serve_async` methods to return `ServerHandle` instead of tuples.
- Adjusted related code in integration tests to utilize the new server handle.
- Cleaned up output handling and execution counts in Jupyter notebooks for clarity.
- Removed unnecessary output and improved shutdown procedures for servers.


src/any_agent/frameworks/any_agent.py
```diff
@@ -23,14 +25,79 @@
...
 
-    import uvicorn
     from opentelemetry.trace import Tracer
-    from pydantic import BaseModel
 
     from any_agent.serving import A2AServingConfig, MCPServingConfig
     from any_agent.tools.mcp.mcp_server import _MCPServerBase
 
 
+class ServerHandle(BaseModel):
```
njbrake: I don't love having the model defined here. However, I can't yet define it in the server folder, because the __init__.py of that folder tries to import a2a stuff which may not exist... I'll handle this organization via another PR.

daavoo: > because the init.py of that folder tries to import a2a stuff which may not exist

what do you mean? what is the thing being imported?

njbrake: I think we have a bug in our code where you still can't serve via MCP if you don't have a2a installed. As soon as you try to import anything from our serving folder it runs the __init__

https://github.com/mozilla-ai/any-agent/blob/main/src/any_agent/serving/__init__.py

And that will throw an error if a2a isn't imported. 

Right now, that's ok because we only import serving code in any_agent.py from either inside the TYPE_CHECKING block or in the function where it's being used (aka serve()). But since ServerHandle has to be imported since we create ServerHandle objects in any_agent.py, now we hit the bug.

src/any_agent/frameworks/any_agent.py
```diff
@@ -23,14 +25,79 @@
...
+        """Check if the server is still running.
+
+        Returns:
+            True if the server task is still running, False otherwise.
+
+        """
+        return not self.task.done()
+
+    @property
+    def port(self) -> int:
```
daavoo: this feels like 🤸 to get a value we define/require at the serving config, or am I missing something?

njbrake: Yeah. If you set port 0, then uvicorn finds an open port and assings it for you, but then you do this to find out which port it assigned. I can add a comment explaining that.

src/any_agent/frameworks/any_agent.py
daavoo: > ut since ServerHandle has to be imported since we create ServerHandle objects in any_agent.py, now we hit the bug.

why don't we just create the ServerHandle objects and return them as part of `serve_a2a` and `serve_mcp`?

njbrake: Great point. I'll do that.

src/any_agent/serving/server_handle.py
```diff
@@ -0,0 +1,79 @@
...
+import asyncio
+from typing import Any
+
+from pydantic import BaseModel, ConfigDict
+from uvicorn import Server as UvicornServer  # noqa: TC002
+
+from any_agent.logging import logger
+
+
+class ServerHandle(BaseModel):
```
javiermtorres: I don't see the need to make it a pydantic model. It won't be serialized, it doesn't have complex validation. A plain typed class should do.

src/any_agent/serving/server_handle.py
```diff
@@ -0,0 +1,79 @@
...
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    async def shutdown(self, timeout_seconds: float = 10.0) -> None:
+        """Gracefully shutdown the server with a timeout.
+
+        Args:
+            timeout_seconds: Maximum time to wait for graceful shutdown before forcing cancellation.
+
+        """
+        if not self.is_running():
```
javiermtorres, line 30: I'd avoid adding more state vars than needed. We should be able to use https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.done to make this decision.

njbrake: the self.is_running is a wrapper around the async task done function 😄 . I'd like to keep the is_running helper so that developers can easily access it

src/any_agent/serving/server_handle.py
njbrake: Good point! I'll switch it away from being a pydantic model in favor of something simpler 👍

njbrake:
Integration tests https://github.com/mozilla-ai/any-agent/actions/runs/15970545955

src/any_agent/serving/server_handle.py
javiermtorres: Ah ok!!!!! I thought it was a separate member 👍🏼

PR #540, opened 2025-06-27
fix: need to create uuid by calling it

njbrake:


daavoo (approved):
ups 👼 

I guess this is hard to catch in tests? how did you find out?

njbrake:
> ups 👼
> 
> I guess this is hard to catch in tests? how did you find out?

It was missed because you can cast a function to a str 😆 

>>> str(uuid4)
'<function uuid4 at 0x10511b380>'

So technically this was a valid id, just not the one we expected. I spotted it from visual inspection while debugging other a2a issues.

daavoo:
> So technically this was a valid id, just not the one we expected. I spotted it from visual inspection while debugging other a2a issues.

What I meant is that if we call this twice there would be 2 sendMessage with the same `id`, but it should be an unique `id` and I am unsure what effects this have in a2a (if it has any effect)

PR #542, opened 2025-06-27
fix: avoid repeating context

njbrake:


njbrake:
Hmmm. Ooh let me think about this over the weekend. This change makes it so that the context is only the previous conversation input and output, but I think we probably want to include a summary of all the steps the served agent took in between...

I asked for review a bit prematurely I think, sorry about that.

njbrake:
Ok. I went with a design where we continue to provide as context all of the steps of the trace that the agent preformed, not just the inputs/outputs of the agent request, same as we have been doing 👍 . This was most comprehensive way to include all of the previous history.

njbrake:
Integration Tests: https://github.com/mozilla-ai/any-agent/actions/runs/15974408697/job/45052992051

PR #544, opened 2025-06-30
Initial docs for MCP serving usage

javiermtorres:
A small section describing how to set up and use MCP-served agents is added in this PR.


javiermtorres:
Frankly there's not much to add to the docs... The examples are fairly short and self-explanatory, and there are fewer options here than for A2A. @njbrake would you miss any info for users?

docs/serving.md
```diff
@@ -146,12 +148,103 @@ config = A2AServingConfig(
...
+
+    agent_trace = await main_agent.run_async(prompt)
+
+    print(agent_trace)
+
+if __name__ == "__main__":
+    asyncio.run(main())
+```
+
+If both are specified, then only the Google expert is contacted. If only the OpenAI expert is left, it is contacted, but the expert will inform the main agent that it cannot answer questions about the Google ADK.
```
njbrake: This can probably removed, since it might add confusion.
```suggestion
```

njbrake (approved):
I would recommend removing the final paragraph (see my other comment) but other than that, looks good, thanks!

PR #549, opened 2025-06-30
Run Cookbook Tests in pytest, in parallel

njbrake:
Make it easier to run cookbook tests locally, similar to integration/unit/docs. Also enables pytest xdist so the notebooks can get run in parallel.


njbrake:
Cookbook tests: https://github.com/mozilla-ai/any-agent/actions/runs/15977191040/job/45062644226

PR #551, opened 2025-06-30
Fix Agno instrumenters

javiermtorres:
Agno 1.7.0 changed the signature of one function we instrument to build our otel spans. Apparently the process is more or less the same (mutating the input params), but the mutated params are not returned. The tests pass and the spans look correct but I'd appreciate a second look, or at least a pointer to a test that could give away possible errors.


daavoo:
> or at least a pointer to a test that could give away possible errors.

You can just run the integration tests only with the version bump and there should be clear failures

Issue #437, opened 2025-06-06
Allow to redirect `gen_ai.input` and `gen_ai.output` to an external destination


daavoo:
Discussed with @davidmanzanoai .

It would be nice for us to introduce a flexible way for users to specify a target location to store potentially sensitive fields (inputs/outpus) and use a reference in the "plain" trace instead of directly storing the fields.

It appears that this idea is already considered and WIP as part of the OTEL semantic conventions (https://github.com/open-telemetry/semantic-conventions/pull/2179).

javiermtorres:
We could use a custom https://opentelemetry-python.readthedocs.io/en/latest/sdk/trace.html#opentelemetry.sdk.trace.SpanProcessor to handle this information. Wdyt?

PR #554, opened 2025-07-01
feat: Introduce `any_agent/callbacks`.

daavoo:


src/any_agent/config.py
```diff
@@ -141,6 +143,12 @@ class AgentConfig(BaseModel):
     See more info at [Tools](../agents/tools.md).
     """
 
+    callbacks: list[Callback] | None = []
+    """List of callbacks to use during agent invocation.
+
+    See more info at [Callbacks](../agents/callbacks.md).
```
njbrake, line 149: Can this link be changed so that it points to something a developer can easily find even if they're not reading this inside the served docs?

src/any_agent/callbacks/span_end.py
```diff
@@ -0,0 +1,36 @@
...
+
+if TYPE_CHECKING:
+    from opentelemetry.trace import Span
+
+    from any_agent.tracing.agent_trace import AgentTrace
+
+
+def _span_end(context: dict[str, Any], operation_name: str) -> None:
+    trace_id = get_current_span().get_span_context().trace_id
+    current_span: Span = context[f"{operation_name}-{trace_id}"]
```
javiermtorres: Why not just use a `current_span` context key? Do you see the context being reused across runs?

javiermtorres: Actually, can we use `get_current_span()` directly?

src/any_agent/callbacks/span_generation/agno.py
```diff
@@ -0,0 +1,144 @@
...
+        span = tracer.start_span(f"call_llm {model_id}")
+        span.set_attributes(
+            {
+                "gen_ai.operation.name": "call_llm",
+                "gen_ai.request.model": model_id,
+            }
+        )
+
+        trace_id = span.get_span_context().trace_id
+        if trace_id not in self.first_llm_calls:
```
javiermtorres: The handling of `first_llm_calls` would seem a candidate for a base class method, since it's shared by all the different implementations.

src/any_agent/callbacks/span_end.py
```diff
@@ -0,0 +1,36 @@
...
+    current_span.end()
+
+    trace: AgentTrace = context["running_traces"][trace_id]
+    trace.add_span(current_span)
+
+
+class SpanEndCallback(Callback):
```
javiermtorres, line 12: I'd make this part of the base class for the different tracing callbacks, instead of a separate class. At least, we could replace the magic strings like `"call_llm"` with constants to keep consistency across callback classes.

src/any_agent/callbacks/base.py
```diff
@@ -0,0 +1,28 @@
...
+from __future__ import annotations
+
+from typing import Any
+
+
+class Callback:
+    """Base class for all callbacks."""
+
+    def before_llm_call(
+        self, context: dict[str, Any], *args, **kwargs
```
javiermtorres: I see a lot of places for params: the context, args and kwargs. What would be the guidelines to place what information where? Can it be simplified?

src/any_agent/callbacks/span_generation/google.py
```diff
@@ -0,0 +1,172 @@
...
+    from google.adk.tools.tool_context import ToolContext
+    from opentelemetry.trace import Span, Tracer
+
+
+def _set_llm_output(llm_response: LlmResponse, span: Span) -> None:
+    content = llm_response.content
+    if not content:
+        return
+    if not content.parts:
+        return
```
javiermtorres: Nitpicking:

```suggestion
    if not content or not content.parts:
        return
```

src/any_agent/callbacks/span_generation/google.py
```diff
@@ -0,0 +1,172 @@
...
+
+    if content.parts[0].text:
```
javiermtorres: There's a very corner case where the text could be "", but I don't really think it's likely at all, right?

src/any_agent/callbacks/span_generation/google.py
```diff
@@ -0,0 +1,172 @@
...
+            }
+        )
+    else:
+        span.set_attributes(
+            {
+                "gen_ai.output": json.dumps(
+                    [
+                        {
+                            "tool.name": getattr(part.function_call, "name", "No name"),
+                            "tool.args": getattr(part.function_call, "args", "No args"),
```
javiermtorres: Maybe we could use a dict literal convention?

```suggestion
                            "tool.args": getattr(part.function_call, "args", "{}"),
```

src/any_agent/callbacks/span_generation/google.py
```diff
@@ -0,0 +1,172 @@
...
+        span: Span = tracer.start_span(
+            name=f"execute_tool {tool.name}",
+        )
+        span.set_attributes(
+            {
+                "gen_ai.operation.name": "execute_tool",
+                "gen_ai.tool.name": tool.name,
+                "gen_ai.tool.description": tool.description,
+                "gen_ai.tool.args": json.dumps(args),
+                "gen_ai.tool.call.id": getattr(
```
javiermtorres: Moving to some base class or using constants would seem appropriate here in most of `before_tool_execution` implementations.

src/any_agent/callbacks/span_generation/langchain.py
```diff
@@ -0,0 +1,161 @@
...
+    from langchain_core.messages import BaseMessage
+    from langchain_core.outputs import LLMResult
+    from opentelemetry.trace import Span, Tracer
+
+
+def _set_llm_input(messages: list[list[BaseMessage]], span: Span) -> None:
+    if not messages:
+        return
+    if not messages[0]:
+        return
```
javiermtorres: Same nitpicking :)

src/any_agent/callbacks/span_generation/llama_index.py
```diff
@@ -0,0 +1,148 @@
...
+            )
+    tool_calls: list[ToolSelection]
+    if tool_calls := output.tool_calls:
+        span.set_attributes(
+            {
+                "gen_ai.output": json.dumps(
+                    [
+                        {
+                            "tool.name": getattr(tool_call, "tool_name", "No name"),
+                            "tool.args": getattr(tool_call, "tool_kwars", "{}"),
```
javiermtorres: does the different key mean that the tool calling conventions in llama_index are somehow different from the other frameworks?

src/any_agent/callbacks/wrappers/agno.py
```diff
@@ -0,0 +1,70 @@
...
+
+        self._original_aprocess_model = agent._agent.model._aprocess_model_response
+
+        async def wrapped_llm_call(*args, **kwargs):
+            for callback in agent.config.callbacks:
+                self.context = callback.before_llm_call(self.context, *args, **kwargs)
+
+            result = await self._original_aprocess_model(*args, **kwargs)
+
+            for callback in agent.config.callbacks:
```
javiermtorres, line 33: Using the "complete" callback as the minimum unit prevents us from, for example, handling resources properly across callbacks. If we have callbacks CB1 and CB2, then the order will always be:

* CB1-before
* CB2-before
* &lt;call&gt;
* CB1-after
* CB2-after

whereas we might want CB2-after to run before CB1-after to remove resources in the proper order.

javiermtorres (commented):
My main concerns are the order of callbacks, as explained in one of the comments, and whether we can remove some boilerplate code by using inheritance or higher order functions.

I'd also appreciate having some example of another creative use of callbacks (e.g. "context engineering" by tweaking the llm response to be added to the list of messages, somehow???)

src/any_agent/callbacks/span_end.py
daavoo: > Why not just use a current_span context key? Do you see the context being reused across runs?

Because the context is at the instance level, it could be shared across runs (i.e. the asyncio.gather test we have).
Maybe I should instead rethink how to set up a separate context for each run instead of relying on the current instance-level shared context.

> Actually, can we use get_current_span() directly?

I didn't verify it but my thinking was that we can't do that because a framework might start arbitrary spans and get_current_span could return the one that is created by the framework.

src/any_agent/callbacks/span_generation/agno.py
daavoo: Good point, I should probably create a base SpanGeneration

src/any_agent/callbacks/span_end.py
daavoo: I deliberately wanted to keep the span end as a separate callback because my idea was that additional callbacks would run in between `SpanGeneration` and `SpanEnd`, so you could modify the span properties via callbacks before they are ended (and thus, exposed to OTEL exporters).

For example, I wanted to implement #437 that way.

src/any_agent/callbacks/span_end.py
javiermtorres: > Maybe I should instead rethink how to set up a separate context for each run instead of relying on the current instance-level shared context.

At least, document very explicitly (maybe typing could help here?) the scope of the context. I understood that the context would be used in a single run.

javiermtorres: > I didn't verify it but my thinking was that we can't do that because a framework might start arbitrary spans and get_current_span could return the one that is created by the framework.

Comment to spawn an issue :) Check how spans in the otel sdk are handled with different libraries operating, and verify that frameworks conform to this usage.

src/any_agent/callbacks/span_end.py
javiermtorres: My concern is whether some of these use cases could be solved with standard span processors. For example, one thing that processors can't do easily is to process the trace "along the run", keeping some state about it. #437 could be implemented that way, but as it is described right now, it can be done with otel processing before exporting.

src/any_agent/callbacks/span_generation/google.py
daavoo: Not sure I follow how (or maybe I am misunderstanding what you mean). The variables from which you extract the values vary from framework to framework (and sometime the keys) so what could be abstracted/reused?

src/any_agent/callbacks/span_generation/llama_index.py
daavoo: Not sure what you mean. These keys (LLM response that includes tool call) are the same in all frameworks

src/any_agent/callbacks/span_end.py
daavoo: > My concern is whether some of these use cases could be solved with standard span processors.

Absolutely. 
My thinking is that things would be easier to configure, maintain and teach (for users on how to add their custom stuff) if we use a single way of implementing things. 

I want to implement everything as a callback simply because it is possible (I also want to replace the console printing with a callback.), whereas stuff like guardrails or what you described can't be implemented via OTEL exporters / processors so we would end with 2 different ways of implementing stuff.

javiermtorres: Ok, let's got his way then 👍🏼

In any case :D please use constants for the trace strings

src/any_agent/callbacks/span_generation/google.py
javiermtorres: Note that the whole:

```
        tracer: Tracer = context["tracer"]

        span: Span = tracer.start_span(
            name=f"execute_tool {tool.name}",
        )
        span.set_attributes(
            {
                "gen_ai.operation.name": "execute_tool",
                "gen_ai.tool.name": tool.name,
                "gen_ai.tool.description": tool.description,
                "gen_ai.tool.args": json.dumps(args),
                "gen_ai.tool.call.id": getattr(
                    tool_context, "function_call_id", "no_id"
                ),
            }
        )

        trace_id = span.get_span_context().trace_id
        context[f"execute_tool-{trace_id}"] = span

        return context
```

Can be further decomposed into

```
build_span(extract_tool_info(*args, **kwargs))
# extract_tool_info(*args, **kwargs) -> (name, description....)
```

where build_span does the otel specific format independently from the implementation, and extract_info would be particular for each implementation. It may not seem a big deal, but ensures a single point for otel traces formatting, and we have 7 implementations that can benefit from it 🙂

src/any_agent/callbacks/span_generation/llama_index.py
javiermtorres: Ok, I see, the `tool_call` is strictly implementation dependent and there are no any-agent specifics there  👍🏼

src/any_agent/callbacks/span_generation/google.py
javiermtorres: I'd propose `None` and `{}` as defaults in all implementations, to follow Python serialization

src/any_agent/callbacks/span_end.py
daavoo: I implemented isolated context per runs in the last changes

src/any_agent/callbacks/span_generation/google.py
daavoo: refactored

src/any_agent/callbacks/base.py
```diff
@@ -0,0 +1,27 @@
...
+        """Will be called before any Tool Execution starts."""
+        return context
+
+    def after_llm_call(self, context: Context, *args, **kwargs) -> Context:
+        """Will be called after any LLM Call is completed."""
+        return context
+
+    def after_tool_execution(self, context: Context, *args, **kwargs) -> Context:
+        """Will be called after any Tool Execution is completed."""
+        return context
```
njbrake, line 27: Will it be called even if a tool execution fails?

docs/agents/callbacks.md
```diff
@@ -0,0 +1,110 @@
+# Agent Callbacks
+
+For greater control when running your agent, `any-agent` includes support for custom [`Callbacks`][any_agent.callbacks.base.Callback] that
+will be called at different points of the [`AnyAgent.run`][any_agent.AnyAgent.run]:
+
+- [`before_llm_call`][any_agent.callbacks.base.Callback.before_llm_call]
```
javiermtorres: Please include a somewhat more detailed description (mermaid maybe????) _if it's not somewhere else_: "a turn in an agent includes an llm call, plus a sequence of zero or more tool calls, blah blah", so the flow of any-agent and where the callbacks will lie is explicit.

docs/agents/callbacks.md
```diff
@@ -0,0 +1,110 @@
...
+- [`after_llm_call`][any_agent.callbacks.base.Callback.after_llm_call]
+- [`before_tool_execution`][any_agent.callbacks.base.Callback.before_tool_execution]
+- [`after_tool_execution`][any_agent.callbacks.base.Callback.after_tool_execution]
+
+Advanced designs such as safety guardrails or custom side-effects can be integrated into your agentic system using this functionality.
+
+Each time an agent is run ( [`agent.run_async`][any_agent.AnyAgent.run_async] or [`agent.run`][any_agent.AnyAgent.run] ), a unique [`Context`][any_agent.callbacks.context.Context] object
+is shared across all callbacks.
```
javiermtorres, line 42: When would the context be discarded/recreated?

docs/agents/callbacks.md
```diff
@@ -0,0 +1,110 @@
...
+Callbacks can raise exceptions to stop agent execution. This is useful for implementing safety guardrails or validation logic:
+
+```python
+class SafetyGuard(Callback):
+    def before_tool_execution(self, context: Context, *args, **kwargs) -> Context:
+        tool_name = context.current_span.attributes.get("gen_ai.tool.name", "")
+
+        # Block dangerous tools
+        if tool_name in ["delete_file", "execute_code"]:
+            raise RuntimeError(f"Tool '{tool_name}' is not allowed for safety reasons")
```
javiermtorres, line 122: Would it be better to use a specific exception to signal that we're breaking the flow, instead of allowing any exception?

src/any_agent/callbacks/base.py
daavoo: Yes.

We wrap any tool with:

https://github.com/mozilla-ai/any-agent/blob/acdc35702e4474271c4fd90e9ca9e03fdb54fa7e/src/any_agent/tools/wrappers.py#L21-L27

And the current span generation (implemented as a callback) actually takes that output and sets the span status to Error if the string matches

docs/agents/callbacks.md
```diff
@@ -0,0 +1,110 @@
...
+            raise RuntimeError("Generated content contains inappropriate language")
+
+        return context
+```
+
+**Note**: Exceptions in callbacks will terminate the agent run immediately. Use this feature carefully to implement safety measures or validation logic.
+
+!!! warning
+
+    Callbacks will be called in the order that they are added, so it is important to pay attention to the order
```
javiermtorres: Clarify that the order is the same regardless of whether the calls are done before or after the event (tool or llm).

src/any_agent/callbacks/span_generation/langchain.py
```diff
@@ -0,0 +1,76 @@
...
+            {
+                "role": str(message.type).replace("human", "user"),
+                "content": str(message.content),
+            }
+            for message in messages[0]
+        ]
+
+        return self._set_llm_input(context, model_id, input_messages)
+
+    def after_llm_call(self, context: Context, *args, **kwargs) -> Context:
```
javiermtorres, line 33: Followup: further factor after_llm_call. There is still quite some boilerplate here.

src/any_agent/callbacks/wrappers/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+) -> (
+    _AgnoWrapper
+    | _GoogleADKWrapper
+    | _LangChainWrapper
+    | _LlamaIndexWrapper
+    | _OpenAIAgentsWrapper
+    | _SmolagentsWrapper
+    | _TinyAgentWrapper
+):
+    if framework is AgentFramework.AGNO:
```
javiermtorres, line 27: Minor nitpicking: maybe use a dict[AgentFramework, lambda]???

src/any_agent/callbacks/wrappers/agno.py
javiermtorres: As discussed with @daavoo , the order will be the same for after_X and before_X callbacks for consistency.

daavoo:
> Integration test (or add to existing integration test) to test out the callbacks

@njbrake the integration tests already test the callbacks because the span generation is implemented as a callback. I could maybe add a custom callback to the integration test or introduce a new test specifically for it

src/any_agent/config.py
daavoo: I guess a developer could follow the type link to the `Callback` and `Context` class definitions, but I am not sure what could replace this link (it is the same format we use for other stuff like `Tools` above

docs/agents/callbacks.md
daavoo: Added pseudocode, please check

src/any_agent/frameworks/any_agent.py
```diff
@@ -71,11 +72,15 @@ def __init__(self, config: AgentConfig):
...
+        self._wrapper = None
+        if self.config.callbacks is not None:
+            self._add_span_callbacks()
+            self._wrapper = _get_wrapper_by_framework(self.framework)
+
         self._tracer: Tracer = otel_trace.get_tracer(SCOPE_NAME)
 
         self._lock = asyncio.Lock()
-        self._running_traces: dict[int, AgentTrace] = {}
+        self._callback_context: dict[int, Context] = {}
```
javiermtorres: Nitpicking: callback_contexts or callback_context_map

src/any_agent/frameworks/any_agent.py
```diff
@@ -160,18 +165,11 @@ def run(self, prompt: str, **kwargs: Any) -> AgentTrace:
         """Run the agent with the given prompt."""
         return run_async_in_sync(self.run_async(prompt, **kwargs))
 
-    async def run_async(
-        self, prompt: str, instrument: bool = True, **kwargs: Any
-    ) -> AgentTrace:
+    async def run_async(self, prompt: str, **kwargs: Any) -> AgentTrace:
```
javiermtorres, line 166: Separate issue: AFAICT it is not clear from the user's pov if several runs can be active on the same agent at the same time (assuming async threads only). If so, let's add a small subsection on that, and check that there are indeed no unchecked shared data

src/any_agent/frameworks/any_agent.py
```diff
@@ -210,23 +213,28 @@ async def run_async(
...
             trace.add_span(invoke_span)
             raise AgentRunError(trace, e) from e
 
-        if instrumentation_enabled:
+        if self._wrapper:
             async with self._lock:
-                self._instrumentor.uninstrument(self)  # type: ignore[arg-type]
-                trace = self._running_traces.pop(trace_id)
+                if len(self._wrapper.callback_context) == 1:
+                    await self._wrapper.unwrap(self)  # type: ignore[arg-type]
```
javiermtorres, line 218: Extract a protocol/interface/abc and avoid the type union

src/any_agent/callbacks/wrappers/smolagents.py
```diff
@@ -0,0 +1,84 @@
...
+
+    from any_agent.callbacks.context import Context
+    from any_agent.frameworks.smolagents import SmolagentsAgent
+
+
+class _SmolagentsWrapper:
+    def __init__(self) -> None:
+        self.callback_context: dict[int, Context] = {}
+        self._original_llm_call: Callable[..., Any] | None = None
+        self._original_tools: Any | None = None
```
javiermtorres, line 20: Material for base classes :D

src/any_agent/frameworks/any_agent.py
```diff
@@ -210,23 +213,28 @@ async def run_async(
...
+                if wrapped_context := self._wrapper.callback_context.pop(
+                    trace_id, None
+                ):
+                    trace = wrapped_context.trace
```
javiermtorres, line 222: Can you please explain a bit this block? I'm not sure what it is doing. Thanks!!!!

docs/agents/callbacks.md
javiermtorres: Fine, thanks!!!!

docs/agents/callbacks.md
daavoo: I am not sure if commenting that clarifies or adds more confusion. I was hoping that the it is clear from the pseudocode, as `for callback in config.callbacks` implies that is the order

src/any_agent/callbacks/span_generation/langchain.py
daavoo: No sure what else can be refactored / what would be the benefit. Each code inside here is very framework dependent

src/any_agent/frameworks/any_agent.py
daavoo: > it is not clear from the user's pov if several runs can be active on the same agent at the same time (assuming async threads only). 

how would it be more clear?

> and check that there are indeed no unchecked shared data

Isn't this exactly what we test here?:

https://github.com/mozilla-ai/any-agent/blob/acdc35702e4474271c4fd90e9ca9e03fdb54fa7e/tests/integration/test_thread_safe.py#L29-L30

src/any_agent/frameworks/any_agent.py
daavoo: If there is a trace being generated by the wrapper, we pop and return that. If there is no wrapping (i.e. someone explicitly disabled wrapping by passing `callbacks=None`) then would just return a trace with a single span (the invoke agent)

daavoo: I will update the docs to explain that `callbacks=None` means disabling wrapping

src/any_agent/callbacks/wrappers/__init__.py
daavoo: I went with the pattern that is used elsewhere, don't see what would be the benefit/difference of using dict?

docs/agents/callbacks.md
daavoo: `Each time an agent is run`

src/any_agent/frameworks/any_agent.py
daavoo: Are you talking about the `arg-type` ignore?
Because that one is about `self` being the base `AnyAgent` class at this point

src/any_agent/callbacks/wrappers/smolagents.py
daavoo: I am not sure if it is worth it because different frameworks wrap different things. For some frameworks the tool instances are recreated, others have a `call_tool` method at the agent level that is wrapped, etc

src/any_agent/frameworks/any_agent.py
daavoo:
```
if len(self._wrapper.callback_context) == 1:
    await self._wrapper.unwrap(self)
```

Only unwraps a framework is there is not another execution of the same instance running in parallel (i.e. to account for `async.gather` case)

---

```
if wrapped_context := self._wrapper.callback_context.pop(
    trace_id, None
):
    trace = wrapped_context.trace
```

Gets the current running context and uses the trace that has being built by the span callbacks

daavoo:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/16076004146

docs/agents/callbacks.md
javiermtorres: Yes, that's true. It's ok.

src/any_agent/callbacks/span_generation/langchain.py
javiermtorres: We can discuss possible options in a follow up PR 👍

src/any_agent/callbacks/wrappers/__init__.py
javiermtorres: It's usually more compact and it can be modified at runtime. But it's not too relevant right now, hence the "minor nitpicking" :D

src/any_agent/frameworks/any_agent.py
javiermtorres: > how would it be more clear?

By stating it in the docs themselves, but I need to check if it's stated that "an agent instance can be reused across async threads, etc etc".

> Isn't this exactly what we test here?:

Yes and no. Since the execution order of statements in threads is not usually well defined, one usually needs to verify (hopefully formally) that undesirable situations cannot happen. Such a test _may_ pass, but cannot guarantee that the design is correct and failures won't happen.

In this case, I assume that the rest of possibly shared structures are guarded/immutable/safe, and the context dict is checked, so this is just a comment in case you can think of something else that could have been shared.

docs/agents/callbacks.md
javiermtorres: Ok, so if an agent is run again on the same thread, the context is recreated.

src/any_agent/frameworks/any_agent.py
javiermtorres: Sorry, the comment is quite unclear as it is stated. It's about the `self._wrapper` property in `AnyAgent`. It is not typed IIRC, plus when getting the wrapper the result is typed as a union:

```
def _get_wrapper_by_framework(
    framework: AgentFramework,
) -> (
    _AgnoWrapper
    | _GoogleADKWrapper
    | _LangChainWrapper
    | _LlamaIndexWrapper
    | _OpenAIAgentsWrapper
    | _SmolagentsWrapper
    | _TinyAgentWrapper
```

It would seem that all of them share similar wrap/unwrap operations, so defining a protocol/interface would be appropriate.

src/any_agent/callbacks/wrappers/smolagents.py
javiermtorres: Sorry, I tend to be very picky on boilerplate. It's just a suggestion anyway.

src/any_agent/frameworks/any_agent.py
```diff
@@ -183,24 +181,29 @@ async def run_async(
...
-                            agent=self,  # type: ignore[arg-type]
+                        trace_id = invoke_span.get_span_context().trace_id
+                        self._wrapper.callback_context[trace_id] = Context(
+                            current_span=invoke_span,
+                            trace=AgentTrace(),
+                            tracer=self._tracer,
+                            shared={},
                         )
 
+                        if len(self._wrapper.callback_context) == 1:
```
javiermtorres: EDIT: Nevermind. You're adding the context just before this, sorry.

Could it be??? Just in case, I'm definitely not sure. Is there an initial context entry to consider?

```suggestion
                        if len(self._wrapper.callback_context) == 0:
```

src/any_agent/frameworks/any_agent.py
javiermtorres: Thanks!!!!

PR #556, opened 2025-07-01
feat: more stable structured outputs in llama_index

njbrake:
This pull request introduces enhancements to the `llama_index` framework by integrating the `litellm` library to enforce structured outputs when specified, ensuring better compatibility with user-defined schemas.

### DIFF
--- a/src/any_agent/frameworks/llama_index.py
+++ b/src/any_agent/frameworks/llama_index.py
@@ -1,5 +1,7 @@
 from typing import TYPE_CHECKING, Any, cast
 
+from litellm import acompletion
+from litellm.utils import supports_response_schema
 from pydantic import BaseModel
 
 from any_agent import AgentConfig, AgentFramework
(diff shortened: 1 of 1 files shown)


njbrake:
Integration test: https://github.com/mozilla-ai/any-agent/actions/runs/16000351493/job/45133438844 

This lines up similarly with how we do structured output in tinyagent

njbrake:
I'll note that this change is motivated because when swapping over to Mistral I was finding that the Mistral model wasn't working with the existing llama_index methodology and required this extra step. So the redesign I put in here is to make it robust to leverage the litellm 'response_format' setting.

src/any_agent/frameworks/llama_index.py
```diff
@@ -97,7 +99,25 @@ async def _run_async(self, prompt: str, **kwargs: Any) -> str | BaseModel:
             msg = f"Agent did not return a valid response: {result.response}"
             raise ValueError(msg)
         if self.config.output_type:
+            # If the user needs a structured output, send it through litellm one last time to enforce the structured output.
```
daavoo, line 102: I am not sure if I am missing something @njbrake . 
If the model was already responding with the correct structured output, we would be unnecessarily making an extra call at this point, no?

Shouldn't we just do this extra call as a fallback?

njbrake: Yes you make a good point. The extra LLM call isn't always necessary. I'll update this.

PR #557, opened 2025-07-01
Add documentation showing how to use agents as tools

### LINKED ISSUE #382: Add `AnyAgent.as_tool()`

javiermtorres:
This small snippet shows how to reuse an agent as a tool for another agent, in an async setting.

@njbrake I am thinking about expanding this to include agents with structured outputs, which should look closer to tools, wdyt?

Closes #382


javiermtorres:
Didn't I already? I definitely agree 😅

docs/agents/tools.md
```diff
@@ -26,6 +26,62 @@ main_agent = AgentConfig(
...
+        AgentConfig(
+            name="google_expert",
+            model_id="gpt-4.1-mini",
+            description="An agent that can answer questions about the Google Agents Development Kit (ADK).",
+            tools=[search_web]
+        )
+    )
+
+    # Note that an agent with structured output
+    async def google_agent_as_tool(query: str) -> AgentTrace:
```
njbrake: Typing is wrong

javiermtorres: True!!!!

docs/agents/tools.md
```diff
@@ -26,6 +26,62 @@ main_agent = AgentConfig(
 )
 ```
 
+### Agents-as-tools
+
+As any callable can be used as a tool, you can wrap calling an agent inside a function and pass it as a tool, as in the following example:
+
+```python
+from uuid import uuid4
```
daavoo:
```suggestion
```

docs/agents/tools.md
```diff
@@ -26,6 +26,62 @@ main_agent = AgentConfig(
...
+        agent_framework=AgentFramework.OPENAI,
+        agent_config=main_agent_cfg,
+    )
+
+    agent_trace = await main_agent.run_async(prompt)
+
+    print(agent_trace.final_output)
+
+if __name__ == "__main__":
+    asyncio.run(main())
```
daavoo: I would try to simplify the code, it doesn't feel necessary to have an e2e example, just: agent tool definition -> wrapping as tool -> passing it to another agent config

```suggestion
```

docs/agents/tools.md
```diff
@@ -26,6 +26,62 @@ main_agent = AgentConfig(
...
+    # Note that an agent with structured output
+    async def google_agent_as_tool(query: str) -> str:
+        out = await google_agent.run_async(prompt=query)
+        return str(out.final_output)
+
+    google_agent_as_tool.__doc__ = google_agent.config.description
+
+    main_agent_cfg = AgentConfig(
+        instructions="Use the available tools to obtain additional information to answer the query.",
+        description="The orchestrator that can use other agents via tools.",
```
daavoo:
```suggestion
```

docs/agents/tools.md
```diff
@@ -26,6 +26,62 @@ main_agent = AgentConfig(
...
+
+```python
+from uuid import uuid4
+import asyncio
+from any_agent import AgentConfig, AgentFramework, AnyAgent, AgentTrace
+from any_agent.tools import search_web
+
+async def main():
+    prompt = "What do you know about the Google ADK?"
+
```
daavoo:
```suggestion
```

docs/agents/tools.md
```diff
@@ -26,6 +26,62 @@ main_agent = AgentConfig(
...
+    print(agent_trace.final_output)
+
+if __name__ == "__main__":
+    asyncio.run(main())
+```
+
+Since the agent will use the function documentation to decide whether it is appropriate to call the tool, we have copied the agent description into the function `__doc__` field. A normal docstring would also work.
+
+Other options to create tools from remote agents include using the MCP (SSE transport) or A2A protocols, as detailed in the following sections.
```
daavoo, line 73: It doesn't need to be part of this PR but I believe it would make more sense to have a multi-agent page or something like that where we have a single use case (i.e. google / openai expert) implemented in tabs with the different approaches.

The agent definition would be the same, and we would just show the different tool definitions in tabs

javiermtorres: Ah, you mean 3 tabs side to side: agent-as-function, agent-as-a2a, agent-as-mcp?

If it would only mean a section at the end with these tabs, with the content being extracted from the existing examples, I'd be open to do it in this PR 👍🏼

javiermtorres:
@daavoo I'll commit your suggestions, and then provide feedback (most probably I'll just agree to the final looks)

docs/agents/tools.md
javiermtorres: I'll add some comment like "the you run the agent, blah blah", but otherwise ok for me

docs/agents/tools.md
```diff
@@ -26,6 +26,47 @@ main_agent = AgentConfig(
...
+
+    google_agent_as_tool.__doc__ = google_agent.config.description
+
+    main_agent_cfg = AgentConfig(
+        instructions="Use the available tools to obtain additional information to answer the query.",
+        tools=[
+            google_agent_as_tool
+        ],
+        model_id="gpt-4.1-nano",
+    )
```
njbrake: Please run this code to make sure it works, because since it's wrapped in a main() that is never called, it will be skipped by our docs tests.

Maybe there should be an asyncio.run(main()) to call the code?

javiermtorres: There was, previously, but David prefers to have a shorter example.

The code indeed works (but it's not pushed... maybe we could have an examples dir, or the like).

daavoo: > There was, previously, but David prefers to have a shorter example.

 It doesn't matter how short/long the example is and it is only about definitions and not executing the agent.
 
The thing is that it is defined inside `async def main`, so it never runs. If you put it outside and use sync API, it would run as part of the docs tests, otherwise you just need to add what Nathan suggested.

Issue #318, opened 2025-05-23
Add TaskStore type to ServingConfig


njbrake:
💅 this could probably be an option to be set in `serving_config` like we have `model_type` and `DEFAULT_MODEL_TYPE` in agents.

_Originally posted by @daavoo in https://github.com/mozilla-ai/any-agent/pull/308#discussion_r2104226790_

PR #560, opened 2025-07-01
Support for push notifications in non-streaming mode

njbrake:
Clients can now subscribe to tasks to be updated when their states change.


src/any_agent/serving/a2a/agent_executor.py
```diff
@@ -77,7 +77,8 @@ async def execute(  # type: ignore[misc]
             task = new_task(context.message)
             await event_queue.enqueue_event(task)
         else:
-            logger.debug("Task already exists: %s", task.model_dump_json(indent=2))
+            logger.info("Task already exists: %s", task.model_dump_json(indent=2))
+        updater = TaskUpdater(event_queue, task.id, task.contextId)
```
daavoo, line 81: Any reason for moving this?

src/any_agent/serving/a2a/server_a2a.py
```diff
@@ -52,6 +53,7 @@ async def _get_a2a_app_async(
     request_handler = DefaultRequestHandler(
         agent_executor=AnyAgentExecutor(agent, task_manager),
         task_store=InMemoryTaskStore(),
+        push_notifier=InMemoryPushNotifier(httpx_client=httpx.AsyncClient()),
```
daavoo: @njbrake it looks like this would soon be broken: https://github.com/a2aproject/a2a-python/pull/212

tests/integration/a2a/test_push_notification.py
```diff
@@ -0,0 +1,240 @@
...
+    try:
+        # Use the helper context manager for agent serving and client setup
+        async with a2a_client_from_agent(
+            agent, serving_config, http_timeout=DEFAULT_LONG_TIMEOUT
+        ) as (client, server_url):
+            # Generate IDs for the conversation
+            first_message_id = str(uuid4())
+
+            # Configure push notifications in the initial message/send request
+            # following the A2A specification example
```
daavoo, line 176: I am not really understanding how users would use this feature from the any-agent perspective. Would it be possible to use push notifications when using a2a_as_tool?

javiermtorres: Let's discuss this next week. I see uses, but a human user won't normally have a defined domain/ip to set up a callback. _Unless_ you have a notification service like a queue to which notifications are sent and the the user would poll them out.

I see it ok as an example of another way of notifying, but maybe we need a more compelling narrative of why one would use this and in which kind of setup.

The a2a_as_tool case _could_ be such a setup. But it's very specific since the original agent would need to set up the callback, plus be able to handle asynchronous notifications from other agents.

njbrake: Push notification support is purely to support people who wish to contact an A2A served any-agent agent using their own client, we have no use for it in `a2a_tool`.

Right now, it has low importance because with a2a streaming=False, the only updates sent out for push notification are the last taskupdate when on_message_send completes. (failed/completed/cancelled/input_required).

This feature is to enable further customization of notifications for complex tasks as well as push notifications to be sent when using the newly available 'callbacks' feature. We still may not have any use for it in the a2a-as-tool multi-agent setup, but by implementing this we are moving towards a more complete implementation of the A2A spec.

src/any_agent/serving/a2a/agent_executor.py
njbrake: Yes: when we support streaming mode and callbacks etc I'll want to pass things into the updater as a part of running the agent. This re-order is to support this planned usage.

src/any_agent/serving/a2a/server_a2a.py
njbrake: Good to know. I'll be ready to patch once that new release drops 👍

tests/integration/a2a/test_push_notification.py
javiermtorres: I'd explore the possibility of exploring the grpc transport if we feel that notifications and async tasks will be important for users later on. Push notifications, as I mentioned, are ok in a distributed setting, but probably not so easy for individual users (unless maybe when using local containers talking among themselves).

src/any_agent/serving/a2a/server_a2a.py
daavoo: similar to #318 , this should probably be exposed as part of the config

daavoo (approved):
LGTM, but not sure how useful it is to have this if it has been already broken by a new a2a version. 

Overall, I feel like we are investing a lot of time in A2A features while they are still half-backed and we didn't really got a signal of people needing / using those.

I know we don't want to release saying that we support A2A with less than half of the features, but feels like they are still trying to figure out stuff on their side and we just suffer the side effects of trying to keep up.

njbrake:
> LGTM, but not sure how useful it is to have this if it has been already broken by a new a2a version.

In this case, the breaking change in the new A2A version is an easy upgrade. But point taken that now our a2a version pin is restrictive because we'll need ==0.2.11 instead of 0.2.8<=n<=0.2.10.

> Overall, I feel like we are investing a lot of time in A2A features while they are still half-backed and we didn't really got a signal of people needing / using those.

Totally agree. I think it makes sense to put a pause on our new A2A feature development (which will align nicely with upcoming paternity leave 😆 ) to focus on supporting and stabilizing the features that are currently available, and can prioritize new features as we get feedback about their usefulness.

src/any_agent/serving/a2a/server_a2a.py
njbrake: Good reminder. I'll take care of the push notifier in this PR and I'll take care of 318 later today.

PR #561, opened 2025-07-01
Increase timeout for message sending in integration tests 🕒✨

njbrake:
Default httpx timeout is 5 seconds. Occasionally an agent takes longer than 5 seconds, so imo it's an unreasonable default for our use case.

This should help stabilize our integration tests too


javiermtorres:
If this is hurting integration, I'm ok with this.

Maybe we can log the request latency to tune this better, if we don't do it already.

javiermtorres (approved):

PR #563, opened 2025-07-02
fix(agent_trace): Avoid creating `BaseModel` when loading a dumped tr…

daavoo:
…ace.

- Drop unnecessary custom serializer.
- Update traces and snapshots.


daavoo:
Need to update the unit tests (these test would also break in `main` branch if assets are updated, see #564 )

src/any_agent/tracing/agent_trace.py
```diff
@@ -220,24 +220,12 @@ class AgentTrace(BaseModel):
...
-        """Serialize the final_output and handle any BaseModel subclass."""
-        if value is None:
-            return None
-        if isinstance(value, str):
-            return value
-        if isinstance(value, BaseModel):
-            # This will properly serialize any BaseModel subclass
-            return value.model_dump()
-        return value
-
```
njbrake, line 240: Huh. I'm surprised it's not necessary... When I implemented the final output feature I distinctly remember it not serializing unless I added this serializer. Do you have any idea why it works now? From looking at the new generated trace assets I agree it seems fine, just scratching my head about why.

src/any_agent/tracing/agent_trace.py
```diff
@@ -220,24 +220,12 @@ class AgentTrace(BaseModel):
     """A list of [`AgentSpan`][any_agent.tracing.agent_trace.AgentSpan] that form the trace.
     """
 
-    final_output: str | BaseModel | None = Field(default=None)
```
njbrake, line 223: I suppose this is inevitable but I guess it means you can't perfectly serialize and deserialize AgentTrace, since the pydantic model will get serialized as a dict and then when you de-serialize it, it will be loaded into final_output as a string, not as whatever type it was.

daavoo: we would need to use pickle if we want to preserve the actual object

src/any_agent/tracing/agent_trace.py
daavoo: > o you have any idea why it works now? From looking at the new generated trace assets I agree it seems fine, just scratching my head about why.

I don't really know why it was not working for you. Do you remember or have the error available?

PR #567, opened 2025-07-02
Avoid import errors in Python 3.11

### LINKED ISSUE #565: Import issues when using A2A and Python 3.11
Some of the classes for serving use the `typing.override` decorator, added in Python version 3.12. When using the serving features with Python version 3.11, this decorator should be avoided.

The integration tests don't currently run on Python 3.11 due to this issue.

javiermtorres:
The `typing.override` decorator is not available in python 3.11. A noop decorator is provided instead for that version.

Closes #565

### DIFF
--- a/src/any_agent/serving/a2a/agent_executor.py
+++ b/src/any_agent/serving/a2a/agent_executor.py
@@ -1,4 +1,20 @@
-from typing import TYPE_CHECKING, override
+import sys
+
+PYTHONEGT312 = sys.version_info >= (3, 12)
+
+from typing import TYPE_CHECKING
+
+if PYTHONEGT312:
+    from typing import override
+else:
+    from collections.abc import Callable
+    from typing import Any, TypeVar
+
+    F = TypeVar("F", bound=Callable[..., Any])
+
+    def override(func: F, /) -> F:  # noqa: D103
+        return func
+
 
 from a2a.server.agent_execution import AgentExecutor, RequestContext
 from a2a.server.events import EventQueue
--- a/src/any_agent/serving/a2a/agent_executor.py
+++ b/src/any_agent/serving/a2a/agent_executor.py
@@ -7,11 +7,22 @@ from typing import TYPE_CHECKING
 if PYTHONEGT312:
     from typing import override
 else:
+    # Fix for Python 3.11
+    # We will define a "noop" decorator that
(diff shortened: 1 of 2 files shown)


src/any_agent/serving/a2a/agent_executor.py
```diff
@@ -1,4 +1,20 @@
...
+if PYTHONEGT312:
+    from typing import override
+else:
+    from collections.abc import Callable
+    from typing import Any, TypeVar
+
+    F = TypeVar("F", bound=Callable[..., Any])
+
+    def override(func: F, /) -> F:  # noqa: D103
+        return func
```
njbrake: I have no idea how this works but I guess if the integration and unit tests work then all good. Is this the implementation in python 3.12 copied pasted here? Would be maybe nice to have a short comment explaining what's going on here.

javiermtorres: Basically, we define a decorator that doesn't decorate XD Trying to define different classes depending on the python version was much messier. The funny looking stuff is just needed for type checking: we accept any function signature, but we ensure that we return something with that same signature. 

I'll include this as a comment. I am running the integration locally, I'll probably use a separate PR to write the GH tasks for v3.12 and v3.11 integration tests.

javiermtorres: AFAICT in python 3.11 the "overriding" constraint is not checked, since this decorator is not defined (hence the noop implementation). This is just a syntactic bandage to avoid the import error.

javiermtorres:
Local integration results:

```
 ⇒ python --version
Python 3.11.9
⇒ pytest -q tests/integration/
[...]
Results (322.17s (0:05:22)):
      72 passed
       2 failed
         - tests/integration/a2a/test_a2a_tool_multiturn.py:182 test_a2a_tool_multiturn
         - tests/integration/a2a/test_a2a_tool_multiturn.py:307 test_a2a_tool_multiturn_async
       4 skipped

daavoo (commented):
Can we add some unit tests that could trigger this error? The unit tests run in a matrix of minor Python versions, including 3.11

javiermtorres:
I was thinking about it. It's weird that no unit tests trigger this even though the serving modules are used AFAICT. I'll open a new issue and work later on it.

PR #568, opened 2025-07-02
Upgrade to support new A2A Protocol

njbrake:
* Save agent convo history under the "context id" not task. Now that a2a has upgraded their spec I now understand that task is just for managing whether a agent completed the task, it's not supposed to manage conversation history. The new version of the a2a spec enforces that design, which is what motivates this change
* When calling an agent as a tool, now the agent has the option to specify either task or context ids.
* A2A tool now responds with a dict, not a string. This maps better with the way that outputs are logged and managed

### DIFF
--- a/src/any_agent/serving/a2a/config_a2a.py
+++ b/src/any_agent/serving/a2a/config_a2a.py
@@ -53,7 +53,7 @@ class A2AServingConfig(BaseModel):
                     description="Search the web for information"
                 )
             ],
-            task_timeout_minutes=15
+            context_timeout_minutes=15
         )
 
     """
@@ -80,7 +80,7 @@ class A2AServingConfig(BaseModel):
 
     version: str = "0.1.0"
 
-    task_timeout_minutes: int = 10
+    context_timeout_minutes: int = 10
     """Task timeout in minutes. Tasks will be cleaned up after this period of inactivity."""
 
     history_formatter: HistoryFormatter = default_history_formatter
--- a/src/any_agent/tools/a2a.py
+++ b/src/any_agent/tools/a2a.py
@@ -3,7 +3,7 @@
 import re
 from collections.abc import Callable, Coroutine
 from contextlib import suppress
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Optional
 from uuid import uuid4
 
 from any_agent.utils.asyncio_sync import run_async_in_sync
@@ -28,8 +28,8 @@ with suppress(ImportError):
 
 
 async def a2a_tool_async(
-    url: str, toolname: str | None = None, http_kwargs: dict[str, Any] | None = None
-) -> Callable[[str], Coroutine[Any, Any, dict[str, Any]]]:
+    url: str, toolname: Optional[str] = None, http_kwargs: dict[str, Any] | None = None
+) -> Callable[[str, Optional[str], Optional[str]], Coroutine[Any, Any, dict[str, Any]]]:
     """Perform a query using A2A to another agent.
 
     Args:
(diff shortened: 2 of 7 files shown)


src/any_agent/serving/a2a/config_a2a.py
```diff
@@ -80,7 +80,7 @@ class A2AServingConfig(BaseModel):
 
     version: str = "0.1.0"
 
-    task_timeout_minutes: int = 10
+    context_timeout_minutes: int = 10
     """Task timeout in minutes. Tasks will be cleaned up after this period of inactivity."""
```
javiermtorres: Could it be...

```suggestion
    """Context timeout in minutes. Contexts will be cleaned up after this period of inactivity."""
```

src/any_agent/tools/a2a.py
```diff
@@ -58,7 +58,13 @@ async def a2a_tool_async(
             A2ACardResolver(httpx_client=resolver_client, base_url=url)
         ).get_agent_card()
 
-    async def _send_query(query: str, task_id: str = str(uuid4())) -> str:
+    # NOTE: Use Optional[T] instead of T | None syntax throughout this module.
```
daavoo, line 61: Is there an issue in Google ADK for this? It feels like it should be supported

tests/integration/a2a/test_a2a_tool_multiturn.py
```diff
@@ -283,8 +290,8 @@ async def test_a2a_tool_multiturn() -> None:
                     "role": "user",
                     "parts": [{"kind": "text", "text": THIRD_TURN_PROMPT}],
                     "messageId": str(uuid4()),
-                    "contextId": response_1.root.result.contextId,  # Same context to continue conversation
-                    "taskId": response_1.root.result.id,
+                    "contextId": response_2.root.result.contextId,  # Same context to continue conversation
+                    "taskId": response_2.root.result.id,
```
daavoo, line 294: So why is the task_id sent in this final message payload?

src/any_agent/tools/a2a.py
njbrake: I can make a ticket for them. Added to https://github.com/google/adk-python/issues/1725#issuecomment-3028245834

tests/integration/a2a/test_a2a_tool_multiturn.py
njbrake: That's because from response_2, the taskstate is returned as `input_required` (which I assert up a little earlier), and so here I'm testing that I can both specificy the context (so that the server adds the previous messages into the prompt context) as well as the task (so that the A2A server properly loads and updates the existing task, which will come in handy when we get to push notification and streaming support.

daavoo (commented):
I trust you and whether the tests pass so no need to block this PR

I see what changed here and makes sense but I feel like I would have to read the A2A spec to understand context vs task iD differences to properly review this 😅

tests/integration/a2a/test_a2a_tool_multiturn.py
javiermtorres: Who decides when the task is finished, is it always the user?

njbrake: Good question! Currently, it is never the client, it is always the server. The server creates the task and the Agent running in the server picks the state. If the state is a terminal state (finished/cancelled/complete) then the state cannot be resumed.

njbrake:
> I trust you and whether the tests pass so no need to block this PR
> 
> I see what changed here and makes sense but I feel like I would have to read the A2A spec to understand context vs task iD differences to properly review this 😅

The A2A team recently released this https://a2aproject.github.io/A2A/latest/topics/life-of-a-task/#task-refinements-follow-ups which is extremely useful for understanding how the task vs context is designed to work. Our initial implementation didn't leverage the difference in how the task vs context is supposed to work.

javiermtorres (approved):

PR #570, opened 2025-07-03
Consolidate the model id into variables

njbrake:
This is to clean up so that my Mistral PR is a little smaller.

### DIFF
--- a/tests/integration/a2a/test_a2a_serve.py
+++ b/tests/integration/a2a/test_a2a_serve.py
@@ -8,7 +8,7 @@ from a2a.client import A2AClient
 # Import your agent and config
 from any_agent import AgentConfig, AnyAgent
 from any_agent.serving import A2AServingConfig
-from tests.integration.helpers import DEFAULT_MODEL_ID, wait_for_server_async
+from tests.integration.helpers import DEFAULT_SMALL_MODEL_ID, wait_for_server_async
 
 from .conftest import A2ATestHelpers, a2a_client_from_agent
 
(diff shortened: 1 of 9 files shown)


daavoo (approved):
Is there a reason this was not originally implemented as a fixture in conftest, so we don't rely on imports that can go wrong depending on the path?

njbrake:
> Is there a reason this was not originally implemented as a fixture in conftest, so we don't rely on imports that can go wrong depending on the path?

Hmm. If I do that then would I need to pass that variable around to the functions/things that need it though?

Thinking about
https://github.com/mozilla-ai/any-agent/blob/main/tests/integration/test_agent.py#L117
https://github.com/mozilla-ai/any-agent/blob/main/tests/integration/a2a/test_a2a_tool_multiturn.py#L153

We could refactor to make it work as a fixture, but my gut reaction is that it's simpler to use as a constant. What do you think @daavoo ?

daavoo:
> Hmm. If I do that then would I need to pass that variable around to the functions/things that need it though?
> 
> Thinking about https://github.com/mozilla-ai/any-agent/blob/main/tests/integration/test_agent.py#L117 https://github.com/mozilla-ai/any-agent/blob/main/tests/integration/a2a/test_a2a_tool_multiturn.py#L153
> 
> We could refactor to make it work as a fixture, but my gut reaction is that it's simpler to use as a constant. What do you think @daavoo ?

I guess we could make `any_agent.testing.helpers` or something that is available as an actual package to be imported instead of relying on relative imports. I would be fine with that

PR #571, opened 2025-07-03
fix: llamaindex should ask for usage stats explicitly

njbrake:
When switching to Mistral I found that by default, usage tokens are not included in the LiteLLM api except for OpenAI. This PR explicitly sets the request for usage to enable other providers.


daavoo (approved):
Approved assuming that:

- Integration tests pass
- it has been verified that other frameworks don't need it

njbrake:
>     * it has been verified that other frameworks don't need it

Llama_index is the only framework we use that leverages the streaming api for liteLLM under the hood, which is why they're a unique case.

njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/16054463933/job/45305231153

PR #573, opened 2025-07-03
fix: Migrate structure output handling into any-agent

njbrake:
When developing the mistral api, I found that mistral was incompatible with the langchain/litellm structured output configuration, because mistral requires that the last message passed in to the api be a user message, but langchain passes in a system message. Rather than try to patch up the langchain and langchain-litellm implementation, I decided it would be smoother to take this opportunity to migrate the structured output handling out of langchain and into our code. The logic I added mirrors langchains own logic, where it makes always makes a final LLM call to create the structured output at the end.

### DIFF
--- a/src/any_agent/frameworks/langchain.py
+++ b/src/any_agent/frameworks/langchain.py
@@ -105,8 +105,12 @@ class LangchainAgent(AnyAgent):
             completion_params: dict[str, Any] = {}
             if self.config.model_args:
                 # only include the temperature and frequency_penalty, not anything related to tools
-                completion_params["temperature"] = self.config.model_args.get("temperature")
-                completion_params["frequency_penalty"] = self.config.model_args.get("frequency_penalty")
+                completion_params["temperature"] = self.config.model_args.get(
+                    "temperature"
+                )
+                completion_params["frequency_penalty"] = self.config.model_args.get(
+                    "frequency_penalty"
+                )
 
             completion_params["model"] = self.config.model_id
             previous_messages = [


src/any_agent/frameworks/langchain.py
```diff
@@ -87,14 +87,40 @@ async def _run_async(self, prompt: str, **kwargs: Any) -> str | BaseModel:
...
-        return str(last_message.content)
+
+        # Post-process for structured output if needed
+        # This emulates the langgraph behavior for structured outputs,
+        # but since it happens outside of langgraph we can control the model call,
+        # because some providers like mistral don't allow for the model to be called without the most recent message being a user message.
+        if self.config.output_type:
+            # Add a follow-up message to request structured output
+            structured_output_message = {
+                "role": "user",
```
javiermtorres, line 102: Ok, so the issue would be in the role for this last message, right?

javiermtorres (approved):

src/any_agent/frameworks/tinyagent.py
```diff
@@ -229,7 +229,8 @@ async def _run_async(self, prompt: str, **kwargs: Any) -> str | BaseModel:
                     completion_params["messages"] = messages
                     if supports_response_schema(model=self.config.model_id):
                         completion_params["response_format"] = self.config.output_type
-                    completion_params["tool_choice"] = "none"
+                    if len(completion_params["tools"]) > 0:
```
daavoo, line 232: Looks like a fix for something that we don't catch in the current tests?

njbrake: Yes, I found it when I was debugging the langchain implementation. Do you want me to try and write a unit test for it?

daavoo: Yes, I think that would be useful

PR #574, opened 2025-07-03
Feat: Use Mistral API for docs and tests instead of OpenAI API

njbrake:
Ok. This PR is ready for review.

What you get:

* Now our docs and tests use and support an open source LLM and open source LLM provider

What you lose:

* The OpenAI models were faster and their hosted service more reliable. So to compensate for this, I had to bump up the timeout settings of pytest and the HTTP_KWARGS.


njbrake:
> high-level: what would be the benefit of using mistral over OpenAI?
> 

The main driver for switching to Mistral over GPT is so that our framework more directly supports the ecosystem of open source LLMs. In the case of Mistral, we get the benefit of using an open source LLM while also being able to use them as a provider which is theoretically as easy as using the GPT hosted solution.

However, after finishing the work to upgrade LiteLLM to work with Mistral the way we need it too, I get intermittent failures and I'm still figuring out whether its some remaining issues with the any-agent -> LiteLLM -> Mistral integration, or something else.

> One thing I was thinking is, no matter what provider we choose for the majority of the tests, it might be useful to run the "main" `test_agent` with several providers. Waybe we can set it up to run with tiny_agent + {mistral, openAI, gemini} separately from the rest of frameworks with a single provider

100% agree. Working on this PR was super useful to understand what needed to be done in order to enable Mistral to be used. I'd like to at least have a harness i can use to check some other providers I'd like us to support like ollama,llamafile,deepseek,etc (even if this is a test we only run manually locally as needed, not a CI/CD run test).

README.md
```diff
@@ -55,7 +55,7 @@ For this example we use a model hosted by openai, but you may need to set the re
 See [our Model docs](https://mozilla-ai.github.io/any-agent/frameworks/#models) for more information about using different models.
 
 ```bash
-export OPENAI_API_KEY="YOUR_KEY_HERE"  # or MISTRAL_API_KEY, etc
+export MISTRAL_API_KEY="YOUR_KEY_HERE"  # or MISTRAL_API_KEY, etc
```
javiermtorres: 😅
```suggestion
export MISTRAL_API_KEY="YOUR_KEY_HERE"  # or OPENAI_API_KEY, etc
```

javiermtorres: Could you give it a quick read, and check the consistency in the use of Mistral vs OpenAI? I guess we want to say that the user can choose whatever model they see fit (and then choose Mistral, OpenAI or whatever as we see fit).

pyproject.toml
```diff
@@ -11,7 +11,7 @@ dynamic = ["version"]
 dependencies = [
   "duckduckgo_search",
   "fire",
-  "litellm==1.72.4", # until https://github.com/BerriAI/litellm/issues/11759 is fixed
+  "litellm @ git+https://github.com/BerriAI/litellm@main",
```
javiermtorres: I'd use a specific tag/commit, just in case.

javiermtorres (commented):
If I understood correctly, the main changes are swapping the model to mistral, and using the right litellm version that connects correctly to the mistral service. LGTM in that case.

pyproject.toml
njbrake: This is a draft and I didn't yet ask for a review, please hold off on comments on the specifics of the code til I'm ready for them :)

pyproject.toml
```diff
@@ -147,7 +147,7 @@ disallow_untyped_decorators = false # mypy gets confused by pytest decorators
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "session"
-timeout = 30
+timeout = 120
```
njbrake, line 150: One of the tradeoffs of switching to Mistral API is that their service in my experience was less reliable, there were times where there was a larger initial lag, so I bumped up this value to be more accomodating.

njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/16126459701/job/45504424618

Cookbook Tests: https://github.com/mozilla-ai/any-agent/actions/runs/16126462978/job/45504757509

pyproject.toml
```diff
@@ -11,7 +11,7 @@ dynamic = ["version"]
 dependencies = [
   "duckduckgo_search",
   "fire",
-  "litellm==1.72.4", # until https://github.com/BerriAI/litellm/issues/11759 is fixed
+  "litellm>=1.74.0",
```
daavoo, line 14: I see the issue is still open, what changed?

daavoo (approved):

pyproject.toml
njbrake: We're stuck in a hard place. That logging is super annoying. But, the Mistral API only works on the newest version. I pinned litellm because the logging is spammy and annoying but it's not critical.

njbrake:
integration https://github.com/mozilla-ai/any-agent/actions/runs/16142103470/job/45551736813

cookbooks: https://github.com/mozilla-ai/any-agent/actions/runs/16142118065

PR #575, opened 2025-07-03
fix: response type message should not call tool

njbrake:
Found when working on mistral: some models may mistakenly try to call a tool. In order to get a response format that we can validate, in the tinyagent design they must return text, they cannot call a tool.

### DIFF
--- a/src/any_agent/frameworks/llama_index.py
+++ b/src/any_agent/frameworks/llama_index.py
@@ -44,13 +44,17 @@ class LlamaIndexAgent(AnyAgent):
     def _get_model(self, agent_config: AgentConfig) -> "LLM":
         """Get the model configuration for a llama_index agent."""
         model_type = agent_config.model_type or DEFAULT_MODEL_TYPE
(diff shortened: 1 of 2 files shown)


src/any_agent/frameworks/tinyagent.py
```diff
@@ -230,6 +230,7 @@ async def _run_async(self, prompt: str, **kwargs: Any) -> str | BaseModel:
                         self.completion_params["response_format"] = (
                             self.config.output_type
                         )
+                    self.completion_params["tool_choice"] = "none"
```
daavoo, line 233: is not a problem of this PR but I have just realized reviewing this. Isn't it a problem that we are directly overriding `self.completion_params` here (and the other parts)?

Like, if there are 2 subsequent runs of the same instance and the first one reaches this point, we would be messing up the params for the next run, no?

njbrake: Oh boy. Yes you're totally right. Great catch. I'll make a patch for that.

PR #577, opened 2025-07-03
fix: isolate completion_params

njbrake:
Each time run_async is called, it shouldn't inherit any changes made by other run_async calls.


src/any_agent/frameworks/tinyagent.py
```diff
@@ -180,10 +180,11 @@ async def _run_async(self, prompt: str, **kwargs: Any) -> str | BaseModel:
 
         num_of_turns = 0
         max_turns = kwargs.get("max_turns", DEFAULT_MAX_NUM_TURNS)
+        completion_params = self.completion_params.copy()
```
javiermtorres, line 183: I assume we want the `self.completion_params` to be defaults, right?

My assumption is that a deep copy is not necessary because all params will be immutable; just in case.

PR #582, opened 2025-07-04
docs: Reorder Agents-As-Tools 

daavoo:
Consolidate into a single section that share the same use case and can be linked.


daavoo:
Merging this because I feel like the current deployed docs are just way more confusing. Don't hesitate to comment/discuss

javiermtorres:
Yes, the new structure feels cleaner 👍

PR #583, opened 2025-07-04
Enable A2A in mypy checks

javiermtorres:
A2A was not installed when doing mypy type checking, and the import errors were being ignores. As a result, the A2A related packages were not inspected.

This PR aims at fixing type checking for the A2A-related functionality. Some checks will be resolved with ignores for the time being, since the Google A2A SDK is still a bit unstable.


pyproject.toml
```diff
@@ -144,6 +144,10 @@ follow_untyped_imports = true
 module = "tests.*"
 disallow_untyped_decorators = false # mypy gets confused by pytest decorators
 
+[[tool.mypy.overrides]]
+module = ["pytest", "mktestdocs", "pytest_lazy_fixtures", "syrupy.assertion"]
```
javiermtorres, line 148: We want to ignore only certain packages.

src/any_agent/serving/a2a/agent_executor.py
```diff
@@ -60,34 +60,41 @@ def __init__(self, agent: "AnyAgent", context_manager: ContextManager):
...
         context: RequestContext,
         event_queue: EventQueue,
     ) -> None:
         query = context.get_user_input()
         task = context.current_task
 
-        context_id = context.message.contextId
-        if not self.context_manager.get_context(context_id):
-            self.context_manager.add_context(context_id)
+        # FIXME context.message may be None
```
javiermtorres: Apparently the message in RequestContext can be None :-/ 
https://github.com/a2aproject/a2a-python/blob/1c8e12e448dc7469e508fccdac06818836f5b520/src/a2a/server/agent_execution/context.py#L96
I'd like to know if there's a good reason for that, and my current assumption is that it won't be in practice, so there are a few ignore[arg-type] around for that reason.

src/any_agent/tools/a2a.py
```diff
@@ -92,28 +95,46 @@ async def _send_query(
...
-                        " ".join(
-                            [
-                                part.root.text
-                                for part in response.root.result.status.message.parts
-                                if part.root.kind == "text"
-                            ]
-                        )
-                    },
-                }
+            elif isinstance(response.root, SendMessageSuccessResponse):
```
javiermtorres, line 104: The response can be a Message or a Task. @njbrake this built dict for Messages should make sense, but I guess you'll know more than me if a Message can reach us in practice instead of a Task.

Also, I'm replacing the checks about the kind with runtime `isinstance` type checks, which should be more reliable.

src/any_agent/frameworks/langchain.py
```diff
@@ -32,15 +32,15 @@
...
+    from langgraph.graph.state import CompiledStateGraph
 
 
 class LangchainAgent(AnyAgent):
     """LangChain agent implementation that handles both loading and running."""
 
     def __init__(self, config: AgentConfig):
         super().__init__(config)
-        self._agent: CompiledGraph | None = None
+        self._agent: CompiledStateGraph[Any] | None = None
```
javiermtorres, line 45: The spec for both (old and new class) must be pretty similar; since we only import this in type checks, tests won't go off.

javiermtorres:
pre-commit, unit and integration run ok locally. _*BUT*_. If I run like this:

```
pytest -v tests/unit/ tests/integration
```

I get this result:

```
Results (259.59s (0:04:19)):
     238 passed
       4 failed
         - tests/integration/test_agent.py:164 test_load_and_run_agent[LANGCHAIN]
         - tests/integration/test_agent.py:164 test_load_and_run_agent[OPENAI]
         - tests/integration/test_agent.py:164 test_load_and_run_agent[AGNO]
         - tests/integration/test_agent.py:164 test_load_and_run_agent[TINYAGENT]
      11 skipped
```

Which is weird :-/

pyproject.toml
njbrake: Is the pyproject.toml file used when running mypy in the pre-commit hook through the run-mypy.sh script?

src/any_agent/serving/a2a/agent_executor.py
njbrake: Please remove FIXME and either fix in this PR or create a github issue to track this fix.

src/any_agent/serving/a2a/agent_executor.py
```diff
@@ -60,34 +60,41 @@ def __init__(self, agent: "AnyAgent", context_manager: ContextManager):
...
         if not task:
-            task = new_task(context.message)
-            await event_queue.enqueue_event(task)
+            if context.message is not None:
+                task = new_task(context.message)
+                await event_queue.enqueue_event(task)
+            else:
+                msg = "Task does not exist but the message in context is None"
+                logger.warning(msg)
+                raise ValueError(msg)
```
njbrake, line 84: This is a behavior change. Why should this throw an exception?

src/any_agent/tools/a2a.py
njbrake: Our a2a server would never return a message but another a2a server might.

njbrake (changes requested):
Thanks for your work on this! Please remove the 'fixme' comments, and also provide more context about the behavior change that was added in the agent_executor. Please also make sure that you run the integration and cookbook tests and that they all pass. Thanks!

pyproject.toml
javiermtorres: At least when running `uv run pre-commit run --all-files --verbose`, the `pyproject.toml` is used. I'll check if the script does something different.

src/any_agent/serving/a2a/agent_executor.py
javiermtorres: Since you've been working on the protocol for some time, would you go for a type check ignore, or try to conform to the protocol to the letter?

src/any_agent/tools/a2a.py
javiermtorres: Ok, I'll go for the whole range of possible responses by default.

src/any_agent/serving/a2a/agent_executor.py
njbrake: I'm fine with going with a type check ignore for the moment. Once the Google a2a sdk hits version 1.0 we can come back to the A2A code and do a closer scrub of the mypy errors but I think it's ok to avoid overfitting to whatever the current state of the a2a-sdk is 👍

src/any_agent/serving/a2a/agent_executor.py
javiermtorres: Because the alternative where `context.message` is None is not considered in the original code, please check https://github.com/a2aproject/a2a-python/blob/1022093110100da27f040be4b35831bf8b1fe094/src/a2a/utils/task.py#L8
By default, I'll use exceptions for these cases unless I can come up with a suitable alternative, which I couldn't think about in this case :-/ Maybe we can just leave everything as it is if the Message and the Task are both None, but imho getting an exception thrown is the surest way to notice possible corner cases.

src/any_agent/serving/a2a/agent_executor.py
javiermtorres: Thanks, I was thinking more or less the same :)

pyproject.toml
njbrake: Ok cool! That's the same command I use so if it picks up the pyproject.toml with that command, works for me 👍

src/any_agent/serving/a2a/agent_executor.py
njbrake: Ah I see. In that case I'm good with keeping it as throwing an exception 👍

javiermtorres:
Integration running ok: https://github.com/mozilla-ai/any-agent/actions/runs/16167525663
Cookbooks running ok: https://github.com/mozilla-ai/any-agent/actions/runs/16167529089

src/any_agent/serving/a2a/config_a2a.py
```diff
@@ -112,7 +112,7 @@ class A2AServingConfig(BaseModel):
     push_notifier_sender_type: type[PushNotificationSender] = BasePushNotificationSender
     """Push notifier sender to be used by the agent.
 
-    If not provided, a default in-memory push notifier sender will be used.
+    If not provided, a default async httpx-based push notifier sender will be used.
```
javiermtorres, line 115: Note, by default the base class uses an async client to notify.

src/any_agent/serving/a2a/server_a2a.py
```diff
@@ -34,7 +34,7 @@ def _get_a2a_app(
     task_manager = ContextManager(serving_config)
     push_notification_config_store = serving_config.push_notifier_store_type()
     push_notification_sender = serving_config.push_notifier_sender_type(
-        httpx_client=httpx.AsyncClient(),
+        httpx_client=httpx.AsyncClient(),  # type: ignore[call-arg]
```
javiermtorres, line 37: To be used as basis for a new issue. We ignore the args here for a constructor, because apparently at compile time we don't know what subclass we'll be using, and the base class doesn't define an `__init__(self, **kwargs)` constructor. I'd force the user to provide a `Callable[[], PushNotificationSender]` (or the like) instead.

tests/integration/a2a/test_push_notification.py
```diff
@@ -176,11 +179,13 @@ async def webhook_handler(request: Request) -> JSONResponse:
             # following the A2A specification example
             params = MessageSendParams(
                 message=Message(
-                    role="user",
+                    role=Role.user,
                     parts=[
                         Part(
-                            kind="text",
-                            text=FIRST_TURN_PROMPT,
+                            root=TextPart(
```
javiermtorres, line 185: Apparently, likely to change in main :-/

tests/integration/a2a/test_push_notification.py
```diff
@@ -196,7 +201,10 @@ async def webhook_handler(request: Request) -> JSONResponse:
             if hasattr(response_1.root, "error"):
                 msg = f"Error: {response_1.root.error.message}, Code: {response_1.root.error.code}, Data: {response_1.root.error.data}"
                 raise RuntimeError(msg)
-            task_id = response_1.root.result.id
+            if isinstance(response_1.root.result, Task):
```
javiermtorres, line 204: Alternatives are considered here, since messages could also be sent. But a sanity check would be ok.

Issue #378, opened 2025-05-29
Add option to `TracingConfig` to limit the maximum `gen_ai.output` length in the console


daavoo:
Set a reasonable default.
Trim the characters exceeding the value.

njbrake:
sounds good so long as we make sure this implementation only effects the console output, and not the AgentTrace content

Issue #288, opened 2025-05-20
Compute token and $ use as a running total instead of a computed property


njbrake:
> I am 50/50 on the method to property but it is mostly a matter of style preference*, I am fine with merging this. The rest of the changes (assets and splitting into usage and cost) look good to me.
> 
> 
> 
> ---
> 
> 
> 
> *I prefer to hint the user that something involves computation by exposing it as a method. I feel tricked when I access a property that takes time / might break computing the value.

I think we may be able to avoid recomputation if we make it a running total instead of only calculating it on-demand. I'll keep it as is for this PR, but let me make a follow up issue to refactor this so that it's not a computed property.  

_Originally posted by @njbrake in https://github.com/mozilla-ai/any-agent/pull/281#issuecomment-2895452684_

PR #586, opened 2025-07-07
feat(callbacks): Introduce default callbacks and convert ConsoleExpor…

daavoo:
…ter to a callback.

- Expose `get_default_callbacks` and use it as default factory in the config.
- Add docs on how to disable the default callbacks.


docs/tracing.md
```diff
@@ -44,15 +44,10 @@ Here is what the console output looks like:
 
 !!! tip
 
-    Showing traces in the console is enabled by default.
+    The spans are printed to the console by default.
```
njbrake:
```suggestion
    The spans are printed to the console by default, using the `callback` mechanism.
```

njbrake:
@daavoo  what do you envision as other default callbacks? It's unclear to me why there should be a default_callbacks call instead of only a console callback, it makes it a little unclear what you're removing/adding by putting it behind the default_callbacks function. I'd rather avoid trying to future proof the design in this area if possible.

I think the design of having the console output be a callback totally makes sense and it's well documented ✅

daavoo:
> @daavoo what do you envision as other default callbacks? 

At the very least, I was thinking to include:

- Something like CostInfo (adding cost to spans, related to https://github.com/mozilla-ai/any-agent/issues/288)
- https://github.com/mozilla-ai/any-agent/issues/378

> It's unclear to me why there should be a default_callbacks call instead of only a console callback, it makes it a little unclear what you're removing/adding by putting it behind the default_callbacks function. I'd rather avoid trying to future proof the design in this area if possible.

Other options to expose this to the user are:

- Add `include_default_callbacks` argument or something like that to the config
- Just document what are the default callbacks and show an example on how you would import them to use alongside your custom callbacks.

I don't have a strong opinion on any.
I would say that I probably went with a callable because it was the simplest way of implement defaults in a way that is simple for us and reusable by users (just passing the same callable you expose to Pydantic `default_factory`)

docs/agents/callbacks.md
```diff
@@ -38,11 +40,10 @@ while True:
 
 Advanced designs such as safety guardrails or custom side-effects can be integrated into your agentic system using this functionality.
 
-Each time an agent is run ( [`agent.run_async`][any_agent.AnyAgent.run_async] or [`agent.run`][any_agent.AnyAgent.run] ), a unique [`Context`][any_agent.callbacks.context.Context] object
-is shared across all callbacks.
+During the agent run ( [`agent.run_async`][any_agent.AnyAgent.run_async] or [`agent.run`][any_agent.AnyAgent.run] ), a unique [`Context`][any_agent.callbacks.context.Context] object is created shared across all callbacks.
```
javiermtorres:
```suggestion
During the agent run ( [`agent.run_async`][any_agent.AnyAgent.run_async] or [`agent.run`][any_agent.AnyAgent.run] ), a unique [`Context`][any_agent.callbacks.context.Context] object is created and shared across all callbacks.
```

docs/agents/callbacks.md
```diff
@@ -69,9 +70,13 @@ class LimitSearchWeb(Callback):
             raise RuntimeError("Reached limit of `search_web` calls.")
 ```
 
-## Providing Callbacks
+## Default Callbacks
+
+`any-agent` comes with a set of default callbacks that will be used by default (if you don't pass a value to `AgentConfig.callbacks`):
+
+- [`ConsolePrintSpan`][any_agent.callbacks.span_print.ConsolePrintSpan]
```
javiermtorres, line 77: So this means that these will only be applied if the user doesn't provide callbacks? If the user provides some callbacks, do they also need to add this callback manually if they want this functionality?

javiermtorres (approved):
LGTM. I'm not really sure about how to handle default callbacks, though, but I wouldn't be able to make a proposal :-/ Take into account that the console exporter probably needs to be located at the end of a list of callbacks to reflect the information from all previous callbacks, but then the user needs to be aware of this.

docs/agents/callbacks.md
daavoo: Yes, that is exactly what I tried to explain in the tabs of the section right below `Providing your own Callbacks`

njbrake:
I don't have a solution in mind, but the thing I don't love is that if you haven't read the docs closely, a developer will be confused when they find that adding a callback will remove all their default callbacks.

Since I expect people to use the console callback heavily when getting up and running, it will be confusing to a person when all of the sudden their logging disappears when they add a callback. People don't generally think of logging as a callback so it's not intuitive that the two would be connected.

I'm still trying to think about what to practically suggest to fix this.

PR #588, opened 2025-07-07
Delete docs/examples/evaluation_case.yaml

njbrake:
Leftover from old evaluation api


njbrake:
Docs test passes https://github.com/mozilla-ai/any-agent/actions/runs/16121808637

PR #590, opened 2025-07-07
feat: integration tests, consolidate model args and http_kwargs into single variables

njbrake:
Found while working on Mistral. We don't set our temp to 0 in all the places where we use an LLM in the integration test, which contributes to the non-deterministic behavior of our integration tests.

I also consolidated them into constants/helpers for easy control.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ smolagents = [
 ]
 
 a2a = [
-  "a2a-sdk>=0.2.8",
+  "a2a-sdk>=0.2.8, <=0.2.10",
   "httpx",
 ]
 


tests/integration/test_agent.py
```diff
@@ -115,7 +118,12 @@ def assert_tokens(agent_trace: AgentTrace) -> None:
 def assert_eval(agent_trace: AgentTrace) -> None:
     """Test evaluation using the new judge classes."""
     # Test 1: Check if agent called write_file tool using LlmJudge
-    llm_judge = LlmJudge(model_id=DEFAULT_SMALL_MODEL_ID)
+    llm_judge = LlmJudge(
+        model_id=DEFAULT_SMALL_MODEL_ID,
+        model_args={
+            "timeout": 30.0
+        },  # Because it's an llm not agent, the default_model_args are not used
```
javiermtorres, line 125: Then maybe create a `get_default_llm_args`?

javiermtorres:
LGTM. Just to clarify: "consolidate model args and http_kwargs into single variable" means that _each of those args_ are consolidated _into their own variable each_, not both in a single variable, right?

tests/integration/test_agent.py
njbrake: Since we only use the LLMJudge in a single place (here) I opted to avoid adding an additional variable. However, if we expand the usage of LLMJudge in our integration tests, I agree consolidation into a new fn will make sense at that point 👍

njbrake:
> LGTM. Just to clarify: "consolidate model args and http_kwargs into single variable" means that _each of those args_ are consolidated _into their own variable each_, not both in a single variable, right?

Correct!

njbrake:
Integration tests pass: https://github.com/mozilla-ai/any-agent/actions/runs/16123539451/job/45494968321

PR #591, opened 2025-07-07
feat: restrict taskstates to those that are used.

njbrake:
Found while working on Mistral:

I noticed that sometimes a model served over A2A would respond with setting TaskStates to one of the task states that we don't support in non-streaming mode, like TaskState.working or TaskState.submitted, which are meaningless in a non-streaming scenario. To this end, I propose restricting to either 'completed ✅ ' 'input_required❓ ' or 'failed ❌ ' since these are the only ones we'd leverage with our implementation.

### DIFF
--- a/src/any_agent/serving/a2a/envelope.py
+++ b/src/any_agent/serving/a2a/envelope.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, Any, Generic, TypeVar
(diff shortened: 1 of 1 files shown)


javiermtorres:
Who is setting this state, the LLM or the A2A SDK?

njbrake:
> Who is setting this state, the LLM or the A2A SDK?

The LLM! 🚀

javiermtorres (approved):

njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/16123625747/job/45495242979

PR #592, opened 2025-07-07
feat: pin a2a sdk version

njbrake:
As discussed via slack: The Google A2A Python SDK and API is still in development and has put breaking changes into patch versions a few times. I'd like to switch over to pin the a2a version and use dependabot to carefully manged the version range until the a2a sdk gets more stable.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ smolagents = [
 ]
 
 a2a = [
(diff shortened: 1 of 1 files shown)


njbrake:
Integration test https://github.com/mozilla-ai/any-agent/actions/runs/16122617314

javiermtorres (approved):

PR #593, opened 2025-07-07
Brake/remove ambiguity

njbrake:
Found while working on Mistral: the multiturn integration test was specifying a generic output_type test with three booleans which l found was giving the model too much leeway to set true/false based on how it defined the task to be 'success'. For instance, the second prompt is designed to return the state `input_required` which then you might consider a success or failure. 

This PR introduces a more robust mechanism for ensuring that the prompts were called.

### DIFF
--- a/tests/integration/a2a/test_a2a_tool_multiturn.py
+++ b/tests/integration/a2a/test_a2a_tool_multiturn.py
@@ -346,16 +346,10 @@ async def test_a2a_tool_multiturn_async() -> None:
     await wait_for_server_async(server_url)
     try:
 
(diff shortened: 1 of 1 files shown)


daavoo (approved):
NGL, I look at the test and wonder what could be simplified

njbrake:
> NGL, I look at the test and wonder what could be simplified

I know what you mean. The two tests are complicated, but imo it's because the behavior we're testing is complicated and I'm trying to do it without using an LLM, which means lots of mocks. 

The test could be simplified by using the new `callback` feature, which would then let us check stuff without needing to add and force a tool call. I can give that a try.

PR #594, opened 2025-07-08
feat: add task store configuration to A2A server

njbrake:
annnnd immediately after merging the Push notification PR I realized that I had forgotten to update the sync version of the `_get_a2a_app` function 🤦 

* Introduced `task_store_type` in `A2AServingConfig` to allow custom task store implementations.
* Updated `_get_a2a_app` and `_get_a2a_app_async` functions to utilize the specified task store type, defaulting to `InMemoryTaskStore` if not provided.


src/any_agent/serving/a2a/config_a2a.py
```diff
@@ -96,3 +97,9 @@ class A2AServingConfig(BaseModel):
 
     If not provided, a default in-memory push notifier will be used.
     """
+
+    task_store_type: type[TaskStore] | None = None
```
javiermtorres: Wouldn't it be better to provide the default here and avoid having it optional?

javiermtorres (approved):
LGTM, but I'd rather move the default to the A2A serving config. I'm also wondering whether the serving is the right place to put tasks, etc, since a local agent could have a similar python interface without being served, but it's out of this scope (but I'd like to hear some comment on it).

src/any_agent/serving/a2a/config_a2a.py
njbrake: Hmm. Good question. Originally I put it here to be consistent with how we did it with DEFAULT_MODEL_TYPES in the main part of any-agent, but in that case we had to avoid importing the framework specific classes in the config. We don't have that same restriction here. I'll update.

njbrake:
> LGTM, but I'd rather move the default to the A2A serving config. I'm also wondering whether the serving is the right place to put tasks, etc, since a local agent could have a similar python interface without being served, but it's out of this scope (but I'd like to hear some comment on it).

if running outside of a2a, we would recommend the use of tools to manage a multiturn conversation instead of a task 👍  https://mozilla-ai.github.io/any-agent/agents/#multi-turn-conversations

njbrake:
Integration tests https://github.com/mozilla-ai/any-agent/actions/runs/16142438763

PR #596, opened 2025-07-08
fix(a2a): Bypass deepcopy of callbacks.

daavoo:
I am not sure why, trying to deepcopy the callback classes was causing some weird failures. This is the simples way I found to bypass.

TBH, deep copying the config feels like something worth revisiting.


njbrake:
I'm good with this to keep things moving. In general, I don't like the design of A2A needing to recreate the agent, this may be a reason to revisit our design choice and maybe instead pursue the design where we force the output type to inherit the A2AEnvelope when the agent is created, which would let us bypass this need to re-create the agent.

daavoo:
https://github.com/mozilla-ai/any-agent/actions/runs/16141915439/job/45551143793

PR #597, opened 2025-07-08
feat(callbacks): Implement `AddCostInfo` and include it in the defaul…

daavoo:
…t list.

### DIFF
--- a/src/any_agent/callbacks/span_cost.py
+++ b/src/any_agent/callbacks/span_cost.py
@@ -1,4 +1,4 @@
-# mypy: disable-error-code="no-untyped-def"
+# mypy: disable-error-code="attr-defined,index,no-untyped-def"
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
--- a/src/any_agent/callbacks/span_cost.py
+++ b/src/any_agent/callbacks/span_cost.py
@@ -30,8 +30,12 @@ def add_cost_info(span: Span) -> None:
                 prompt_tokens=int(attributes.get("gen_ai.usage.input_tokens", 0)),  # type: ignore[arg-type]
                 completion_tokens=int(attributes.get("gen_ai.usage.output_tokens", 0)),  # type: ignore[arg-type]
             )
-            attributes["gen_ai.usage.input_cost"] = cost_prompt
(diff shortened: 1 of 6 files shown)


src/any_agent/callbacks/span_cost.py
```diff
@@ -0,0 +1,51 @@
...
+    from any_agent.callbacks.context import Context
+    from any_agent.tracing.otel_types import AttributeValue
+
+
+def add_cost_info(span: Span) -> None:
+    """Use litellm to compute cost and add it to span attributes."""
+    attributes: Mapping[str, AttributeValue] = span.attributes
+    if any(
+        key in attributes
+        for key in ["gen_ai.usage.input_tokens", "gen_ai.usage.output_tokens"]
```
njbrake, line 25: Is it worth elevating `gen_ai.usage.input_tokens` etc to be constant values, or do we expect this never to change?

daavoo:
> Can you add a unit test or two to verify the behavior?

Yes sir, it was WIP 😅 fixing the Integration tests that indirectly test this (and are currently failing)

src/any_agent/callbacks/span_cost.py
daavoo: > Is it worth elevating `gen_ai.usage.input_tokens` etc to be constant values, or do we expect this never to change?

I am always thinking whether it is worth it or not. I think I am going to elevate all the values in a separate PR because now, with the callbacks stuff, we might want to expose those values to users so they can easily access span information

daavoo:
https://github.com/mozilla-ai/any-agent/actions/runs/16147986768

tests/unit/callbacks/test_span_cost.py
```diff
@@ -0,0 +1,81 @@
...
+
+from any_agent.callbacks.span_cost import AddCostInfo
+
+
+def test_span_cost() -> None:
+    context = MagicMock()
+    current_span = MagicMock()
+
+    current_span.attributes = {
+        "gen_ai.request.model": "gpt-4.1-mini",
```
njbrake, line 11: doesn't matter but I tried to generally wipe out references to gpt in favor of mistral. Since it's a unit test I realize it doesn't matter though, nbd.

Issue #452, opened 2025-06-10
Set up sockets to provide to uvicorn


javiermtorres:
Thanks for your work on this. I'll offer one other design option for you to consider: what if you bound to a socket in the any_agent serve function and then passed that socket in to the underlying serve functions? That way the AnyAgent.serve_async handles all the logic around picking the right port, and the underlying functions aren't responsible for creating and parsing the port.

This would also make it so that when we add support for MCP/ACP etc, you won't need to re-implement the port selection and parsing logic (even though it's only a few lines of code so no big deal).

Feel free to take this or leave it, since what you have here I think is also good.

<img width="867" alt="Screenshot 2025-06-09 at 12 02 19 PM" src="https://github.com/user-attachments/assets/97cce12d-96a0-4d63-968a-f9e4912d5419" />

_Originally posted by @njbrake in https://github.com/mozilla-ai/any-agent/pull/442#pullrequestreview-2910541566_

PR #598, opened 2025-07-08
Add client support for streamable http MCP servers

javiermtorres:
Enables any-agents to talk to remote MCP servers using the Streamable HTTP protocol.

### DIFF
--- a/.github/workflows/tests-cookbook.yaml
+++ b/.github/workflows/tests-cookbook.yaml
@@ -33,3 +33,8 @@ jobs:
           TAVILY_API_KEY: ${{ secrets.TAVILY_API_KEY }}
         run: |
           pytest tests/cookbooks -v -n 4
+
+      - name: Upload finished cookbooks
+        uses: actions/upload-artifact@v4
+        with:
+          path: "docs/cookbook/*.ipynb"


src/any_agent/tools/mcp/frameworks/agno.py
```diff
@@ -73,13 +75,47 @@ async def list_tools(self) -> list["AgnoMCPTools"]:
...
+            include_tools=list(self.mcp_tool.tools) if self.mcp_tool.tools else None,
+        )
+        return await super().list_tools()
+
+
+class AgnoMCPStreamableHttpConnection(AgnoMCPConnection):
+    mcp_tool: MCPStreamableHttp
+    _get_session_id_callback: Callable[[], str | None] | None = None
+
+    def get_session_id(self) -> str | None:
```
javiermtorres, line 98: The streamable http mcp implementation returns a function to retrieve the session id. I don't see that we'd be needing the session right now, but we may need to link e.g. a context to the specific mcp session this agent is using in this context.

tests/conftest.py
```diff
@@ -86,6 +114,36 @@ async def echo_sse_server() -> AsyncGenerator[dict[str, str], None]:
...
+@pytest.fixture(scope="session")
+async def date_streamable_http_server(worker_id: str) -> AsyncGenerator[dict[str, Any]]:
+    """This fixture runs a FastMCP server in a subprocess.
+    I thought about trying to mock all the individual mcp client calls,
+    but I went with this because this way we don't need to actually mock anything.
+    This is similar to what MCPAdapt does in their testing https://github.com/grll/mcpadapt/blob/main/tests/test_core.py
+    """
+    import asyncio
+
+    port = 19010
```
javiermtorres, line 127: Again, issues with server ports, this time due to xdist setting up several processes. Possible alternative approaches would be:

1. Moving to in-place async servers instead of subprocess, and bind to random ports.
1. Grouping tests to ensure some fixtures are only spawned in a single worker, where needed.

tests/integration/test_agent.py
```diff
@@ -249,3 +302,71 @@ def write_file(text: str) -> None:
             f.write(html_output.replace("<!DOCTYPE html>", ""))
 
     assert_eval(agent_trace)
+
+
+def test_load_and_run_agent_streamable_http(
```
javiermtorres: This is a somewhat lighter version of the existing test. I opted to go for integration tests to ensure the streamable http protocol worked ok.

javiermtorres:
Cookbook tests seem to be broken on main: https://github.com/mozilla-ai/any-agent/actions/runs/16172156043/job/45648178322
Otherwise integration tests run ok: https://github.com/mozilla-ai/any-agent/actions/runs/16171648669

javiermtorres:
Cookbook now running ok: https://github.com/mozilla-ai/any-agent/actions/runs/16172504758
Thanks @daavoo :)

src/any_agent/tools/mcp/frameworks/__init__.py
```diff
@@ -87,11 +93,47 @@ def _get_sse_mcp_server(mcp_tool: MCPSse, agent_framework: AgentFramework) -> MC
...
+        from .tinyagent import TinyAgentMCPServerStreamableHttp
+
+        return TinyAgentMCPServerStreamableHttp(mcp_tool=mcp_tool)
+    assert_never(agent_framework)
+
+
 def _get_mcp_server(mcp_tool: MCPParams, agent_framework: AgentFramework) -> MCPServer:
     if isinstance(mcp_tool, MCPStdio):
         return _get_stdio_mcp_server(mcp_tool, agent_framework)
     if isinstance(mcp_tool, MCPSse):
```
njbrake, line 133: Should we add some kind of deprecation warning for those that try to use MCPSse?

tests/conftest.py
```diff
@@ -86,6 +114,36 @@ async def echo_sse_server() -> AsyncGenerator[dict[str, str], None]:
...
+    if worker_id and "gw" in worker_id:
+        port += int(worker_id.strip("gw"))
```
njbrake, line 129: Can you explain what this is? Its not clear to me why this is different from the SSE server fixture that already exists

tests/integration/test_agent.py
```diff
@@ -249,3 +302,71 @@ def write_file(text: str) -> None:
...
+    env_check = validate_environment(kwargs["model_id"])
+    if not env_check["keys_in_environment"]:
+        pytest.skip(f"{env_check['missing_keys']} needed for {agent_framework}")
+
+    model_args: dict[str, Any] = (
+        {"parallel_tool_calls": False}
+        if agent_framework not in [AgentFramework.AGNO, AgentFramework.LLAMA_INDEX]
+        else {}
+    )
+    model_args["temperature"] = 0.0
```
njbrake: use the default model args from any_agent.testing.helpers

tests/integration/test_agent.py
```diff
@@ -249,3 +302,71 @@ def write_file(text: str) -> None:
...
+        output_type=Steps,
+        **kwargs,  # type: ignore[arg-type]
+    )
+    agent = AnyAgent.create(agent_framework, agent_config)
+    update_trace = request.config.getoption("--update-trace-assets")
+    if update_trace:
+        for callback in agent.config.callbacks:
+            if isinstance(callback, ConsolePrintSpan):
+                console = callback.console
+                callback.console.record = True
```
njbrake: Should this be here?

tests/integration/test_agent.py
```diff
@@ -249,3 +302,71 @@ def write_file(text: str) -> None:
...
+    assert_tokens(agent_trace)
+
+    if update_trace:
+        trace_path = Path(__file__).parent.parent / "assets" / agent_framework.name
+        with open(f"{trace_path}_trace.json", "w", encoding="utf-8") as f:
+            f.write(agent_trace.model_dump_json(indent=2))
+            f.write("\n")
+        html_output = console.export_html(inline_styles=True)
+        with open(f"{trace_path}_trace.html", "w", encoding="utf-8") as f:
+            f.write(html_output.replace("<!DOCTYPE html>", ""))
```
njbrake: I'm not sure this needs to be here

njbrake (changes requested):
Nice! A few changes for the tests, but nothing too major. Overall, if the integration tests are all working, I'll trust that this implementation is ok 👍

njbrake:
Oh, @javiermtorres  also make sure that you update the docs appropriately to document this new feature.

src/any_agent/tools/mcp/frameworks/__init__.py
javiermtorres: Hum..... Yes, most probably :-/ I'm sure there a lot of people with SSE tools right now, but it's true that it shouldn't be used anymore, so I feel it's ok to give users a heads-up.

tests/conftest.py
javiermtorres: Oomph 😅

Again the port issue. Since we have xdist, we have several test workers starting their own fixtures at session scope. This means that sometimes, depending on how the tests are being handed out to workers, the different servers won't start because they're binding to the same port. I noticed 4-5 failures (instead of 1) when using 12 workers locally, which suits the hypothesis, and the failures went away when assigning different ports. Basically, since each worker will start their own copy of the fixture, the worker id is retrieved and added to the port base, effectively getting non-overlapping servers.

The SSE fixture was iirc being used in one test (if it's a parameterized per platform, I'll recheck my hypothesis), so it wouldn't start more than once. Also, if all test instances fall in the same worker somehow, we won't have this problem (but being concurrency as it is, at some point the default for test-dealing could be work-stealing and the situation might fll unexpectedly).

We could take a closer look at how the servers are created in fixtures. When https://github.com/mozilla-ai/any-agent/issues/452 is done, we can most probably work out something similar for the MCP servers as a test util, so we bind to available ports and this problem would be solved.

_UPDATE:_ it seems xdist is waiving the problem away 😁
https://pytest-xdist.readthedocs.io/en/stable/how-to.html#making-session-scoped-fixtures-execute-only-once
We can maybe follow this approach, or see if we could run things reliably on the controller instead of in the workers. Or put tests into suites/groups to ensure they're scheduled together.

tests/integration/test_agent.py
javiermtorres: I don't know... I'm following the existing test case, but I don't know if every test would need to record the spans?

javiermtorres:
> also make sure that you update the docs appropriately to document this new feature

Ouch, totally true 😔

tests/integration/test_agent.py
njbrake: It shouldn't. I think you can delete the update-trace-assets part of the code 👍

tests/integration/test_agent.py
daavoo: can we put this test in another file? Maybe under the existing `tests/integration/mcp`?

javiermtorres: Ok, I'll move it 👍🏼

src/any_agent/tools/mcp/frameworks/agno.py
```diff
@@ -68,18 +71,57 @@ class AgnoMCPSseConnection(AgnoMCPConnection):
 
     async def list_tools(self) -> list["AgnoMCPTools"]:
         """List tools from the MCP server."""
+        warnings.warn(
```
daavoo: I think we could just raise the warning once (when instantiating a MCPSSe config object)

javiermtorres: I'll add a custom validator to the MCPSse model then.

javiermtorres:
Added a test to check the deprecation warning.

src/any_agent/config.py
```diff
@@ -66,6 +67,37 @@ class MCPStdio(BaseModel):
 
 
 class MCPSse(BaseModel):
+    @model_validator(mode="before")
+    @classmethod
+    def sse_deprecation(cls, data: Any) -> Any:
+        warnings.warn(
+            "SSE is deprecated in the MCP specification as of version 2025-03-26",
```
njbrake: Maybe something like this to point the developer in the direction of what they should do about it.
```suggestion
            "SSE is deprecated in the MCP specification in favor of Streamable HTTP",
```

tests/integration/mcp/test_mcp_streamable_http.py
```diff
@@ -0,0 +1,214 @@
...
+def uvx_installed() -> bool:
+    try:
+        result = subprocess.run(
+            ["uvx", "--version"],  # noqa: S607
+            capture_output=True,
+            check=True,
+        )
+        return True if result.returncode == 0 else False  # noqa: TRY300
+    except Exception:
+        return False
```
njbrake: I'm getting dejavu. Why do I continue to see this function copied and pasted into tests 😅 . If a function is being re-used, please put this in a helper function. We want to avoid copying the same functions whenever posible.

tests/integration/mcp/test_mcp_streamable_http.py
```diff
@@ -0,0 +1,214 @@
...
+    def assert_first_llm_call(llm_call: AgentSpan) -> None:
+        """Checks the `_set_llm_inputs` implemented by each framework's instrumentation."""
+        assert llm_call.attributes.get("gen_ai.input.messages", None) is not None
+        # input.messages should be a valid JSON string (list of dicts)
+        input_messages_raw = llm_call.attributes.get("gen_ai.input.messages")
+        assert input_messages_raw is not None
+        input_messages = json.loads(input_messages_raw)
+        assert input_messages[0]["role"] == "system"
+        assert input_messages[1]["role"] == "user"
+
```
njbrake, line 34: Let's move this into the any_agent.testing.helpers if this is the same code as where it used to exist.

src/any_agent/config.py
javiermtorres: Ok, I'll include both bits

tests/integration/mcp/test_mcp_streamable_http.py
daavoo: wait why are we doing the same assertions as in `test_agent` in the first place? 
Do we expect the mcp mode to change anything related to what is being asserted here?

tests/integration/mcp/test_mcp_streamable_http.py
javiermtorres: Since this function is only used in tests and will not be useful for the general user, would it be ok to move it to conftest instead? I don't like to mix test helpers with the source code, unless the helpers are useful in themselves, outside a testing setting.

daavoo: is the function even used in this tests? the uvx stuff is only used for mcp Stdio as far as I can see

tests/integration/mcp/test_mcp_streamable_http.py
javiermtorres: Hum. I normally keep the test code as specific as possible, specially assertions, since tests quickly deviate from the norm and one needs to make a fresh copy (helpers are another issue since they are used in the same way in all tests). In this case, the assertions are fairly general, but they don't need to be. In this case, they differ in the tiny detail of the number of tool executions (1 vs 2), in the middle of the code, which imho makes it difficult to generalize the function to more cases.

javiermtorres:
> it has resulted in a decent amount of code duplication.

My rule to duplicate or not test code is whether it is helper code that can be used as-is in all cases. I prefer to have assertion related code in the test itself, so they dev does not need to jump across files to check a case. Every case also tends to have subtly different assertions, so often it doesn't pay to try and generalize.

I fully agree with the `uvx` comment, though, and I'll check if there is something else that could be extracted in this case. But I'd say this is better solved in a further PR, taking all test cases into consideration.

tests/integration/mcp/test_mcp_streamable_http.py
javiermtorres: Hum, in this case it was an oversight. It is not used anyway in the new test. I'll remove it.

javiermtorres:
...but there's more code that isn't used in the test. Sorry, I thought my IDE was keeping track of it. I'll take a closer look.

tests/integration/mcp/test_mcp_streamable_http.py
daavoo: my point was if these assertions are actually needed for your new test

javiermtorres: I take them as sanity checks. I've removed a few, but I don't feel the remaining will take extra time to check.

javiermtorres: And yes, we're checking the number of tool calls, etc. The test assumes that, if a final response is given and there is a tool call, then the streamable http connection has succeeded.

javiermtorres:
Hi @daavoo @njbrake could you please take a last look at https://github.com/mozilla-ai/any-agent/pull/598/files#diff-4e5a8f6e342c5ff4f37529fc33ab6c7deb391e1cba5c185eef2ed01272359eb1 ? Everything else should be in place.

I have removed the unused checks, factored some functionality with potential for reuse, and deleted the trace dump. I feel that the rest of the checks don't hurt (and I'd keep as many checks as possible in integration/e2e checks) but if you insist I can remove some of them.

javiermtorres:
Integration tests running ok: https://github.com/mozilla-ai/any-agent/actions/runs/16315619603
But there are a few unrelated tests that seem a bit flaky, like https://github.com/mozilla-ai/any-agent/actions/runs/16315359485/job/46080324581

PR #603, opened 2025-07-09
docs(callbacks): Add `SensitiveDataOffloader` example.

daavoo:
I feel like having an example in the docs would be more useful than providing an actual callback implementation.

### DIFF
--- a/docs/agents/callbacks.md
+++ b/docs/agents/callbacks.md
@@ -188,3 +188,34 @@ class ContentFilter(Callback):
 !!! warning
 
     Raising exceptions in callbacks will terminate the agent run immediately. Use this feature carefully to implement safety measures or validation logic.
+
+## Example: Offloading sensitive information
+
+Some inputs and/or outputs in your traces might contain sensitive information that you don't want
+to be exposed in the [traces](../tracing.md).
+
+You can use callbacks to offload the sensitive information to an external location and replace the span
+attributes with a reference to that location:
+
+```python
+from pathlib import Path
+from time import time
+
+class SensitiveDataOffloader(Callback):
+
+    def __init__(self, output_dir: str) -> None:
+        self.output_dir = Path(output_dir)
+
+    def before_llm_call(self, context: Context, *args, **kwargs) -> Context:
+
+        span = context.current_span
+
+        input_messages = span.attributes.pop("gen_ai.input.messages")
+
+        output_file = self.output_dir / f"{time.now()}.txt"
+        output_file.write_text(str(input_messages))
+
+        span.set_attribute("gen_ai.input.messages_ref", str(output_file))
+
+        return context
+```


docs/agents/callbacks.md
```diff
@@ -188,3 +188,34 @@ class ContentFilter(Callback):
...
+        span = context.current_span
+
+        input_messages = span.attributes.pop("gen_ai.input.messages")
+
+        output_file = self.output_dir / f"{time.now()}.txt"
+        output_file.write_text(str(input_messages))
+
+        span.set_attribute("gen_ai.input.messages_ref", str(output_file))
+
+        return context
```
njbrake: May be worth adding a silly if. statement like `if 'my SSN is` in any(span.attributes.get("gen_ai.input.messages"))` so that the example shows selectively wiping out the attribute as opposed to always removing it

njbrake (approved):
Is there a simple unit test that could be written to ensure this logic would work, or is that already covered by tests? I'm not sure if removing the input_messages attribute would break some downstream design assumptions.

daavoo:
> Is there a simple unit test that could be written to ensure this logic would work, or is that already covered by tests? I'm not sure if removing the input_messages attribute would break some downstream design assumptions.

Good point, I will add a test for this

docs/agents/callbacks.md
daavoo: I ended not adding this because I would prefer to showcase the original use case of #437 and adding a condition should be (IMO) simple enough to understand

daavoo:
@njbrake decided to add a cookbook instead

docs/cookbook/callbacks.ipynb
```diff
@@ -0,0 +1,744 @@
+{
```
njbrake, line 1: General comment: can reasonably you do this without Tavily? It's an added complexity for someone that would want to run the cookbook. I would rather that the only requirement is to have the MISTRAL_API_KEY set

pyproject.toml
```diff
@@ -126,6 +126,7 @@ extend-select = ["ALL"]
 "src/any_agent/config.py" = ["D101", "D102"]
 "src/any_agent/evaluation/*" = ["D101", "D103", "D107"]  # TODO: Revisit when working on evaluation
 "src/any_agent/callbacks/*" = ["D102"]
+"docs/cookbook/*" = ["D"]
```
njbrake, line 129: What is this skipping?

docs/cookbook/callbacks.ipynb
daavoo: I was getting some rate limit errors locally. Will try to update it in the CI and see if it works with ddg

pyproject.toml
daavoo: Missing docstrings on public classes (for the callback class definition inside the cookbook)

PR #606, opened 2025-07-09
feat: remove sync serving option

njbrake:
Putting this PR here to see how you all feel about it.

The sync server implementation is basically a wrapper around our async functions except that it adds a bunch of complexity and testing needs. Additionally, if someone is serving an agent, arguably they should not be using the sync version anyways for performance reasons.

With this in mind, I propose removing our `anyagent.serve` function, and adding a short section to our documentation explaining how easy it is to wrap the server in a sync function and run it with asyncio if they choose to do so.

### DIFF
--- a/docs/serving.md
+++ b/docs/serving.md
@@ -10,7 +10,41 @@ these protocols, as explaining them is out of the scope of this page.
 
 In order to use A2A serving, you must first install the 'a2a' extra: `pip install 'any-agent[a2a]'`
 
-You can configure and serve an agent using the [`A2AServingConfig`][any_agent.serving.A2AServingConfig] and the [`AnyAgent.serve`][any_agent.AnyAgent.serve] or [`AnyAgent.serve_async`][any_agent.AnyAgent.serve_async] method.
+You can configure and serve an agent using the [`A2AServingConfig`][any_agent.serving.A2AServingConfig] and the [`AnyAgent.serve_async`][any_agent.AnyAgent.serve_async] method.
+
+## Running Async Servers in Sync Environments
+
+Since `any-agent` uses async/await patterns for better performance and resource management, the serving functions are async by default. However, you can easily run async servers in sync environments using Python's `asyncio` utilities:
+
+### Using `asyncio.run()`
+
+The simplest approach is to wrap your async code in `asyncio.run()`:
+
+```python
+import asyncio
+from any_agent import AgentConfig, AnyAgent
+from any_agent.serving import A2AServingConfig
+
+async def main():
+    agent = await AnyAgent.create_async(
+        "tinyagent",
+        AgentConfig(
+            name="my_agent",
+            model_id="mistral/mistral-small-latest",
+            description="A helpful agent"
+        )
+    )
+
+    server_handle = await agent.serve_async(A2AServingConfig(port=8080))
+
+    try:
+        # Keep the server running
+        await server_handle.task
+    except KeyboardInterrupt:
+        await server_handle.shutdown()
+
+asyncio.run(main())
+```
 
 ## Serving via A2A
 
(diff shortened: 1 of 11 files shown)


javiermtorres:
Before getting into details, I definitely agree with the approach. I feel most frameworks are going async (maybe except SmolAgents?) since they need to juggle several dialogues towards llms anyway, so I don't think we lose much by going full async ourselves. This may put some burden on the user in case they want to use a purely sync interface, but I think we can address this with the docs.

njbrake:
> Before getting into details, I definitely agree with the approach. I feel most frameworks are going async (maybe except SmolAgents?) since they need to juggle several dialogues towards llms anyway, so I don't think we lose much by going full async ourselves. This may put some burden on the user in case they want to use a purely sync interface, but I think we can address this with the docs.

Just to clarify: I'm not proposing that we go full async. I'm only proposing that we remove the option to serve an agent over the sync api. This PR will not affect the agent.run() sync method 👍

javiermtorres:
Fine for me, serving is where the async behaviour will be most useful. Thanks!!!!

docs/serving.md
```diff
@@ -10,7 +10,41 @@ these protocols, as explaining them is out of the scope of this page.
...
 In order to use A2A serving, you must first install the 'a2a' extra: `pip install 'any-agent[a2a]'`
 
-You can configure and serve an agent using the [`A2AServingConfig`][any_agent.serving.A2AServingConfig] and the [`AnyAgent.serve`][any_agent.AnyAgent.serve] or [`AnyAgent.serve_async`][any_agent.AnyAgent.serve_async] method.
+You can configure and serve an agent using the [`A2AServingConfig`][any_agent.serving.A2AServingConfig] and the [`AnyAgent.serve_async`][any_agent.AnyAgent.serve_async] method.
+
+## Running Async Servers in Sync Environments
+
+Since `any-agent` uses async/await patterns for better performance and resource management, the serving functions are async by default. However, you can easily run async servers in sync environments using Python's `asyncio` utilities:
+
+### Using `asyncio.run()`
```
javiermtorres, line 19: I wouldn't see the need for a separate heading since we are showcasing only one package. I assume going for the base async is going to be a bit harder for the casual user. Would it be helpful to point out other alternatives (provided that there is any as relevant as asyncio)?

docs/serving.md
```diff
@@ -23,43 +57,53 @@ For illustrative purposes, we are going to define 2 separate scripts, each defin
...
+                model_id="mistral/mistral-small-latest",
+                description="An agent that can answer questions specifically and only about the Google Agents Development Kit (ADK). Reject questions about anything else.",
+                tools=[search_web]
+            )
         )
-    )
 
-    agent.serve(A2AServingConfig(port=5001))
+        server_handle = await agent.serve_async(A2AServingConfig(port=5001))
+        await server_handle.task
```
javiermtorres, line 77: What about.....
```suggestion
        server_handle = await agent.serve_async(A2AServingConfig(port=5001))
        # Other work can be started and awaited
        await server_handle.task
```
But probably if you are using async you know why :-/

docs/serving.md
```diff
@@ -23,43 +57,53 @@ For illustrative purposes, we are going to define 2 separate scripts, each defin
...
-    )
 
-    agent.serve(A2AServingConfig(port=5002))
+        server_handle = await agent.serve_async(A2AServingConfig(port=5002))
+        await server_handle.task
+
+    asyncio.run(main())
     ```
 
 We can then run each of the scripts in a separate terminal and leave them running in the background.
```
javiermtorres, line 109: Is it worth mentioning that we can run both on the same async loop?

njbrake:
Integration tests https://github.com/mozilla-ai/any-agent/actions/runs/16194938082
Cookbook tests https://github.com/mozilla-ai/any-agent/actions/runs/16194941177

PR #608, opened 2025-07-09
chore: Vendor `langchain_litellm`.

daavoo:


daavoo:
https://github.com/mozilla-ai/any-agent/actions/runs/16171268530

src/any_agent/frameworks/any_agent.py
```diff
@@ -30,6 +31,9 @@
     from any_agent.tools.mcp.mcp_server import _MCPServerBase
 
 
+litellm.drop_params = True
+
+
```
njbrake: I think this will bite us in the butt later when params are silently ignored.

njbrake (changes requested):
I'm worried about adding a global drop_unsupported_params.

src/any_agent/frameworks/any_agent.py
daavoo: dropped it

PR #611, opened 2025-07-09
Cache MyPy and Pre-Commit for faster lint

njbrake:
Cache both pre-commit (uv env/python deps) and mypy (cached mypy command artifacts)

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -27,14 +27,14 @@ jobs:
       - uses: actions/cache@v4
         with:
           path: .mypy_cache
-          key: ${{ runner.os }}-mypy-${{ hashFiles('**/pyproject.toml') }}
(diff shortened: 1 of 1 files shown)


njbrake:
Normally this took 2.5 minutes. Now it's 46 seconds

Screenshot 2025-07-09 at 1 22 33 PM

daavoo (approved):
Not a blocker: I would've been nice to have 2 changes decoupled to measure improvements separately (I remember that caching the pre-commit a while back was not bringing much improvement)

njbrake:
> Not a blocker: I would've been nice to have 2 changes decoupled to measure improvements separately (I remember that caching the pre-commit a while back was not bringing much improvement)


I checked this: the precommit caching only gave about a 15 second improvement (this is what we cached before)

The mypy cache is the caching that gives the big boost and slices off the minutes of time.

PR #612, opened 2025-07-09
docs: wider margin for main part of screen

njbrake:


njbrake:
Was
Screenshot 2025-07-09 at 2 38 59 PM


Now:
Screenshot 2025-07-09 at 2 39 16 PM

docs/assets/custom.css
```diff
@@ -34,3 +34,8 @@
   vertical-align: middle;
   margin-left: 0.3em;
 }
+
+/*  From https://github.com/squidfunk/mkdocs-material/discussions/6404#discussioncomment-7695243 */
+.md-main__inner.md-grid {
+  max-width: initial;
```
daavoo, line 41: would there be a way to only apply this to specific pages (i.e. the cookbook and tracing)? I am not sure if I really like this setting for other pages (but I am ok with going with this for now)

PR #614, opened 2025-07-10
feat(tracing): Add `any_agent/tracing/attributes`.

daavoo:
- Define constants and reuse them across the codebase.
- Add references in docs.

### DIFF
--- a/src/any_agent/callbacks/span_cost.py
+++ b/src/any_agent/callbacks/span_cost.py
@@ -21,21 +21,25 @@ if TYPE_CHECKING:
 def add_cost_info(span: Span) -> None:
     """Use litellm to compute cost and add it to span attributes."""
     attributes: Mapping[str, AttributeValue] = span.attributes
-    try:
-        cost_prompt, cost_completion = cost_per_token(
-            model=str(attributes.get(GenAI.REQUEST_MODEL, "")),
-            prompt_tokens=int(attributes.get(GenAI.USAGE_INPUT_TOKENS, 0)),  # type: ignore[arg-type]
-            completion_tokens=int(attributes.get(GenAI.USAGE_OUTPUT_TOKENS, 0)),  # type: ignore[arg-type]
-        )
-        span.set_attributes(
-            {
-                GenAI.USAGE_INPUT_COST: cost_prompt,
-                GenAI.USAGE_OUTPUT_COST: cost_completion,
-            }
-        )
-    except Exception as e:
-        msg = f"Error computing cost_per_token: {e}"
-        logger.warning(msg)
+    if any(
+        key in attributes
+        for key in ["gen_ai.usage.input_tokens", "gen_ai.usage.output_tokens"]
+    ):
+        try:
+            cost_prompt, cost_completion = cost_per_token(
+                model=str(attributes.get(GenAI.REQUEST_MODEL, "")),
+                prompt_tokens=int(attributes.get(GenAI.USAGE_INPUT_TOKENS, 0)),  # type: ignore[arg-type]
+                completion_tokens=int(attributes.get(GenAI.USAGE_OUTPUT_TOKENS, 0)),  # type: ignore[arg-type]
+            )
+            span.set_attributes(
+                {
+                    GenAI.USAGE_INPUT_COST: cost_prompt,
+                    GenAI.USAGE_OUTPUT_COST: cost_completion,
+                }
+            )
+        except Exception as e:
+            msg = f"Error computing cost_per_token: {e}"
+            logger.warning(msg)
 
 
 class AddCostInfo(Callback):
--- a/tests/unit/callbacks/test_span_cost.py
+++ b/tests/unit/callbacks/test_span_cost.py
@@ -80,6 +80,4 @@ def test_span_cost_missing_all() -> None:
 
     callback.after_llm_call(context)
 
-    call_args = context.current_span.set_attributes.call_args[0][0]
-    assert call_args[GenAI.USAGE_INPUT_COST] == 0
-    assert call_args[GenAI.USAGE_OUTPUT_COST] == 0
+    context.current_span.set_attributes.assert_not_called()


src/any_agent/tracing/attributes.py
```diff
@@ -0,0 +1,28 @@
+"""Constants exported for convenience to access span attributes.
```
njbrake: No strong opinion here but just offering up that this may be a candidate to be a `class GenAI(str, Enum)` instead of a collection of constants.

This might make it a more logical collection instead of generic constants like `OUTPUT` which might get confusing to people reading the code.

tests/unit/callbacks/test_span_cost.py
```diff
@@ -1,14 +1,15 @@
...
+from any_agent.tracing.attributes import MODEL_ID
 
 
 def test_span_cost() -> None:
     context = MagicMock()
     current_span = MagicMock()
 
     current_span.attributes = {
-        "gen_ai.request.model": "gpt-4.1-mini",
+        MODEL_ID: "mistral/mistral-small-latest",
```
njbrake: You could also use `DEFAULT_SMALL_MODEL_ID` from the any_agent.testing.helpers module.

tests/unit/tracing/test_agent_trace.py
```diff
@@ -10,7 +11,7 @@ def create_llm_span(input_tokens: int = 100, output_tokens: int = 50) -> AgentSp
         status=Status(),
         context=SpanContext(span_id=123),
         attributes={
-            "gen_ai.operation.name": "call_llm",
+            OPERATION: "call_llm",
             "gen_ai.usage.input_tokens": input_tokens,
             "gen_ai.usage.output_tokens": output_tokens,
```
njbrake: Why not have usage as a constant enum as well?

src/any_agent/tracing/attributes.py
daavoo: Since all these is defined in a separate module, would it be ok for you if I just update the imports and docs to use something like:

```
from any_agent.tracing import span_attrs
...
    span_attrs.OUTPUT: "foo" 
```

I don't really see what would the benefit of adding a separate class or Enum

njbrake: Yep fine with me 👍

tests/unit/callbacks/test_span_cost.py
daavoo: done

tests/unit/tracing/test_agent_trace.py
daavoo: Missed these (and input/output cost). Pushed an update

src/any_agent/frameworks/any_agent.py
```diff
@@ -202,11 +203,11 @@ async def run_async(self, prompt: str, **kwargs: Any) -> AgentTrace:
 
                 invoke_span.set_attributes(
                     {
-                        "gen_ai.operation.name": "invoke_agent",
+                        span_attrs.OPERATION: "invoke_agent",
                         "gen_ai.agent.name": self.config.name,
                         "gen_ai.agent.description": self.config.description
```
njbrake: Should these also be constants?

src/any_agent/tracing/span_attrs.py
```diff
@@ -0,0 +1,40 @@
+"""Constants exported for convenience to access span attributes.
+
+Trying to follow OpenTelemetry's [Semantic Conventions for Generative AI](https://opentelemetry.io/docs/specs/semconv/gen-ai/).
```
njbrake: Sorry, I should have thought of this earlier and made the comment: can we import them from https://github.com/open-telemetry/opentelemetry-python/blob/0e4cb44858b35bb756ecf79a52779466e7b43356/opentelemetry-semantic-conventions/src/opentelemetry/semconv/_incubating/attributes/gen_ai_attributes.py#L25

daavoo: In order to document the stuff and because we use some attributes that are not part of the semconv (yet), I ended up exposing a `GenAI` class with the attributes. U

docs/agents/callbacks.md
```diff
@@ -40,11 +40,14 @@ while True:
...
+## Context
+
 During the agent run ( [`agent.run_async`][any_agent.AnyAgent.run_async] or [`agent.run`][any_agent.AnyAgent.run] ), a unique [`Context`][any_agent.callbacks.context.Context] object is created and shared across all callbacks.
 
 `any-agent` populates the [`Context.current_span`][any_agent.callbacks.context.Context.current_span]
-property so that callbacks can access information in a framework-agnostic way. You can check the attributes available
-for LLM Calls and Tool Executions in the [example spans](../tracing.md#spans).
+property so that callbacks can access information in a framework-agnostic way.
+
+You can check the attributes available for LLM Calls and Tool Executions in the [`GenAI`][any_agent.tracing.attributes.GenAI].
```
njbrake: Something like
```suggestion
You can see what attributes are available for LLM Calls and Tool Executions by examining the [`GenAI`][any_agent.tracing.attributes.GenAI] class.
```

src/any_agent/callbacks/span_cost.py
```diff
@@ -20,25 +21,21 @@
...
-                    "gen_ai.usage.input_cost": cost_prompt,
-                    "gen_ai.usage.output_cost": cost_completion,
-                }
-            )
-        except Exception as e:
-            msg = f"Error computing cost_per_token: {e}"
-            logger.warning(msg)
+    try:
+        cost_prompt, cost_completion = cost_per_token(
+            model=str(attributes.get(GenAI.REQUEST_MODEL, "")),
```
njbrake: Why the behavior change here? The if statement disappeared

src/any_agent/tools/a2a.py
```diff
@@ -68,7 +68,7 @@ async def a2a_tool_async(
     # with the traditional Optional[T] syntax for automatic function calling.
     # Using T | None syntax causes"Failed to parse the parameter ... for automatic function calling"
     async def _send_query(
-        query: str, task_id: Optional[str] = None, context_id: Optional[str] = None
+        query: str, task_id: str | None = None, context_id: str | None = None
```
njbrake: Why does my note on lines 66-68 no longer apply? If true, then please remove my note

src/any_agent/tracing/agent_trace.py
```diff
@@ -318,12 +317,6 @@ def cost(self) -> CostInfo:
...
             if span.is_llm_call():
-                if any(
-                    key in span.attributes
-                    for key in ["gen_ai.usage.input_cost", "gen_ai.usage.output_cost"]
-                ):
-                    sum_input_cost += span.attributes.get("gen_ai.usage.input_cost", 0)
-                    sum_output_cost += span.attributes.get(
-                        "gen_ai.usage.output_cost", 0
-                    )
+                sum_input_cost += span.attributes.get(GenAI.USAGE_INPUT_COST, 0)
```
njbrake, line 320: Not sure why the if statement was removed

njbrake (changes requested):
I am 👍  on the addition of the attributes, but a few functional changes were made in this PR that need to be clarified.

src/any_agent/tracing/agent_trace.py
daavoo: I should/could have applied the changes in a separate PR but I just noticed it when applying this changes.

The `if` is irrelevant/redundant for the final behavior, as summing 0 is equivalent to not doing anything

src/any_agent/tools/a2a.py
daavoo: I messed something with the pre-commit somehow 🙏 Reverting these

src/any_agent/tools/a2a.py
```diff
@@ -173,8 +173,8 @@ async def _send_query(
 
 
 def a2a_tool(
-    url: str, toolname: Optional[str] = None, http_kwargs: dict[str, Any] | None = None
-) -> Callable[[str, Optional[str], Optional[str]], str]:
+    url: str, toolname: str | None = None, http_kwargs: dict[str, Any] | None = None
```
javiermtorres: Just curious, why remove the Optional? Was it giving headaches in type checking?

src/any_agent/callbacks/span_cost.py
daavoo: reverted

src/any_agent/tools/a2a.py
daavoo: This was reverted. my auto-formatter messed it up

tests/integration/test_agent.py
```diff
@@ -41,19 +42,17 @@ def uvx_installed() -> bool:
 def assert_trace(agent_trace: AgentTrace, agent_framework: AgentFramework) -> None:
     def assert_first_llm_call(llm_call: AgentSpan) -> None:
         """Checks the `_set_llm_inputs` implemented by each framework's instrumentation."""
-        assert llm_call.attributes.get("gen_ai.input.messages", None) is not None
-        # input.messages should be a valid JSON string (list of dicts)
-        input_messages_raw = llm_call.attributes.get("gen_ai.input.messages")
+        input_messages_raw = llm_call.attributes.get(GenAI.INPUT_MESSAGES)
```
njbrake, line 45: not sure if you meant to get rid of the assert

njbrake (approved):
🚀

tests/integration/test_agent.py
daavoo: yes, because it is redundant (same is not None was running right below after assigning it to a variable)

PR #615, opened 2025-07-10
docs: more mkdocs features

njbrake:
* Noto Sans font instead of Robot for a little differentiation from the standard (used by https://hummingbot.org/docs/)
* Take account for user preference to auto-pick light vs dark theme
* Change header/footer bar color from black to blue gray
* auto expand sidebar (since currently we don't have too many different pages, we can re-assess as our docs grow
* Add navigation footer to easily go to next page when finished reading current page.

### DIFF
--- a/tests/snapshots/__snapshots__/test_trace.ambr
+++ b/tests/snapshots/__snapshots__/test_trace.ambr
@@ -1,5 +1,5 @@
 # serializer version: 1
-# name: test_agent_trace_snapshot[AGNO_trace].330008117380762187083381322541040191233
+# name: test_agent_trace_snapshot[AGNO_trace].219771956564580706934509047402402461169
   dict({
     'final_output': dict({
     }),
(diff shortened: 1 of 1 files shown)


njbrake:
Screenshot 2025-07-10 at 8 48 28 AM

mkdocs.yml
```diff
@@ -38,23 +38,31 @@ nav:
...
   name: material
+  font:
+    text: Noto Sans
+    code: Noto Sans Mono
   palette:
-    - scheme: default
-      primary: black
+    - media: "(prefers-color-scheme: light)"
+      scheme: default
+      primary: blue gray
```
daavoo, line 47: why the change in color?

njbrake: I'm no style guy, but black seemed a little lifeless to me, especilally in dark mode. blue gray is still dark but gives a little more pop.

hasangzl:
The new font and color look good to me, but I suggest adding the full logo to the navbar instead of just the thumbnail and “any-agent” text in a different font. Because right now, it kinda looks like another logo.


Screenshot 2025-07-11 at 11 52 48 am

PR #618, opened 2025-07-10
Option 1: feat: breaking change with requirements for A2A serving

njbrake:
Pros:

* There is no more re-creating of the agent when serving, which makes it more reliable when attaching and accessing attributes of the agent that is being served
* It makes it clear what a2a serving is doing under the hood: namely that it requires structured outputs.

Cons:

Serving an Agent over A2A is now more complicated.


docs/serving.md
```diff
@@ -12,6 +12,35 @@ In order to use A2A serving, you must first install the 'a2a' extra: `pip instal
 
 You can configure and serve an agent using the [`A2AServingConfig`][any_agent.serving.A2AServingConfig] and the [`AnyAgent.serve_async`][any_agent.AnyAgent.serve_async] method.
 
+## A2A Output Type Requirements
+
+When serving an agent via A2A, you **must** configure your agent's `output_type` to inherit from [`A2AEnvelope`][any_agent.serving.A2AEnvelope]. This is required because A2A responses need to include task status information along with the actual response data.
```
javiermtorres, line 17: Hum... Can't we provide this wrapping when serving, instead of putting the burden on the developer?

javiermtorres:
What was the reason to copy the agent, in the first place?

docs/serving.md
njbrake: This is the reason for this pull request. We currently provide the wrapping when serving, which is what forces the copying of the agent.

javiermtorres: Please excuse any dumb questions 😅
Why do we need to copy the agent when serving?

njbrake: No worries! I didn't explicitly explain what's going on.

It's because we are changing the output_type parameter which you cannot do after agent initiation because our code sets parameters based on the output_type parameter in agent initialization (the things it does are different based on framework). 

That being said, this has me thinking, I wonder if we could get around this by adding a `update_output_type` member function to the AnyAgent class. That design might give us the best of both worlds, if possible.. Hmm let me workshop with this for a bit.

njbrake:
Going with option 2, closing this.

PR #620, opened 2025-07-10
llms.txt and llms-full.txt

njbrake:
https://llmstxt.org/

https://mintlify.com/blog/the-value-of-llms-txt-hype-or-real


.gitignore
```diff
@@ -151,6 +151,9 @@ venv.bak/
 
 # mkdocs documentation
 /site
+# Generated llms files (dynamically created at build time)
+llms.txt
+llms-full.txt
```
daavoo: why are these gitignored?

njbrake: Good point, they're already gitignored because they get generated into the `site` directory at build time (which is already gitignored). I'll remove

njbrake:
```suggestion
```

scripts/hooks.py
```diff
@@ -0,0 +1,301 @@
+"""Generate llms.txt and llms-full.txt files for the documentation.
+
+It also serves markdown files as text files for LLM consumption.
+It was initiallywritten in collaboration with Claude 4 Sonnet.
```
njbrake:
```suggestion
It was initially written in collaboration with Claude 4 Sonnet.
```

scripts/hooks.py
```diff
@@ -0,0 +1,301 @@
...
+def on_post_build(config, **kwargs):
+    """Generate llms.txt and llms-full.txt files, and serve markdown as text."""
+    docs_dir = Path(config["docs_dir"])
+    site_dir = Path(config["site_dir"])
+
+    # Get navigation configuration
+    nav_config = config.get("nav", [])
+
+    # Serve markdown files as text
+    serve_markdown_as_text(docs_dir, site_dir, nav_config)
```
javiermtorres, line 295: Being really picky, I'd rename this as `convert_markdown_to_text` since the serving is a separate process from doc generation (like if some param was being given to `mkdocs serve` to set the content type while serving).

PR #621, opened 2025-07-11
fix: consistent prompts and simplification of structured output prompt

njbrake:
* turn class into function because there's no need for the class
* Now all the frameworks that add the final answer as a tool use the same function and prompt.

I have some work left on the frameworks that weren't using FinalOutputTool, which I'll tackle next.

### DIFF
--- a/src/any_agent/frameworks/google.py
+++ b/src/any_agent/frameworks/google.py
@@ -4,7 +4,7 @@ from uuid import uuid4
 from pydantic import BaseModel
 
 from any_agent.config import AgentConfig, AgentFramework
-from any_agent.tools.final_output import FinalOutputTool
+from any_agent.tools.final_output import prepare_final_output_config_function
 
 from .any_agent import AnyAgent
 
@@ -61,13 +61,10 @@ class GoogleAgent(AnyAgent):
 
         instructions = self.config.instructions or ""
         if self.config.output_type:
-            instructions += (
-                "You must call the final_output tool when finished."
-                "The 'answer' argument passed to the final_output tool must be a JSON string that matches the following schema:\n"
-                f"{self.config.output_type.model_json_schema()}"
+            instructions, final_output_function = prepare_final_output_config_function(
+                self.config.output_type, instructions
             )
-            output_fn = FinalOutputTool(self.config.output_type)
-            tools.append(output_fn)
+            tools.append(final_output_function)
 
         self._agent = agent_type(
             name=self.config.name,
--- a/src/any_agent/frameworks/smolagents.py
+++ b/src/any_agent/frameworks/smolagents.py
@@ -1,9 +1,11 @@
+from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
 
 from pydantic import BaseModel
 
 from any_agent.config import AgentConfig, AgentFramework
 from any_agent.frameworks.any_agent import AnyAgent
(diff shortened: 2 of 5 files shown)


src/any_agent/tools/final_output.py
```diff
@@ -1,32 +1,31 @@
 import json
+from collections.abc import Callable
 
 from pydantic import BaseModel, ValidationError
 
 
-class FinalOutputTool:
-    """A serializable final output tool that avoids closure issues."""
+def prepare_final_output_config_function(
```
daavoo: 💅 
```suggestion
def prepare_final_output(
```

src/any_agent/frameworks/google.py
```diff
@@ -61,13 +61,10 @@ async def _load_agent(self) -> None:
 
         instructions = self.config.instructions or ""
         if self.config.output_type:
-            instructions += (
-                "You must call the final_output tool when finished."
-                "The 'answer' argument passed to the final_output tool must be a JSON string that matches the following schema:\n"
-                f"{self.config.output_type.model_json_schema()}"
+            instructions, final_output_function = prepare_final_output_config_function(
```
daavoo: 💅 

```suggestion
            instructions, final_output_tool = prepare_final_output_config_function(
```

src/any_agent/frameworks/google.py
```diff
@@ -61,13 +61,10 @@ async def _load_agent(self) -> None:
...
+                self.config.output_type, instructions
             )
-            output_fn = FinalOutputTool(self.config.output_type)
-            tools.append(output_fn)
+            tools.append(final_output_function)
```
daavoo:
```suggestion
            tools.append(final_output_tool)
```

daavoo (approved):
No strong opinion, just name feels verbose and I would prefer to use `tool` instead of `function`

src/any_agent/frameworks/smolagents.py
```diff
@@ -71,28 +73,45 @@ async def _load_agent(self) -> None:
...
-                inputs = {  # noqa: RUF012
-                    "answer": {
-                        "type": "string",
-                        "description": f"The final answer to the problem. The input must be a string that conforms to the{output_type.__name__} object.",
+            instructions, final_output_function = prepare_final_output_config_function(
+                self.config.output_type, self.config.instructions
+            )
+
+            # Create a custom tool for smolagents that wraps our final output function
+            class FinalAnswerToolWrapper(FinalAnswerTool):  # type: ignore[no-untyped-call]
```
javiermtorres, line 81: since there's an init and (presumably) the class does not depend on the enclosing scope, could it be moved out of the `_load_agent` definition?

njbrake: It does depend on the enclosing scope, so unfortunately it needs to stay here.

njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/16219583848

Cookbook tests: https://github.com/mozilla-ai/any-agent/actions/runs/16219588295

src/any_agent/frameworks/smolagents.py
javiermtorres: Isn't the previous local dependency (output_type) now being handled explicitly as the final_output_func member? Or is there some other dependency I'm not noticing? It is defined conditionally, but instanced the same in every call vs being instanced just once if moved out of the function body.

PR #622, opened 2025-07-11
Wrap structured output calls for langchain and tinyagent

njbrake:
I'm new to the wrapped code so i'm not confident that my refactor of langchain callbacks is what you would expect. Def open to opinions here.

### DIFF
--- a/src/any_agent/callbacks/span_generation/langchain.py
+++ b/src/any_agent/callbacks/span_generation/langchain.py
@@ -14,50 +14,99 @@ if TYPE_CHECKING:
 
 class _LangchainSpanGeneration(_SpanGeneration):
     def before_llm_call(self, context: Context, *args, **kwargs) -> Context:
-        messages: list[list[BaseMessage]] = args[1]
-        if not messages or not messages[0]:
+        # Handle both cases: messages as positional args (normal LangChain callback)
+        # or messages as keyword args (when called via call_model wrapper)
+        messages: list[list[BaseMessage]] | list[dict[str, Any]] | None = None
+        
+        if len(args) > 1:
+            # Normal LangChain callback case
+            messages = args[1]
+        elif "messages" in kwargs:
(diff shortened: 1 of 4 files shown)


src/any_agent/callbacks/span_generation/langchain.py
```diff
@@ -7,57 +7,115 @@
...
+                }
+                for message in messages[0]
+            ]
 
         return self._set_llm_input(context, model_id, input_messages)
 
     def after_llm_call(self, context: Context, *args, **kwargs) -> Context:
-        response: LLMResult = args[0]
-        if not response.generations or not response.generations[0]:
+        if not args:
```
daavoo: not sure I get the case of this new if

src/any_agent/callbacks/span_generation/langchain.py
```diff
@@ -7,57 +7,115 @@
...
+                    ]
+
+            input_tokens = 0
+            output_tokens = 0
+            if llm_output := getattr(response, "llm_output", None):
+                if token_usage := llm_output.get("token_usage", None):
+                    input_tokens = token_usage.prompt_tokens
+                    output_tokens = token_usage.completion_tokens
+
+            return self._set_llm_output(context, output, input_tokens, output_tokens)
```
daavoo: 💅 similar to the changes in `_before_llm_call`, you could just have a single return out of the if/else by renaming `litellm_output` to `output`

src/any_agent/callbacks/span_generation/langchain.py
njbrake: Good call, scope creep. I'll make a separate pr to touch up this code a bit

njbrake:
Integration: https://github.com/mozilla-ai/any-agent/actions/runs/16220977222
Cookbook: https://github.com/mozilla-ai/any-agent/actions/runs/16220980774

PR #624, opened 2025-07-11
Option2: feat: fix output type without recreating agent

njbrake:
here's the other option instead of a breaking change: we could handle the recreation on a framework-to-framework basis

https://github.com/mozilla-ai/any-agent/pull/618


javiermtorres:
You're gonna hate me, I think I like the original validation 😅
What do you think about a ServingAgent subclass that adds the serve method to an AnyAgent? In the init we can add the original validation and throw an exception if it doesn't comply with our requirements (or automagically wrap things ourselves). But we'd force the user to be clear about what use they're going to give the agent.

javiermtorres:
My main issue is recreating something... We either let the user create a correct agent, or stop him from creating something that cannot be used in the way they want later on.

daavoo (commented):
Let's go with this one. It opens up the possibility to change how we handle things under the hood without breaking the public API

njbrake:
> @njbrake I'd say we go with this one. It opens up the possibility to change how we handle things under the hood without breaking the public API

Excellent, I agree.

njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/16349443631
Cookbook tests: https://github.com/mozilla-ai/any-agent/actions/runs/16349355582/job/46191693953

PR #632, opened 2025-07-14
feat(hooks): Drop `serve_markdown_as_text` and use links to raw `.md`…

daavoo:
… files instead.

### DIFF
--- a/scripts/hooks.py
+++ b/scripts/hooks.py
@@ -10,8 +10,9 @@ from pathlib import Path
 
 # Constants
 MARKDOWN_EXTENSION = ".md"
-TEXT_EXTENSION = ".txt"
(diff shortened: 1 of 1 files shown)


njbrake (commented):
As long as it works, good with me. I was having trouble getting mkdocs to serve a raw md file so just check that this works before merging.

daavoo:
> As long as it works, good with me. I was having trouble getting mkdocs to serve a raw md file so just check that this works before merging.

I am linking to the raw files hosted in `raw.githubusercontent`, so no mkdocs, because they were full url links so they don't need to live on the generated mkdocs site:

https://raw.githubusercontent.com/mozilla-ai/any-agent/refs/heads/main/docs/index.md

PR #635, opened 2025-07-15
Experimental Support for Any-LLM

njbrake:


pyproject.toml
```diff
@@ -61,6 +61,10 @@ all = [
   "any-agent[agno,google,langchain,llama_index,openai,smolagents]",
 ]
 
+any_llm = [
+  "any-llm-sdk",
```
daavoo, line 65: was `any-llm` taken?

src/any_agent/frameworks/tinyagent.py
```diff
@@ -227,11 +231,19 @@ async def _run_async(self, prompt: str, **kwargs: Any) -> str | BaseModel:
...
+                    if self.use_any_llm or supports_response_schema(
+                        model=self.config.model_id
+                    ):
                         completion_params["response_format"] = self.config.output_type
-                    if len(completion_params["tools"]) > 0:
-                        completion_params["tool_choice"] = "none"
+                    if "tools" in completion_params:
+                        completion_params.pop("tools")
+                        completion_params.pop("tool_choice", None)
+                        completion_params.pop("parallel_tool_calls", None)
```
daavoo, line 241: Wouldn't this change the behavior also if `use_any_llm=False`? Do we want that?

daavoo: I guess the change makes sense because we don't want tools for structured outputs, was just confused about the behavior change

src/any_agent/frameworks/tinyagent.py
```diff
@@ -227,11 +231,19 @@ async def _run_async(self, prompt: str, **kwargs: Any) -> str | BaseModel:
...
                     response = await self.call_model(**completion_params)
+                    if self.use_any_llm:
```
daavoo, line 243: Am I missing something or this is the same return as below? Why duplicate it under `if`?

pyproject.toml
njbrake: Yes :(

src/any_agent/frameworks/tinyagent.py
njbrake: Yeah sorry for not clarifying, It is a behavior change for all cases. LiteLLM wasn't throwing an error but it should have.

src/any_agent/frameworks/tinyagent.py
njbrake: Because LiteLLM didn't type content as an attribute, it's a dict you have to pull. Any-LLM types the content as an attribute so you access it via `.content` instead of `['content']`

src/any_agent/frameworks/tinyagent.py
```diff
@@ -240,8 +252,12 @@ async def _run_async(self, prompt: str, **kwargs: Any) -> str | BaseModel:
         return "Max turns reached"
 
     async def call_model(self, **completion_params: dict[str, Any]) -> ModelResponse:
```
daavoo, line 254: Shouldn't the typing be updated here?

njbrake: Eventually yes. For the moment, I add type:ignore for the returns in call_model because any_llm uses the OpenAI types and I don't want to introduce that dependency into any-agent just yet.

njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/16299477278/job/46030046989

There's a failing test around push notifications, but it's currently failing on main and unrelated to my changes.

PR #636, opened 2025-07-15
humanize: hooks

daavoo:
Came for [Drop extract_description_from_markdown. Use subheadings.](https://github.com/mozilla-ai/any-agent/commit/4c013cffccccf2c898dec8bad36b550efaf7b06c) but couldn't help myself to do some cleanup/humanization of the code.

Some examples:

| Section| Before | After |
|-------- |--------|--------|
| Agents - Index | To define any agent system you will always use the same imports: | Defining Agents - Running Agents | 
| Agents - Callbacks | For greater control when running your agent, any-agent includes support for custom [Callbacks][any_agent.callbacks.base.Callback] that | Context - Implementing Callbacks - Default Callbacks - Providing your own Callbacks - Example: Offloading sensitive information |

### DIFF
--- a/scripts/hooks.py
+++ b/scripts/hooks.py
@@ -10,8 +10,9 @@ from pathlib import Path
 
 # Constants
 MARKDOWN_EXTENSION = ".md"
-TEXT_EXTENSION = ".txt"
-BASE_URL = "https://mozilla-ai.github.io/any-agent/"
+BASE_URL = (
+    "https://raw.githubusercontent.com/mozilla-ai/any-agent/refs/heads/main/docs/"
+)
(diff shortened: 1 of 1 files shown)


javiermtorres:
What if we load the generated llm-related txt files as artifacts as part of the doc process? It would help with issues like this.

daavoo:
> What if we load the generated llm-related txt files as artifacts as part of the doc process? It would help with issues like this.

The generated files can be accessed in https://github.com/mozilla-ai/any-agent/tree/gh-pages .
Locally, you can `mkdocs build` and check them under the gitignored `site` directory.

PR #638, opened 2025-07-15
docs(cookbooks): Evaluation (programmatic, llm, agent)

ividal:
No outputs today as *ahem* I hit the DDG daily limit. 

Notebook starts where my_first_agent.ipynb stopped (I just used a more specific question to help with content evaluation).

- Keeps duckduckgo_search so people don't need another API key, but mentions Tavily as an alternative.
- Shows programmatic, llm and agentic evaluation.
- Shows the issue of qualitative/ambiguous criteria.
- Shows how the agent judge can also use the search tool to verify.


njbrake:
`Notice here again how hard it is to make the agent understand the length requirements. Do play around with them and notice how the assessment changes!`

Can we adjust this cookbook so that it highlights something the value of the agent as judge? I think the idea of giving the agent judge web access is a good idea, but I would like to have a stronger display of its value.

ividal:
Fair point. What I want people to understand with this cookbook is they should "use the right tool for the job". But I agree we can do that in a more positive light, so like we discussed I'll:

* reduce the criteria in the LLMJudge to only qualitative criteria plus the length (to prove the point);
* reduce the number of criteria to AgentJudge to a single qualitative criterion that involves using the search independently.

This way the user can't miss that:
1. if you just care about a measurable criterion, then use plain code, BUT
2. for something qualitative, use an LLM
3. you can have a doer-checker by giving the judge tools to run searches on their own.

ividal:
⚠️ Changed to Tavily as the default to avoid the daily cap, kept the mention of search_web for folks who do not want an API key.

njbrake (approved):
This is great! Before you merge please run the cookbook tests action to make sure it passes (you'll need to manually trigger it from the github actions tab)

and also please update the mkdocs.yaml and readme.md to add this new cookbook. 

Thanks 🏅 🚀 🙏

ividal:
OK, cookbook tests green, all green here... ready to goooo!

PR #643, opened 2025-07-16
Add cookbook example of a fully local agent

Kostis-S-Z:
Add a cookbook example demonstrating how to set up and run a fully offline agent using a local LLM (w/ ollama).

The example in this case is an agent that can read the files in a directory (codebase) and create a summary README.md describing the project. Thus the agent will have both read and write permissions to a directory that the user specifies. To complete the example here, I added a trimmed down version of [structured-q-a](https://github.com/mozilla-ai/structured-q-a) as an example codebase.

Note that the performance will really depend on which model you choose to run locally with Ollama and the context length you provide. In my 22GB RAM computer, granite3.3 with 14k context length works, _most of the time_...

Please try it out locally and let me know any thoughts / concerns! Especially if you can try with different models / context lengths, that would be great so we can document what works best.


njbrake:
Thanks for the work here!

1: could the ollama be simplified so that the n_ctx parameter is provided at runtime and doesn't require a restart and custom serving of ollama?
2. Could the example be simplified so that it uses the existing any-agent codebase, instead of creating a new toy codebase?
3. Bonus: could you try this with tinyagent + any-llm instead to see how this works?

Kostis-S-Z:
@njbrake thanks for the feedback!

> 1: could the ollama be simplified so that the n_ctx parameter is provided at runtime and doesn't require a restart and custom serving of ollama?

Great point, updated it! This should work now:

```
OLLAMA_CONTEXT_LENGTH=24000 OLLAMA_DEBUG=1 ollama serve
```
```
ollama run granite3.3
```

(I had tried to run `/set parameter num_ctx 40000` but it only takes effect inside that chat in ollama run so when you ping it from another terminal it still uses the default.)

 2. Could the example be simplified so that it uses the existing any-agent codebase, instead of creating a new toy codebase?
 
This was the initial idea, but the any-agent codebase is quite big and I think for many people's local setup it might struggle feeding the whole codebase in the context window they can "afford". I have 22GB RAM and 8GB VRAM and granite3.3 with more than ~30.000 context window is struggling. I could just focus on the `docs`, but I think it would be interesting to showcase that just reading code it can extrapolate a README.

 3. Bonus: could you try this with tinyagent + any-llm instead to see how this works?

I ll give it a try!  Note that this is not working currently:
```
pip install 'any-agent[any_llm]'
```
[This](https://github.com/mozilla-ai/any-agent/commit/ea806051944230808a60372c846583e3998314c7) commit wasn't included in `0.25.0` (see [here](https://github.com/mozilla-ai/any-agent/blob/0.25.0/pyproject.toml)). Not an issue, I can check it out locally, just noting it :)

njbrake:
> (I had tried to run `/set parameter num_ctx 40000` but it only takes effect inside that chat in ollama run so when you ping it from another terminal it still uses the default.)

Can you add it to model_args so that the config dict gets passed to ollama like https://github.com/mozilla-ai/any-llm/blob/main/src/any_llm/providers/ollama/ollama.py#L79 ?

> 
>     2. Could the example be simplified so that it uses the existing any-agent codebase, instead of creating a new toy codebase?
> 
> 
> This was the initial idea, but the any-agent codebase is quite big and I think for many people's local setup it might struggle feeding the whole codebase in the context window they can "afford". I have 22GB RAM and 8GB VRAM and granite3.3 with more than ~30.000 context window is struggling. I could just focus on the `docs`, but I think it would be interesting to showcase that just reading code it can extrapolate a README.

Makes sense. Could you refactor it in some way though so that the user doesn't need to copy the toy codebase alongside of the notebook? Maybe you could have the notebook do a git clone of a public github repo? I'm not a huge fan of the toy_codebase being a folder inside of the cookbooks.

> [This](https://github.com/mozilla-ai/any-agent/commit/ea806051944230808a60372c846583e3998314c7) commit wasn't included in `0.25.0` (see [here](https://github.com/mozilla-ai/any-agent/blob/0.25.0/pyproject.toml)). Not an issue, I can check it out locally, just noting it :)

ooo good point! I'll put out a patch version so that it will be supported.

Kostis-S-Z:
> Can you add it to model_args so that the config dict gets passed to ollama like https://github.com/mozilla-ai/any-llm/blob/main/src/any_llm/providers/ollama/ollama.py#L79 ?

Even better, great shout! Done in [810a0bb](https://github.com/mozilla-ai/any-agent/pull/643/commits/810a0bbe89ca3c0494c656964d96fb44f592d5ec)

> Makes sense. Could you refactor it in some way though so that the user doesn't need to copy the toy codebase alongside of the notebook? Maybe you could have the notebook do a git clone of a public github repo? I'm not a huge fan of the toy_codebase being a folder inside of the cookbooks.

I understand. Initially, I had just set as the target "codebase" the `docs/cookbook` dir to make it simple, but my machine was still struggling since the notebooks are quite large (since they include all of this metadata). I think I will just revert to this anyway. For this example to work successfully, it really depends on the local hardware it will run... Even 16Gb is not really sufficient... see [43ef142](https://github.com/mozilla-ai/any-agent/pull/643/commits/43ef1428917ce5daa5f0385c4e0345ac4c25dff1)
Does that work?

Kostis-S-Z:
@njbrake 
> 3. Bonus: could you try this with tinyagent + any-llm instead to see how this works?

I thought I made it work just one time, but I can't seem to reproduce now. Maybe I hallucinated it... 

Now I keep getting timeouts: 

```
The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<my_path>/any_llm_test.py", line 52, in <module>
    agent_trace = agent.run(
        "Follow the instructions below:"
    ...<4 lines>...
        "3. Finally, create and write this summary in a README.md file in the same directory."
    )
  File "<my_path>/any-agent/src/any_agent/frameworks/any_agent.py", line 164, in run
    return run_async_in_sync(self.run_async(prompt, **kwargs))
  File "<my_path>/any-agent/src/any_agent/utils/asyncio_sync.py", line 45, in run_async_in_sync
    return loop.run_until_complete(coro)
           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "<my_path>/.pyenv/versions/3.13.1/lib/python3.13/asyncio/base_events.py", line 720, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "<my_path>/any-agent/src/any_agent/frameworks/any_agent.py", line 224, in run_async
    raise AgentRunError(trace, e) from e
any_agent.frameworks.any_agent.AgentRunError: timed out
```

Kostis-S-Z:
> @njbrake
> 
> > 3. Bonus: could you try this with tinyagent + any-llm instead to see how this works?> 

Update: Thanks for the hint for the `timeout` arg in `model_args`! Adding  `model_args={"timeout": 1000},` in AgentConfig solves this! Can confirm that ollama + any-agent + any-llm works, though I am still struggling to get meaningful results... I'll actually spin up a VM with a H100 to test with a proper model and come back to you to confirm it worked as expected :muscle:

njbrake:
>  though I am still struggling to get meaningful results...

Just checking, were you getting meaningful results before switching to any_llm? It's not clear if you're saying that any_llm is responsible for reduced performance 😅

Kostis-S-Z:
> Just checking, were you getting meaningful results before switching to any_llm? It's not clear if you're saying that any_llm is responsible for reduced performance 😅

I am trying to figure that out now ^^ I am getting very inconsistent results, but I dont want to fully blame any-llm yet.

Kostis-S-Z:
> Just checking, were you getting meaningful results before switching to any_llm? It's not clear if you're saying that any_llm is responsible for reduced performance 😅

I am trying to figure that out now ^^ I am getting very inconsistent results, but I dont want to fully blame any-llm yet.

I added a callback to track time before/after llm call and these are the results I am getting right now

This is toy test
```
for i in range(10):
    agent.run("List the files in the allowed directory and any folder inside it recursively.")
```
This is the callback:

```
times = []

class Timer(Callback):
    def __init__(self) -> None:
        self.start_time = 0

    def before_llm_call(self, context: Context, *args, **kwargs) -> Context:
        self.start_time = time.time()
        return context

    def after_llm_call(self, context: Context, *args, **kwargs) -> Context:
        time_elapsed = time.time() - self.start_time
        print(f"Time taken for LLM call: {time_elapsed:.2f} seconds")
        times.append(time_elapsed)
        return context
```

litellm: average 0.8 sec 
any-llm: average 24.7 sec :grimacing:

njbrake:
> litellm: average 0.8 sec any-llm: average 24.7 sec 😬

0.8 seconds for a response from a local model seems wild 😅

Kostis-S-Z:
@njbrake I was too ambitious about this: 

> Initially, I had just set as the target "codebase" the docs/cookbook dir to make it simple, but my machine was still struggling since the notebooks are quite large (since they include all of this metadata). I think I will just revert to this anyway.

I just cant make it run, even on 32GB with 40.000 context length... the notebooks are even bigger, and the model starts too hallucinate. I tried git cloning https://github.com/mozilla-ai/structured-qa but its still quite big and maxes out the context length... unless I remove lots of files first... Maybe this just points to that the example use case of the "readme-maker" is just not the right fit for small models :( I ll probably have to think of another use-case...

Kostis-S-Z:
> 0.8 seconds for a response from a local model seems wild 😅

its just an 8b model with 10k context length generating a dozen tokens, so its not _too crazy_ I would say. More than half is also running a RTX2060 GPU

docs/cookbook/agent_with_local_llm.ipynb
```diff
@@ -0,0 +1,256 @@
...
+   "source": [
+    "### Ollama setup\n",
+    "\n",
+    "First, install Ollama by following their instructions: https://ollama.com/download\n",
+    "\n",
+    "### Picking an LLM\n",
+    "\n",
+    "Pick a model that you can run locally based on your hardware and running it in your terminal. For example:\n",
+    "\n",
+    "16-24GB RAM -> `granite3.3`  or  `deepseek-r1:8b` with ~20–35k context length\n",
```
ividal: Nit, as a user who will show this to a bunch of excited people soon 😄 Could you suggest which granite3.3? (For reference, 2b was iffy, but it was a big codebase - maybe your experience was better?)

docs/cookbook/agent_with_local_llm.ipynb
```diff
@@ -0,0 +1,256 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Fully offline Agent!\n",
+    "\n",
+    "_(no colab for this one since the point is to be run entirely locally)_\n",
+    "\n",
+    "This tutorial will guide on how to run agent, fully locally / offline i.e., running with a local LLM and a local MCP server, so no data will leave your machine!\n",
```
ividal: Nit (?), but might help someone: You may want to remind people that they need Docker running, since the MCP servers you use are docker based.

docs/cookbook/agent_with_local_llm.ipynb
```diff
@@ -0,0 +1,256 @@
...
+  {
+   "cell_type": "code",
+   "metadata": {},
+   "source": [
+    "from any_agent import AgentConfig, AnyAgent\n",
+    "from any_agent.tools import show_plan\n",
+    "\n",
+    "# Define the agent\n",
+    "agent = AnyAgent.create(\n",
+    "    \"smolagents\",\n",
```
ividal, line 215: Not sure if it's just me: `smolagents` yielded a ToolCallingAgent not found for me. `tinyagent` did work.

ividal:
Adding comments as an eager user 😁

ividal (commented):
Adding comments as a happy 'n' eager user 😁

docs/cookbook/agent_with_local_llm.ipynb
Kostis-S-Z: Ugh yes, I accidentally removed this [here](https://github.com/mozilla-ai/any-agent/commit/810a0bbe89ca3c0494c656964d96fb44f592d5ec) when switching around frameworks. You will need to add this when defining the agent           

```
agent = AnyAgent.create(
    "smolagents",
    AgentConfig(
        model_id="ollama/granite3.3",
        instructions="""
        You must use the available tools to find an answer.
        """,
        tools=[mcp_filesystem, show_plan],
        model_args={"num_ctx": 10000, "tool_choice": "required"}},  # <--- Note the tools_choice arg
    ),
)

```

Or, simply, as you already did, change to `tinyagent`.

docs/cookbook/agent_with_local_llm.ipynb
Kostis-S-Z: > "This tutorial will guide on how to run agent, fully locally / offline i.e., running with a local LLM and a local MCP server  **_(inside a Docker container)_**, so no data will leave your machine!\n"


Would it work if I write it like that? I am also mentioning very explicitly below in the [Pick which tools to use ](https://github.com/mozilla-ai/any-agent/blob/f9ad388387ed7a02de518d0f3f9c4046e7771ae8/docs/cookbook/agent_with_local_llm.ipynb#L121)section:

> Before running the code below, make sure you have [Docker](https://docs.docker.com/get-started/) running in the background.

Kostis-S-Z:
After many evaluations with multiple models, the results have been incredibly inconsistent. I tried: 
- to remove the MCP server and just provide my own simple callable functions of reading/writing a file with just a simple .py file to read from
- completely changing the use-case to the example of another cookbook to search the web for the 5 best trending films
- the following models
  - `granite3.3:8`
  - `gemma3:12b`
  - `deepseek-r1:8b`
  - `deepseek-r1:14b`
  - `magistral:24b`
Even `deepseek-r1:70b` struggled a few times!?!

Its only the minority of the runs that any of these models actually managed to complete the task. 
I think we should discuss whether it makes sense to have such a "fragile" example in the cookbook. If anyone else has a better experience with local models, lets discuss!

Kostis-S-Z:
Update:
- I couldn't make it work with the local notebooks, nor without git cloning a whole repo. So I dont have a better alternative to simply checking in this `toy codebase` here.
- I swapped the mcp/filesystem for my own python callable functions as tools. this significantly increased the success rate on granite3.3

@ividal @daavoo

daavoo:
> Update:
> 
>     * I couldn't make it work with the local notebooks, nor without git cloning a whole repo. So I dont have a better alternative to simply checking in this `toy codebase` here.
> 
>     * I swapped the mcp/filesystem for my own python callable functions as tools. this significantly increased the success rate on granite3.3
> 
> 
> @ividal @daavoo

@Kostis-S-Z why don't you just pick a different use case instead of sticking to the toy_codebase? 

The purpose of the cookbook is to showcase how to set up local agents, and the current use case is not actually relevant because it doesn't work with realistic repositories.

daavoo (changes requested):
@Kostis-S-Z why don't you just pick a different use case instead of sticking to the toy_codebase? 

The purpose of the cookbook is to showcase how to set up local agents, and the current use case is not actually relevant because it doesn't work with realistic repositories.

Kostis-S-Z:
> @Kostis-S-Z why don't you just pick a different use case instead of sticking to the toy_codebase?

@daavoo I updated the use-case to simply reading a single file locally and writing a markdown file with potential vulnerabilities and/or bugs. Let me know what you think

daavoo (changes requested):
New usecase seems fine to me.

Still need to :

- Drop `toy_codebase` from the PR.
- Drop the `.pre-commit` config changes.

Kostis-S-Z:
> New usecase seems fine to me.
> 
> Still need to :
> 
>     * Drop `toy_codebase` from the PR.
> 
>     * Drop the `.pre-commit` config changes.
> 
>     * Run the cookbook and include example outputs

@daavoo Done in [7e5a649](https://github.com/mozilla-ai/any-agent/pull/643/commits/7e5a64945c4e2b9ea9d9dace8878b18b2d3ff1ca), [bb4432e](https://github.com/mozilla-ai/any-agent/pull/643/commits/bb4432e7d74fbd900ede86de0dc1ed199a6eacc0), [26868b5](https://github.com/mozilla-ai/any-agent/pull/643/commits/26868b5754c8acc3ef4e2c8f94dea5c10debd176)

daavoo (approved):

PR #647, opened 2025-07-17
chore(docs): Update tool usage checks.

daavoo:
We should showcase the iteration on `agent_trace.spans` so we are consistent with the API we use in the callbacks.

Also, the check based on messages can yield false positives because we are doing plain string matching on any assistant message.

For example, if the assistant returns something like:

```
I decided to not use the tool search_web because I can provide the answer without it
```

The current check will pass but the tool was not actually used.


daavoo:
Cookbook Tests: https://github.com/mozilla-ai/any-agent/actions/runs/16340749983

PR #666, opened 2025-07-24
fix(smolagents): Replace `client_session_timeout_seconds` with `conne…

daavoo:
…ct_timeout`.

You can see that the parameters are passed to `MCPAdapt`:

https://github.com/huggingface/smolagents/blob/b5f869604fe06843239936d56a816281ddcdb2be/src/smolagents/mcp_client.py#L108

Which doesn't have a `client_session_timeout_seconds`.

This was never working but the integration test was being skipped and in unit tests, well, we just assert our own assumptions.


daavoo:
Integration tests fail https://github.com/mozilla-ai/any-agent/actions/runs/16495024346/job/46638586394

njbrake:
Good catch, sorry about missing it 🙈

PR #675, opened 2025-07-25
chore(test_agent): Use smaller gemini and gpt models.

daavoo:


tests/integration/test_agent.py
```diff
@@ -160,7 +160,8 @@ class Steps(BaseModel):
 
 
 @pytest.mark.parametrize(
-    "model_id", ["gemini/gemini-2.5-pro", "openai/gpt-4.1-mini", DEFAULT_SMALL_MODEL_ID]
+    "model_id",
```
njbrake, line 163: Should we elevate the model ids to some sort of list or models stored in the testing helpers? This seems like a good thing to have access to in other places, idk.

PR #685, opened 2025-07-29
Add A2A streaming support

javiermtorres:
Enable streaming plus tool related events through the new `stream_tool_usage` parameter in the `A2AServingConfig` class.


javiermtorres:
The PR includes moving the fw-dependent information to the wrappers from the span generation, so all callbacks can find the same information in the same place. Some further refactoring to simplify certain parts is expected to happen later on.

daavoo (changes requested):
@javiermtorres I see 3 major things being mixed here:

- Refactor of wrappers/span_generation 
- Change of Callbacks API from sync to async
- A2A changes with a new callback
- A2A updates for a new release (I assume the renames in A2A args are this)

It would be nice to split each of those changes into separate PRs for easier discussion and review

javiermtorres:
Ok, let me restructure the PRs then. Would you mind starting comments with your concerns for the different changes here while I prepare the PRs? Thanks in advance.

javiermtorres:
> A2A updates for a new release (I assume the renames in A2A args are this)

Not exactly. I suspect we may not be flushing the cache whenever a package update is done:
https://github.com/mozilla-ai/any-agent/actions/runs/16652366445/workflow?pr=694#L27
so we end up reusing a stale mypy cache but the types for e.g. the new a2a sdk have changed. I'd either remove the mypy cache, or maybe store the uv.lock and fire an action to rebuild the cache whenever the uv.lock changes. Wdyt?

daavoo:
Closing in favor of https://github.com/mozilla-ai/any-agent/pull/728

PR #688, opened 2025-07-30
Use context events in sse starlette in integration tests

javiermtorres:
As a fix to https://github.com/sysid/sse-starlette/issues/140 context vars are used to store the "sse starlette" events, instead of using a single global. Since we don't use signal handlers in the uvicorns we manage in tests, we need to shut down sse_starlette manually, now using the appropriate context event.


javiermtorres:
Integration tests ok: https://github.com/mozilla-ai/any-agent/actions/runs/16624941085/job/47039323008
Fixing now lint issues...

PR #694, opened 2025-07-31
Use async callbacks

javiermtorres:
Some callbacks need to be async due to IO operations being handled in them. This PR makes most callback definition async, and updates existing callbacks appropriately.


javiermtorres:
Unit tests work ok. Integration in main is flaky: https://github.com/mozilla-ai/any-agent/actions/runs/16654377397

javiermtorres:
I need to include the type fixes in this PR not to break up things further.

javiermtorres:
Callback wrapping for the case of SmolAgents is not easy to solve within the scope of this PR. I will skip the `tests/integration/test_error_handling.py::test_runtime_error[SMOLAGENTS]` test for the moment and address this in a follow up PR.

javiermtorres:
I would weigh whether offloading to a thread pool is going to be an effective strategy since all llm related operations are now async (since they're IO based) and 6 out of 7 frameworks use async operations more or less by default (plus the a2a sdk). A callback writer can still write their callback in the same way; the task of making sure the result is awaited is on our code, after all. I'm also inspecting the use of `run_async_in_sync` since I'm not totally sure that exceptions that are not caught inside the sync part are really transferred across thread contexts (see https://github.com/mozilla-ai/any-agent/pull/694#issuecomment-3140898078).

Inspecting the callbacks is a strategy; but makes us work double. Unfortunately this is a strategy that many developers need to take (take for example the `AsyncCallbackHandler` vs `BaseCallbackHandler` in LangChain). Note that the current callbacks do not require changes except the change in signature, making the impact smaller.

I'll go for the wrappers taking one action or another depending on the type of callback, but this will make it a bit hard to check the type at compile time, defeating the point of type checks (eventually leading us to the same double interface in LangChain).

PR #695, opened 2025-07-31
Check precommit cache

javiermtorres:
The precommit cache may also keep obsolete mypy data. This will be rolled back if it does not affect type checks.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -24,12 +24,5 @@ jobs:
       - name: Install pre-commit
         run: uv sync --group lint
 
-      - uses: actions/cache@v4
-        with:
-          path: ~/.cache/pre-commit
-          key: ${{ runner.os }}-pre-commit-${{ hashFiles('.pre-commit-config.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-pre-commit-
-
       - name: pre-commit
         run: uv run pre-commit run --all-files --verbose


njbrake:
@javiermtorres wait, why was this done? I added caching via https://github.com/mozilla-ai/any-agent/pull/611 and I see that you and @daavoo  removed the caching in two newer PRs but its not clear to me why you did that instead of clearing the cache in the github settings if something was stale, or adjusting a command?

daavoo:
> @javiermtorres wait, why was this done? I added caching via #611 and I see that you and @daavoo removed the caching in two newer PRs but its not clear to me why you did that instead of clearing the cache in the github settings if something was stale, or adjusting a command?

mypy caching was based on the `pyproject.toml` hash, so we were having a mismatch between local envs and CI (because of pins like `>=0.X`). For example, an a2a minor release was causing type failures not caught by CI.

This cache removal, I am not sure why it was needed (as I assumed that mypy was the only problem). 

> instead of clearing the cache in the github settings if something was stale, or adjusting a command?

We shouldn't rely on manually clearing the github cache, we would need to clean it every time a dep releases a version that doesn't result in a pyproject.toml update.

We can probably figure out a better pattern for caching making sure that we don't encounter the situation described above and still benefit from caching but, thinking short-term, I decided to drop the cache because:

- It was the simplest solution 
- "correctness" felt more urgent than speed.
- The times of the full CI check are still upper bounded by integration tests.

njbrake:
@daavoo  @javiermtorres Thanks for the clarification. I'll create a Github issue to track putting the cache back in, to make sure that we address these concerns before a cache is re-enabled. 

Created https://github.com/mozilla-ai/any-agent/issues/704

PR #697, opened 2025-08-01
test: Fix dumping of `final_output`.

daavoo:
The dumped traces were not containing `final_output` when using `output_type` because it is a `BaseModel` instance. `serialize_as_any` needs to be passed to ensure the JSON contains `final_output`.

### DIFF
--- a/tests/snapshots/__snapshots__/test_trace.ambr
+++ b/tests/snapshots/__snapshots__/test_trace.ambr
@@ -1,7 +1,21 @@
 # serializer version: 1
-# name: test_agent_trace_snapshot[AGNO_trace].111709810402177282691297741314422280934
+# name: test_agent_trace_snapshot[AGNO_trace].95544105696794577450773972010078374733
   dict({
     'final_output': dict({
+      'steps': list([
+        dict({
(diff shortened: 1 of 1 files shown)


javiermtorres:
As a quick fix it seems ok, but I'd propose taking a look why the result is considered a BaseModel (maybe it needs to be typed as a `type[BaseModel]` and not just a `BaseModel`?).

tests/integration/test_agent.py
```diff
@@ -200,7 +200,7 @@ def write_file(text: str) -> None:
     if update_trace:
         trace_path = Path(__file__).parent.parent / "assets" / agent_framework.name
         with open(f"{trace_path}_trace.json", "w", encoding="utf-8") as f:
-            f.write(agent_trace.model_dump_json(indent=2))
+            f.write(agent_trace.model_dump_json(indent=2, serialize_as_any=True))
```
njbrake, line 203: Why is this the right place to put this fix? Won't everyone who tries to dump an agent_trace hit this same problem? This issue was why originally I had a custom serialization method in agent trace, maybe this is an indication that we should put this back.

PR #698, opened 2025-08-01
docs(tracing): Include and explain `serialize_as_any`

daavoo:


docs/tracing.md
```diff
@@ -109,9 +109,14 @@ agent = AnyAgent.create(
...
 with open("agent_trace.json", "w", encoding="utf-8") as f:
-  f.write(agent_trace.model_dump_json(indent=2))
+  f.write(agent_trace.model_dump_json(indent=2, serialize_as_any=True))
 ```
 
+!!! tip
+
+    Passing `serialize_as_any=True` makes sure that the `final_output` gets dumped
+    even when [`AgentConfig.output_type`][any_agent.AgentConfig.output_type] is used.
+
```
njbrake, line 119: I'd like to discuss this, my gut feeling is that we should take care of this for the developer, they shouldn't need to care about this.

PR #699, opened 2025-08-01
test(agent_judge): Update `model_id`.

daavoo:
Tests were not passing with `-nano`.

### DIFF
--- a/tests/integration/test_evaluation.py
+++ b/tests/integration/test_evaluation.py
@@ -26,7 +26,7 @@ def test_llm_judge(agent_trace: AgentTrace) -> None:
 
 def test_agent_judge(agent_trace: AgentTrace) -> None:
     agent_judge = AgentJudge(
-        model_id="openai/gpt-4.1-nano",
+        model_id="openai/gpt-4.1-mini",
         model_args=get_default_agent_model_args(AgentFramework.TINYAGENT),
     )
 


tests/integration/test_evaluation.py
```diff
@@ -26,7 +26,7 @@ def test_llm_judge(agent_trace: AgentTrace) -> None:
 
 def test_agent_judge(agent_trace: AgentTrace) -> None:
     agent_judge = AgentJudge(
-        model_id="openai/gpt-4.1-nano",
```
njbrake, line 29: Why are we back on using openai/gpt-4.1 and not mistral or https://github.com/mozilla-ai/any-agent/blob/main/src/any_agent/testing/helpers.py#L12 ?

daavoo: Because mistral was causing more flaky failures (sometimes because the model is just worse and sometimes because the API is less reliable).

I understand the value of promoting mistral in the docs/examples due to it being open weights, but for our internal tests I consider mistral just another item in the matrix of providers we want to test and would prefer to use whatever model/provider is more reliable for each test (happy to keep using mistral as default in those tests that work consistently)

njbrake: Good with me. Can you put some comment in the code explaining this?

PR #701, opened 2025-08-01
chore: Split `test-integration` into separate parallel jobs.

daavoo:
Instead of using pytest `-n auto`, as that  was a source of different flaky tests.


javiermtorres:
A specially relevant doc section is https://pytest-xdist.readthedocs.io/en/stable/how-to.html#making-session-scoped-fixtures-execute-only-once

njbrake:
Hmm. I'm not totally convinced that this is a good change. Most of the time is installing dependencies, which is now duplicated multiple times. pytest-xdist was already splitting out the tests across processes to parallelize things.

What problem is this PR trying to solve @daavoo ?

daavoo:
> Most of the time is installing dependencies, which is now duplicated multiple times. 

It is not exactly duplicated, as each job installs different group of deps (for example, tinyagent and mcp only install the "barebones" lib).
Anyhow, the installation is done in parallel for each job, so the total runtime of the suite doesn't increase.

> pytest-xdist was already splitting out the tests across processes to parallelize things.

Tests actually take less time with this approach:

https://github.com/mozilla-ai/any-agent/actions/runs/16675055661

Compared to the previous single-job xdist:

https://github.com/mozilla-ai/any-agent/actions/runs/16624941085

> What problem is this PR trying to solve @daavoo ?

My hypothesis/motivations were:

- xdist was actually making things slower than parallelizing at the job level (you can see the results int he link above). xdist is still used in some jobs.
- xdist the whole suite was causing flakyness as a lot of stuff is going on within a single machine (a2a/mcp servers are being spinned up, concurrent async jobs launched, API calls to remote server) making it hard to understand the failures. We still do a lot of stuff still inside each job, but at least the variety of stuff is reduced.
- splitting into jobs allows for detecting what failed quicker at a higher level (is it `a2a`, is it everything?)

njbrake:
@daavoo  Works for me, thank you for the clarification 🙏

PR #705, opened 2025-08-04
fix(a2a): Make instructions of `_send_query` more clear.

daavoo:
The cause of the failures was the agent deciding to pass `"'0"` as task_id instead of `None`.

Screenshot 2025-08-04 132218

Integration tests:

https://github.com/mozilla-ai/any-agent/actions/runs/16721786710/job/47327664062

### DIFF
--- a/src/any_agent/tools/a2a.py
+++ b/src/any_agent/tools/a2a.py
@@ -156,10 +156,11 @@ async def a2a_tool_async(
 
         Args:
             query (str): The query to send to the agent.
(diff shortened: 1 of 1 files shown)


daavoo:
> The cause of the failures was the agent deciding to pass "'0" as task_id instead of None.

Although the descriptions could indeed be improved, ultimately this was a "model" issue where `mistral-small` was failing to understand the correct arguments for the function whereas gpt-4.1-mini was working as expected.

PR #709, opened 2025-08-04
tests: Enable xai integration tests.

daavoo:
https://github.com/mozilla-ai/any-agent/actions/runs/16726796358/job/47344465955

image


njbrake:
@daavoo  is it worth adding a test that checks this with tinyagent + any_llm enabled?

PR #710, opened 2025-08-05
feat(tiny_agent): Update instructions.

daavoo:
Focus more on tool calling and `final_answer` behavior.

The idea is to help smaller language models.

### DIFF
--- a/src/any_agent/frameworks/tinyagent.py
+++ b/src/any_agent/frameworks/tinyagent.py
@@ -24,11 +24,12 @@ if TYPE_CHECKING:
 
 
 DEFAULT_SYSTEM_PROMPT = """
-You are an agent - please keep going until the user's query is completely resolved, before ending your turn and yielding back to the user. Only terminate your turn when you are sure that the problem is solved, or if you need more info from the user to solve the problem.
+You are an agent that uses tools (whenever they are available) to answer the user's query.
 
-If you are not sure about anything pertaining to the user's request, use your tools to read files and gather the relevant information: do NOT guess or make up an answer.
+You will keep calling tools until you find a final answer.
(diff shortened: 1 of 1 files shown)


daavoo:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/16745970665

src/any_agent/frameworks/tinyagent.py
```diff
@@ -24,11 +24,12 @@
 
 
 DEFAULT_SYSTEM_PROMPT = """
-You are an agent - please keep going until the user's query is completely resolved, before ending your turn and yielding back to the user. Only terminate your turn when you are sure that the problem is solved, or if you need more info from the user to solve the problem.
+You are an agent that uses tools (whenever they are available) to answer the user's query.
```
njbrake, line 27: Why the "whenever they are available" term? When are the tools not available? I'm all for PRs that change or improve a system prompt, but I think it would be useful to provide some small explanation of what was going wrong and how this solved it.

daavoo: > Why the "whenever they are available" term? When are the tools not available?

We allow the user to don't pass any tools, so that is the scenario (we actually have some tests without tools).

njbrake: But I thought that we refactored it so there will always be the final answer tool? I see that it's only added if tool_choice = "required" so maybe my misunderstanding is that we don't always require the final_answer tool to be used.

daavoo: > But I thought that we refactored it so there will always be the final answer tool?

That is true. I see now why this can be confusing. I was meaning the "user tools". Will drop that

I would love to have a more informed way to have this prompt changes but for now 🤷

PR #711, opened 2025-08-05
feat(agent_with_local_llm): New use case, new model (`Qwen/Qwen3-1.7B`) and new framework (`tinyagent`).

daavoo:
- Update the cookbook to use a model hosted in `HF_ENDPOINT` when running on CI.

- Update `tests-cookbook` with `HF_TOKEN` and `HF_ENDPOINT` secrets.

- Drop skip from `test_cookbooks`.


njbrake:
This is one way to do it. Another option could be to run ollama in the github action, see how I do it in any-llm.

https://github.com/mozilla-ai/any-llm/blob/main/.github/workflows/tests-integration.yaml#L31-L43

The nice thing about that approach is that then you don't need to put any test code into the cookbook, which may very slightly distract the people using the cookbook. Not a big deal though.

docs/cookbook/agent_with_local_llm.ipynb
```diff
@@ -195,31 +161,41 @@
...
    "source": [
+    "import os\n",
+    "\n",
     "from any_agent import AgentConfig, AnyAgent\n",
     "\n",
-    "# Define the agent\n",
+    "if os.getenv(\"IN_PYTEST\") == \"1\":\n",
+    "    # This is a Qwen/Qwen3-1.7B hosted in a HuggingFace Endpoint\n",
+    "    model_id = \"huggingface/tgi\"\n",
+    "    model_args = {\"api_base\": os.environ[\"HF_ENDPOINT\"]}\n",
```
njbrake: This might benefit from a better comment. People will read this and have no idea what's going on, best case scenario they realize that IN_PYTEST doesn't apply to them, but there's no explanation about why huggingface/tgi maps to Qwen. It might be better to omit the stuff about Qwen for a more generic comment like "this is an endpoint we use for our automated testing" or something like that, idk.

njbrake (approved):

daavoo:
> Another option could be to run ollama in the github action, see how I do it in any-llm.
> 
> https://github.com/mozilla-ai/any-llm/blob/main/.github/workflows/tests-integration.yaml#L31-L43

I saw (and tried) that. 
The issue here is that we require a little bigger model and more "complex" queries, which need more resources than the default github runner provides. 

We could do stuff like setting up a larger github runner and only run cookbooks there, etc.

In the end, what I felt that is valuable testing is that `any-agent`&`tinyagent` can do "agentic" things with a small language model (<8B). 

Bringing ollama in the background (vs a hosted small model in remote API) only adds an additional layer of complexity that we don't care here (because any-agent doesn't take care of the communication with the model, it is the job of litellm / any-llm).

I do believe serving an actual ollama in CI makes sense for any-llm (because it is responsible for integrating with ollama)

tests/cookbooks/test_cookbooks.py
```diff
@@ -34,6 +32,8 @@ def test_cookbook_notebook(
                 "MISTRAL_API_KEY": os.environ["MISTRAL_API_KEY"],
                 "TAVILY_API_KEY": os.environ["TAVILY_API_KEY"],
                 "OPENAI_API_KEY": os.environ["OPENAI_API_KEY"],
+                "HF_TOKEN": os.environ["HF_TOKEN"],
+                "HF_ENDPOINT": os.environ["HF_ENDPOINT"],
                 "PATH": os.environ["PATH"],
                 "IN_PYTEST": "1",  # For mcp_agent notebook which needs to mock input
```
njbrake: This now also applies to the agent_with_local_llm notebook fwiw

daavoo:
Successful execution can be found in:

https://github.com/mozilla-ai/any-agent/actions/runs/16746204608

Executed notebook can be found in the artifacts

docs/cookbook/agent_with_local_llm.ipynb
daavoo: I updated the comment with more info but I didn't `omit the stuff about Qwen`. I think is valuable for me or others to understand to which kind of model the endpoint points to (type/size of the model, etc)

njbrake:
> > Another option could be to run ollama in the github action, see how I do it in any-llm.
> > https://github.com/mozilla-ai/any-llm/blob/main/.github/workflows/tests-integration.yaml#L31-L43
> 
> I saw (and tried) that. The issue here is that we require a little bigger model and more "complex" queries, which need more resources than the default github runner provides.
> 
> We could do stuff like setting up a larger github runner and only run cookbooks there, etc.
> 
> In the end, what I felt that is valuable testing is that `any-agent`&`tinyagent` can do "agentic" things with a small language model (<8B).
> 
> Bringing ollama in the background (vs a hosted small model in remote API) only adds an additional layer of complexity that we don't care here (because any-agent doesn't take care of the communication with the model, it is the job of litellm / any-llm).
> 
> I do believe serving an actual ollama in CI makes sense for any-llm (because it it actually it's job to integrate with ollama)

Good with me, thanks 🙏

tests/cookbooks/test_cookbooks.py
daavoo: updated

docs/cookbook/agent_with_local_llm.ipynb
njbrake: Good with me, so long as whenever the endpoint is updated this comment is updated too: it risks going out of date since we don't have visibility into what the HF_ENDPOINT is pointing to (other than that it's a Github Secret that points to something). Of course we can look at the agent trace to confirm during execution.

PR #713, opened 2025-08-05
chore: Bring back lint cache.

daavoo:
Include `run-mypy` fix from https://github.com/mozilla-ai/any-llm/commit/fdd672508de5271c0266f08a992e4efb4192d3af

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -24,5 +24,19 @@ jobs:
       - name: Install pre-commit
         run: uv sync --group lint
 
+      - uses: actions/cache@v4
+        with:
+          path: .mypy_cache
+          key: ${{ runner.os }}-mypy-${{ hashFiles('pyproject.toml') }}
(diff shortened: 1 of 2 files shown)


javiermtorres:
The main issue I see here is that, although the cache will save time, it could potentially be much less than expected due to -U on heavy packages being upgraded. Also, CI jobs will not be repeatable, depending instead on what the latest view would be at the time the job is run, making us doubt if a CI run is ok or not for a new PR because some package may have been updated during the PR.

I'd rather depend on some cached uv.lock file, plus some specific action triggered on maybe dependabot PRs to update it. Since this file would be cached and not in git, the user would not be constrained when it comes to package versions. Wdyt?

We can of course use this in the meantime, it may at least save a significant amount of CI build time.

daavoo:
@njbrake I think I left this open because it was showcasing some issue with the cache and I wanted to use it for debugging (but didn't really find the time)

PR #715, opened 2025-08-05
tests: Enable `Qwen/Qwen3-1.7B` integration test.

daavoo:
This should serve as baseline to confirm that `tinyagent` + `small LM` can perform agentic tasks

### DIFF
--- a/tests/integration/frameworks/test_agent.py
+++ b/tests/integration/frameworks/test_agent.py
@@ -113,6 +113,7 @@ class Steps(BaseModel):
     [
         "anthropic/claude-3-5-haiku-latest",
         "gemini/gemini-2.5-flash",
+        "huggingface/tgi",  # This is a Qwen/Qwen3-1.7B hosted in https://endpoints.huggingface.co/mozilla-ai/endpoints/dedicated
         "openai/gpt-4.1-nano",
         "xai/grok-3-mini-latest",
         DEFAULT_SMALL_MODEL_ID,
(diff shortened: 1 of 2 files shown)


javiermtorres (approved):

javiermtorres:
Btw, as mentioned some time before, maybe we could make a suite of "not-necessary-tests" to indicate the status of this kind of feature (which may not be essential but valuable to track nonetheless), istead of un/skipping tests. Wdyt?

tests/integration/frameworks/test_agent.py
```diff
@@ -113,6 +113,7 @@ class Steps(BaseModel):
     [
         "anthropic/claude-3-5-haiku-latest",
         "gemini/gemini-2.5-flash",
+        "huggingface/tgi",  # This is a Qwen/Qwen3-1.7B hosted in https://endpoints.huggingface.co/mozilla-ai/endpoints/dedicated
```
njbrake, line 116: Is there some way to make this a specific endpoint name? My worry is that we have no way of knowing if it is no longer Qwen3-1.7B

njbrake (approved):

tests/integration/frameworks/test_agent.py
daavoo: I didn't really find a way neither on HF side nor litellm https://docs.litellm.ai/docs/providers/huggingface#dedicated-inference-endpoints 🤷 
Showing the `api_base` URL also doesn't help because it is an arbitrary hash

daavoo:
https://github.com/mozilla-ai/any-agent/actions/runs/16749840330/job/47417507505#step:5:30

image

PR #717, opened 2025-08-07
fix(serving): Don't require `a2a` when  using `mcp`.

daavoo:


daavoo:
MCP integration tests passing without `a2a` dep installed: 

https://github.com/mozilla-ai/any-agent/actions/runs/16799974984/job/47578887439


Failure when accessing the methods instead of importing:

image

PR #720, opened 2025-08-07
feat(agno): Replace default LiteLLM with AnyLLM for `DEFAULT_MODEL_TY…

daavoo:
…PE`.

### DIFF
--- a/src/any_agent/frameworks/agno.py
+++ b/src/any_agent/frameworks/agno.py
@@ -1,6 +1,7 @@
-from typing import TYPE_CHECKING, Any
+from __future__ import annotations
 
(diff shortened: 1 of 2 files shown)


src/any_agent/frameworks/agno.py
```diff
@@ -1,25 +1,236 @@
...
 
 if TYPE_CHECKING:
     from agno.agent import RunResponse
-    from agno.models.base import Model
+    from agno.models.message import Message
+    from pydantic import BaseModel
+
+
+@dataclass
+class AnyLLM(Model):
```
daavoo, line 30: This should be contributed upstream to Agno

PR #731, opened 2025-08-14
fix(tinyagent): Drop `max_turns` and `DEFAULT_MAX_TURNS`.

daavoo:
This was intended to be removed in 1.0 in favor of using callbacks as documented in https://mozilla-ai.github.io/any-agent/agents/callbacks/#example-limit-the-number-of-steps .

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -167,7 +167,7 @@ def mock_litellm_response() -> ModelResponse:
 def mock_litellm_tool_call_response() -> ModelResponse:
     """Fixture to create a mock LiteLLM response that includes tool calls"""
     return ModelResponse.model_validate_json(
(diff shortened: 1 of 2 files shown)


njbrake:
Does the demo need to be updated too? https://github.com/mozilla-ai/any-agent/blob/2b5986822c0d49fb2b3895e14a4cc625f7e9cb06/demo/services/agent.py#L170

daavoo:
> Does the demo need to be updated too?
> 
> https://github.com/mozilla-ai/any-agent/blob/2b5986822c0d49fb2b3895e14a4cc625f7e9cb06/demo/services/agent.py#L170

Yes but not as part of this PR, the demo is fully outdated (and pinned to older any-agent version), tracked in https://github.com/mozilla-ai/any-agent/issues/682

PR #739, opened 2025-08-18
feat: Refactor and Simplify MCP

njbrake:
Introduced `openai==1.99.6` dependency in `pyproject.toml`. (apparently some new code broke Litellm, openai may push a fix soon)

Now we just have a single MCP class that everyone uses, we convert the MCPs to callables and pass that to each agent framework, saves us about 1,200 lines of code 🎉 . It works for everybody but Smolagents, because they try to re-create callables to turn them into classes, so I may need to keep the original MCP logic for them, if I can't figure out a better way to make them happy about the MCP as a callable


pyproject.toml
```diff
@@ -12,6 +12,7 @@ dependencies = [
   "any-llm-sdk>=0.14.1,<1",
   "openai<1.100",
   "litellm>=1.74.0",
+  "openai==1.99.6",
```
daavoo: It's already pinned above.

```suggestion
```

src/any_agent/tools/mcp/mcp_client.py
```diff
@@ -0,0 +1,243 @@
...
+# Check for MCP dependencies
+mcp_available = False
+with suppress(ImportError):
+    from mcp import ClientSession, StdioServerParameters
+    from mcp.client.sse import sse_client
+    from mcp.client.stdio import stdio_client
+    from mcp.client.streamable_http import streamablehttp_client
+    from mcp.types import Tool as MCPTool
+
+    mcp_available = True
```
daavoo: I know we were using this pattern before, but I've found it problematic as it hides the traceback of the original ImportError, which is useful when weird things caused it.

I would prefer if we start using something like:

```suggestion
missing_mcp_error = None
try:
    from mcp import ClientSession, StdioServerParameters
    from mcp.client.sse import sse_client
    from mcp.client.stdio import stdio_client
    from mcp.client.streamable_http import streamablehttp_client
    from mcp.types import Tool as MCPTool
except ImportError as e:
    missing_mcp_error = e
```

And then when raising the exception we can do `raise ... from missing_mcp_error` to show the original traceback to the user.

src/any_agent/tools/mcp/mcp_client.py
```diff
@@ -0,0 +1,243 @@
...
+        )
+        client_session = ClientSession(read, write, timeout)
+        self._session = await self._exit_stack.enter_async_context(client_session)
+        await self._session.initialize()
+
+    def get_session_id(self) -> str | None:
+        """Get session ID for streamable HTTP connections."""
+        if self._get_session_id_callback:
+            return self._get_session_id_callback()
+        return None
```
daavoo: How/Where is this used?

src/any_agent/tools/mcp/mcp_client.py
```diff
@@ -0,0 +1,243 @@
...
+        # Set function metadata
+        mcp_tool_function.__name__ = name
+        mcp_tool_function.__doc__ = enhanced_description
+        mcp_tool_function.__signature__ = signature  # type: ignore[attr-defined]
+        mcp_tool_function.__annotations__ = {**annotations, "return": str}
+        mcp_tool_function.__input_schema__ = input_schema  # type: ignore[attr-defined]
+
+        return mcp_tool_function
+
+    def _schema_type_to_python_type(self, schema_type: str) -> type:
```
daavoo: 💅 This could be probably just a TYPE_MAPPING constant or something

tests/unit/tools/mcp/test_unit_all_mcp.py
```diff
@@ -5,70 +5,115 @@
...
+    mcp_client.config = mcp_client.config.model_copy(
         update={"tools": ["tool1", "nonexistent"]}
     )
     with pytest.raises(ValueError) as excinfo:  # noqa: PT011
-        mcp_connection._filter_tools(all_tools)
+        mcp_client._filter_tools(all_tools)
     assert "Missing: ['nonexistent']" in str(excinfo.value)
+
+
+def test_mcp_client_tool_conversion(
```
daavoo, line 89: not a blocker: I guess we could also test the "enhanced description" we create and stuff like that

src/any_agent/tools/mcp/mcp_client.py
njbrake: Helper function for the session id that we get back from streamable http 🤷  I'll remove it atm because you're right we don't actually use it anywhere yet

njbrake:
> High-level the change makes sense to me and it looks great 🔥
> 
> My worry is that we don't have tests with multiple MCP servers, afaik. I would feel more safe if we could cover those scenarios.

Do you think I should edit the test_agent test to add a second MCP server? I did this locally just to make sure it worked ok (which it did)

PR #740, opened 2025-08-18
Update pyproject.toml

njbrake:
Until https://github.com/openai/openai-python/issues/2564 is resolved


njbrake:
They patched it in 1.100.1

PR #743, opened 2025-08-20
Use absolute URL for logo in README

dpoulopoulos:
Update the image logo in the README to use an absolute URL pointing to the raw GitHub content. This change ensures that the image is displayed correctly on external sites that render the README, such as PyPI.


stefanfrench (approved):
LGTM - thanks!

PR #746, opened 2025-08-20
refactor(tinyagent): Updates to better handle `output_type`.

daavoo:
- Drop `_return_output_type`
- Implement generic logic using `FINAL_ANSWER_TOOL` where the description includes output_type schema (if present).
- Implement openai-specific logic using `response_format`.


src/any_agent/frameworks/tinyagent.py
```diff
@@ -105,12 +102,17 @@ def __init__(self, config: AgentConfig) -> None:
         self.completion_params = {
             "model": self.config.model_id,
             "tools": [],
-            "tool_choice": "required",
             **(self.config.model_args or {}),
         }
 
-        if self.completion_params["tool_choice"] == "required":
-            self.config.tools.append(final_answer)
+        uses_openai = "openai" in self.config.model_id
```
njbrake: What if instead, we use the split_model_provider function from any-llm? https://github.com/mozilla-ai/any-llm/blob/main/src/any_llm/provider.py#L353

I'm worried that maybe someone will name their model with 'openai' instead of just the provider, now that openai has an Open-weight model out here.

src/any_agent/frameworks/tinyagent.py
```diff
@@ -105,12 +102,17 @@ def __init__(self, config: AgentConfig) -> None:
...
+
+        user_tool_choice = self.completion_params.get("tool_choice")
+        if user_tool_choice and user_tool_choice != "required" and not uses_openai:
+            logger.warning(
+                "TinyAgent framework only works with tool_choice=required."
+                f"Provided option will be ignored: {user_tool_choice}"
+            )
```
njbrake, line 129: Should we throw an exception instead of logging a warning?

src/any_agent/frameworks/tinyagent.py
```diff
@@ -158,20 +159,56 @@ async def _load_agent(self) -> None:
...
 
+            input_schema["additionalProperties"] = False  # type: ignore[assignment]
             self.completion_params["tools"].append(
                 {
                     "type": "function",
                     "function": {
                         "name": tool_name,
                         "description": tool_desc,
                         "parameters": input_schema,
+                        "strict": True,
```
daavoo: openai requires this and the `additionalProperties = False` when using tool calls + response_format . The rest of providers don't seem to be bothered by it.

daavoo:
Closing in favor of https://github.com/mozilla-ai/any-agent/pull/747 .

Decided to only implement the openai changes to unblock https://github.com/mozilla-ai/agent-factory/issues/283 .

Opened a separate issue https://github.com/mozilla-ai/any-agent/issues/748 to discuss the overall `A2A` streaming setup (which was the main conflict trying to be solved with this changes) .

PR #749, opened 2025-08-21
chore(docs+deps): Update any-llm references

Kostis-S-Z:
pyproject.toml
- Remove any-llm dep group (since any-llm already a core dep)

docs/agents/frameworks/tinyagent.md
- Remove outdate ref to experimental any-llm support

README.md
- Update ref of smolagents from fan-made website to official GH repo
- Update agno ref to their GH repo + add a shield


njbrake:
@Kostis-S-Z  Thanks for the updates! Totally agree, we can drop the `### Experimental any-llm Support` content. But lets keep the rest as is: soon any-llm will be the default everywhere so there won't be a need to reference it explicitly with tinyagent.

njbrake (approved):
LGTM. Thank you! 🚀

PR #750, opened 2025-08-21
feat(openai): Set default max_turns to infinity in run method

njbrake:
they don't allow passing in None, so we set it inf as an alternative


njbrake:
@daavoo I may misunderstand your comment. The problem i was encountering was that if you don't hardcode anything on our side, OpenAI has a hardcoded default of 10.

PR #753, opened 2025-08-22
fix(tinyagent): Use `any_llm.acompletion` instead of `completion`.

daavoo:
Switching to `completion` meant that there was not a single `await` happening inside tinyagent `_run_async`, causing it to effectively block the async loop until the agent finished.

This was the root issue behind https://github.com/mozilla-ai/agent-factory/pull/292#issuecomment-3213697688


daavoo:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/17158421372

PR #758, opened 2025-08-24
Fix A2A streaming callback accumulation

### LINKED ISSUE #729: When using A2A streaming, the tool calls appear repeatedly

anivar:
## What's happening?

When using A2A streaming with the OpenAI agent, tool calls were showing up multiple times - like an echo that gets louder with each request. 

## The culprit

The streaming callback was being added to the agent's callback list but never removed properly. Each request added another callback, causing the accumulation.

## The fix

Instead of modifying the shared callback list directly, we now:
1. Make a copy of the callbacks
2. Add our streaming callback to the copy
3. Use the modified copy for this request only
4. Always restore the original list when done

This way, each request starts fresh without any leftover callbacks from previous requests.

## How I tested it

Added tests that simulate multiple requests and verify:
- The callback count stays constant (no accumulation)
- Everything works even if errors happen mid-request
- The original callbacks are preserved

The fix is minimal - just 13 lines changed in the executor, using standard Python patterns already common in the codebase.

Fixes #729

### DIFF
--- a/src/any_agent/serving/a2a/agent_executor.py
+++ b/src/any_agent/serving/a2a/agent_executor.py
@@ -169,7 +169,12 @@ class AnyAgentExecutor(AgentExecutor):
         )
 
         # Put a callback here to record tool calling and send updates
+        # Store original callbacks to restore later
+        original_callbacks = None
         if self.stream_tool_usage:
+            original_callbacks = self.agent.config.callbacks
+            # Create a copy to avoid mutating the shared list
+            self.agent.config.callbacks = original_callbacks.copy()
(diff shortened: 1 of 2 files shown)


njbrake:
Hi @anivar , there is already a PR in progress for this issue. See https://github.com/mozilla-ai/any-agent/pull/730

PR #759, opened 2025-08-24
Improve MCP client lazy initialization pattern

### LINKED ISSUE #605: Move MCP client init code to a constructor
Currently some init code for MCP clients is run in the `list_tools` method, for example:
https://github.com/mozilla-ai/any-agent/blob/114e51b4ec76142144fee93a1d09766cbc822dd2/src/any_agent/tools/mcp/frameworks/agno.py#L82
MCP clients and connection should have proper constructors where the session is prepared, regardless of whether one or another MCP procedure is called.

anivar:
## The problem

Issue #605 pointed out that MCP client initialization was happening in `list_tools` instead of having "proper constructors where the session is prepared."

The issue was right - but moving initialization to the constructor would break things since Python constructors can't be async while MCP connections need async operations.

## What I found

The codebase actually already had a lazy connection pattern through `_ensure_connected()`, but it wasn't consistent enough. This created confusion about when connections actually happen.

## The fix

Instead of forcing sync constructors to handle async connections, I made the existing lazy pattern more robust:

- **Clear connection tracking**: Added a `_connected` flag so we know the state
- **Idempotent connections**: Calling `connect()` multiple times is safe 
- **Automatic connection**: Any operation needing the session will connect first
- **Consistent behavior**: Both regular and Smolagents clients work the same way
- **Cleaner tool loading**: Removed the explicit `connect()` call since it happens automatically

## Why this approach works better

- **Async-friendly**: No trying to jam async operations into sync constructors
- **Fail-fast where it matters**: Connection errors happen when you actually try to use the client
- **Resource efficient**: Don't connect unless you're going to use it
- **Easy to test**: Mock connection behavior without complex constructor logic
- **Framework agnostic**: Each framework can decide when it needs connections

## Backward compatibility

All existing code continues to work:
- If you call `connect()` explicitly, it still works (just becomes a no-op if already connected)
- If you just call `list_tools()`, it connects automatically like before
- No breaking changes anywhere

This addresses the concern in #605 while keeping the code maintainable and async-friendly.

Fixes #605

### DIFF
--- a/src/any_agent/tools/mcp/mcp_client.py
+++ b/src/any_agent/tools/mcp/mcp_client.py
@@ -41,6 +41,7 @@ class MCPClient(BaseModel):
     _get_session_id_callback: Callable[[], str | None] | None = PrivateAttr(
         default=None
     )
+    _connected: bool = PrivateAttr(default=False)
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
@@ -50,8 +51,16 @@ class MCPClient(BaseModel):
             msg = "You need to `pip install 'any-agent[mcp]'` to use MCP."
             raise ImportError(msg) from missing_mcp_error
 
+    async def _ensure_connected(self) -> None:
+        """Ensure the client is connected, connecting if necessary."""
+        if not self._connected:
+            await self.connect()
+
     async def connect(self) -> None:
         """Connect using the appropriate transport type."""
+        if self._connected:
+            return
+        
         if isinstance(self.config, MCPStdio):
             server_params = StdioServerParameters(
                 command=self.config.command,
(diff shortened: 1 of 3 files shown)


njbrake:
Hi @anivar  thank you so much for  your work on this! We love the enthusiasm and are excited at your contributions. However, we're a small team and have limited bandwith to manage contributions. We're not quite ready to review and implement additional MCP/ACP/A2A functionality.

We're working on continuing to update our [Contributing.md](https://github.com/mozilla-ai/any-agent/blob/main/CONTRIBUTING.md) to help clarify how we can accept contributions. 

We are really thankful for your effort and contributions, but we need to take a step back and coordinate better to make sure we are focused on the issues that we will have time to properly review and collaborate on. Thanks for your understanding 🙏

Issue #607, opened 2025-07-09
Past custom context into run command


njbrake:
Should we add an extra perimeter to our run commands to allow users to pass in custom context objects?

I'm thinking about a scenario where I may want a call back to update something that doesn't exist until runtime.

One example of this would be the a2a task, updater object, which isn't created until runtime, but a callback may want to use to selectively update the state

javiermtorres:
In that case, I'd wonder if a "broader" interface would be beneficial. Right now, the interface for tool calling is fairly clear and does not need much extensibility: the LLM will ask for tools as the result of a single round, the result will be passed together with the original context, and the LLM will elaborate again.

Then the result of the LLM will be appended to the history. One can then choose to update some own context (typed like `AnyAgent[UserContext]`, which defines not only the context type member, but also probably a `Callable[[UserContext, Message], UserContext]` method). This context could further take over or extend the history, allowing the agent designer to do such things as backtracking (by removing the latest messages according to some quality metric) or what you are describing here.

My concern is that the defined interface would be too specific for the a2a sdk concerns, where we can instead define a broader interface that allows more possibilities later on.

anivar:
I recently fixed a callback accumulation bug in A2A streaming and can see how this would be useful. The A2A executor creates runtime objects like TaskUpdater that callbacks would benefit from accessing.

Here are a few ways this could work:

**Option 1: Add context parameter to run methods**
```python
await agent.run_async("my query", context={"updater": task_updater, "session": user_session})
```

**Option 2: Extend the existing callback Context object**  
The Context class already has a `shared` dict field. We could add a `runtime_context` field that gets populated when the agent runs. This would be straightforward to implement.

**Option 3: Agent-level context setting**
```python
agent.set_runtime_context({"updater": updater})
# This gets passed to all callbacks automatically
```

Option 2 seems most natural since the Context object already flows through all callback methods and has the infrastructure for shared data.

Happy to implement this if we can agree on the approach!

anivar:
I went ahead and implemented this in PR #760. Took the approach we discussed (extending the Context object) since it seemed like the most natural fit.

The implementation is pretty straightforward - just added a `runtime_context` parameter that flows through to callbacks via the existing Context object. Should handle the A2A task updater use case nicely.

Would be great to get some feedback on whether this covers what you had in mind!

PR #760, opened 2025-08-24
Add runtime context parameter for callbacks

anivar:
## My take on the discussion in #607

The issue raised a good point about callbacks needing access to runtime objects like A2A task updaters. I took a shot at implementing this with what seemed like the most straightforward approach.

## What I built

Added a `runtime_context` parameter to run methods that gets passed through to callbacks:

```python
# When running the agent, pass runtime objects
runtime_context = {"task_updater": updater, "user_session": session}
await agent.run_async("query", runtime_context=runtime_context)
```

Callbacks can then grab what they need:

```python
def before_agent_invocation(self, context, prompt, **kwargs):
    if context.runtime_context and "task_updater" in context.runtime_context:
        updater = context.runtime_context["task_updater"]
        updater.update_status("started", "Processing query...")
    return context
```

## Implementation approach

- Added an optional `runtime_context` field to the existing Context object
- Made it a keyword-only parameter so it doesn't mess with existing code
- Keeps it simple - just a dictionary that flows through the callback system
- Backward compatible since it defaults to None

## Why this approach

Seemed like the cleanest way to solve the A2A task updater problem without over-engineering it. The Context object already flows through all callbacks, so extending it felt natural.

Happy to adjust based on feedback\!

Addresses #607

### DIFF
--- a/src/any_agent/callbacks/context.py
+++ b/src/any_agent/callbacks/context.py
@@ -31,3 +31,10 @@ class Context:
 
     shared: dict[str, Any]
     """Can be used to store arbitrary information for sharing across callbacks."""
+
+    runtime_context: dict[str, Any] | None = None
+    """Runtime-specific objects passed when the agent runs.
+    
+    This allows callbacks to access objects that only exist at runtime,
+    such as task updaters, user sessions, or other request-specific data.
+    """
--- a/src/any_agent/frameworks/any_agent.py
+++ b/src/any_agent/frameworks/any_agent.py
@@ -162,24 +162,30 @@ class AnyAgent(ABC):
         self._mcp_clients.extend(mcp_clients)
         return tools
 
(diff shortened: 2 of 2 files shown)


njbrake:
Hi @anivar  thank you so much for  your work on this! We love the enthusiasm and are excited at your contributions. However, we're a small team and have limited bandwith to manage contributions. We're not quite ready to review and implement additional runtime context functionality at the moment.

We're working on continuing to update our [Contributing.md](https://github.com/mozilla-ai/any-agent/blob/main/CONTRIBUTING.md) to help clarify how we can accept contributions. 

We are really thankful for your effort and contributions, but we need to take a step back and coordinate better to make sure we are focused on the issues that we will have time to properly review and collaborate on. Thanks for your understanding 🙏

PR #779, opened 2025-08-28
Implement safe argument casting in tinyagent.py

njbrake:
Added a safe casting function to handle argument types, including union types.

Found when working in any-forge, We had some bool | None types that were silently being ignored

### DIFF
--- a/src/any_agent/frameworks/tinyagent.py
+++ b/src/any_agent/frameworks/tinyagent.py
@@ -3,9 +3,8 @@ from __future__ import annotations
 import asyncio
 import inspect
 import json
-import types
 from contextlib import suppress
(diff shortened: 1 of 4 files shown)


src/any_agent/frameworks/tinyagent.py
```diff
@@ -62,7 +63,9 @@ async def call_tool(self, request: dict[str, Any]) -> str:
                 for arg_name, arg_type in func_args.items():
                     if arg_name in arguments:
                         with suppress(Exception):
```
daavoo, line 65: didn't notice this before @njbrake but we should probably not use `supress` here but rather catch and log a warning or something

PR #793, opened 2025-09-03
feat: Support `composio` via custom `CallableProvider`.

daavoo:
Generic provider that can be used with any of the frameworks.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,7 @@ composio = [
 ]
 
 all = [
-  "any-agent[agno,google,langchain,llama_index,openai,smolagents,composio]",
+  "any-agent[agno,google,langchain,llama_index,openai,smolagents]",
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,8 +54,12 @@ a2a = [
   "httpx",
 ]
 
+composio = [
+  "composio>=0.8.9,<1"
+]
+
 all = [
-  "any-agent[agno,google,langchain,llama_index,openai,smolagents]",
+  "any-agent[agno,google,langchain,llama_index,openai,smolagents,composio]",
 ]
 
 [project.urls]
--- a/src/any_agent/tools/composio.py
+++ b/src/any_agent/tools/composio.py
@@ -1,0 +1,98 @@
+# mypy: disable-error-code="attr-defined,operator,type-arg"
+import inspect as i
+import typing as t
+from types import FunctionType
+
+try:
+    from composio.client.types import Tool
+    from composio.core.provider import AgenticProvider
+    from composio.core.provider.agentic import AgenticProviderExecuteFn
+except ImportError as e:
+    msg = "Composio is not installed. Please install it with `pip install composio`."
+    raise ImportError(msg) from e
+
+
+TYPE_MAPPING = {
+    "string": str,
+    "integer": int,
+    "number": float,
+    "boolean": bool,
+    "array": list,
+    "object": dict,
+}
+
+
+def _get_parameters(tool: Tool) -> list[i.Parameter]:
+    parameters = []
+    if tool.input_parameters and isinstance(tool.input_parameters, dict):
+        properties = tool.input_parameters.get("properties", {})
+        required = tool.input_parameters.get("required", [])
+
+        for param_name, param_info in properties.items():
+            base_param_type = TYPE_MAPPING.get(param_info.get("type", "string"), str)
+
+            if param_name not in required:
+                param = i.Parameter(
+                    param_name,
+                    i.Parameter.KEYWORD_ONLY,
+                    default=None,
+                    annotation=base_param_type | None,
+                )
+            else:
+                param = i.Parameter(
+                    param_name,
+                    i.Parameter.KEYWORD_ONLY,
+                    annotation=base_param_type,
+                )
+            parameters.append(param)
+
+    return parameters
+
+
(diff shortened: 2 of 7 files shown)


daavoo:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/17432826346/job/49495065595

image

tests/integration/tools/test_composio.py
```diff
@@ -0,0 +1,33 @@
...
+
+def test_composio(agent_framework: AgentFramework) -> None:
+    cpo = Composio(CallableProvider())
+    tools = cpo.tools.get(
+        user_id=os.environ["COMPOSIO_USER_ID"],
+        search="repository issues",
+        toolkits=["GITHUB"],
+        limit=10,
+    )
+    agent = AnyAgent.create(
```
njbrake, line 19: Can you help me understand this: what does the .get call return, a dict or a callable? 

I liked the original design where this gives us back a json schema, because that is what lets us serialize it and save it in the any-forge state json file. 

I assume your design returns a callable since there were no edits made to the typing of AgentConfig.tools.

pyproject.toml
```diff
@@ -54,8 +54,12 @@ a2a = [
   "httpx",
 ]
 
+composio = [
+  "composio>=0.8.9,<1"
+]
```
njbrake, line 59: Did you have a look at how heavy of a dependency this is, and whether we really want to include it in the "all"? Like we don't include the A2A dep in `all` just figured I'd mention that we should consider whether we'd like this to be included.

src/any_agent/tools/composio.py
njbrake, line 1: Should this be a contribution to the Composio lib itself, not into any-agent? Maybe we merge this in to our code but at the same time open up a PR on the composio repo.

src/any_agent/tools/composio.py
```diff
@@ -0,0 +1,98 @@
...
+    "string": str,
+    "integer": int,
+    "number": float,
+    "boolean": bool,
+    "array": list,
+    "object": dict,
+}
+
+
+def _get_parameters(tool: Tool) -> list[i.Parameter]:
```
njbrake, line 25: Can we use this logic to simplify and clean up the schema -> callable conversion that we do in `tinyagent._load_agent`?

tests/integration/tools/test_composio.py
daavoo: This returns a callable. 

> I liked the original design where this gives us back a json schema, 

In any-forge we were using the default composio provider (OpenAI) and then we converting it to a callable at some point between the composio call and the agent framework call, which is kind of a misuse of their API.

The pattern that composio documents is to implement custom providers to pass the results directly downstream . The results should have the type that the framework/API expects as input (i.e. the `tools` argument in a framework).

For example:

- Google ADK: [Implementation](https://github.com/ComposioHQ/composio/blob/next/python/providers/google_adk/composio_google_adk/provider.py) - [Demo](https://github.com/ComposioHQ/composio/blob/next/python/providers/google_adk/google_adk_demo.py)
- OpenAI Agents: [Implementation](https://github.com/ComposioHQ/composio/blob/next/python/providers/openai_agents/composio_openai_agents/provider.py) - [Demo](https://github.com/ComposioHQ/composio/blob/next/python/providers/openai_agents/openai_agents_demo.py).

I added a generic callable as provider because any-agent accepts callables in `tools` (that we then wrap to each framework stuff when needed).

> because that is what lets us serialize it and save it in the any-forge state json file.

What we should serialize is the list of tool names (list of strings). 

I will send a PR updating any-forge but basically what we will do in our internal code and/or exported script is (https://docs.composio.dev/docs/fetching-tools#filtering-by-tool):

```python
cpo = Composio(CallableProvider())
AgentConfig(
    ...
    tools=cpo.tools.get(user_id=user_id, tools=SELECTED_TOOLS)
)
```

pyproject.toml
daavoo: It is not heavy but it is a good point, I will leave `all` reserved to "framework"

src/any_agent/tools/composio.py
daavoo: Yes, I will send a PR but want to merge this to not be blocked.

src/any_agent/tools/composio.py
daavoo: Not really. Here it's from JSON schema to `inspect.` objects, while in `_load_agent` it's from `inspect.` objects  to JSON Schema.

I do wonder whether it makes sense to create a small package that allows for simple bidirectional conversion `Python callable`<>`OpenAI Function Schema`. All frameworks (and composio) end up re-implementing this with some subtle inconsistencies.

daavoo:
> Also, there should be some docs to accompany this.

You are right. I will send a followup PR (I wanted to also change some things in the current Tools page)

src/any_agent/tools/composio.py
njbrake: Cool. Yeah even if we end up abandoning the schema -> callable generic approach, I think it would be useful to have that conversion logic in a single place somewhere (that was the initial motivation for https://github.com/mozilla-ai/any-forge/issues/19)

tests/integration/tools/test_composio.py
njbrake: Thanks for the clarification. Makes total sense, good with me.

njbrake (approved):
My concerns have been addressed. Needs documentation but it's ok with me if that comes in a follow up PR

PR #806, opened 2025-09-08
fix: trace should show the casted arguments

njbrake:


njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/17561512758

src/any_agent/callbacks/span_generation/tinyagent.py
```diff
@@ -4,6 +4,7 @@
 from typing import TYPE_CHECKING, Any
 
 from any_agent.callbacks.span_generation.base import _SpanGeneration
+from any_agent.utils.cast import safe_cast_argument
```
daavoo: Doesn't feel optimal having to duplicate this operation here.

Not sure I have a cleaner alternative to propose. Maybe we could extract the argument casting and we do it before/outside the `call_tool` so here in the callbacks we already get the casted args?

daavoo (approved):
Left a non-blocker comment

njbrake:
Integration tests https://github.com/mozilla-ai/any-agent/actions/runs/17588221186

PR #809, opened 2025-09-11
feat(google): Replace LiteLlm with AnyLlm as DEFAULT_MODEL_TYPE.

daavoo:


daavoo:
Need https://github.com/mozilla-ai/any-llm/pull/434 to be released

PR #817, opened 2025-09-17
Migrate `smolagents` to use `any_llm`

HareeshBahuleyan:


debug_script.py
```diff
@@ -0,0 +1,18 @@
+import dotenv
```
HareeshBahuleyan: Will remove this before the PR is ready to be merged.

src/any_agent/frameworks/smolagents.py
```diff
@@ -1,25 +1,208 @@
...
+                include={"role", "content", "tool_calls"}
+            ),
+            raw=response.model_dump(),
+            token_usage=TokenUsage(
+                input_tokens=response.usage.prompt_tokens,
+                output_tokens=response.usage.completion_tokens,
+            ),
+        )
+
+    def generate_stream(
```
HareeshBahuleyan, line 150: Keeping this to be able to contribute to smolagents sdk later on.
Tested generate_stream directly inside of smolagents and it works.

src/any_agent/frameworks/smolagents.py
```diff
@@ -137,7 +320,10 @@ async def _run_async(self, prompt: str, **kwargs: Any) -> str | BaseModel:
         if not self._agent:
             error_message = "Agent not loaded. Call load_agent() first."
             raise ValueError(error_message)
-        result = self._agent.run(prompt, **kwargs)
+        # Run the blocking `self._agent.run` call in a background thread to avoid
+        # calling synchronous code from within the current event loop.
+        result = await asyncio.to_thread(self._agent.run, prompt, **kwargs)
```
HareeshBahuleyan: If this is not the right approach, please let me know, I'd be happy to change it with an alternate approach. The integration `test_thread_safe` I think also fails due to this reason.

Currently using `result = self._agent.run(prompt, **kwargs)` results in an error.

HareeshBahuleyan: Will be fixed with https://github.com/mozilla-ai/any-llm/pull/483

src/any_agent/callbacks/span_generation/smolagents.py
```diff
@@ -46,8 +46,8 @@ def after_llm_call(self, context: Context, *args, **kwargs) -> Context:
         output_tokens = 0
         if raw := response.raw:
             if token_usage := raw.get("usage", None):
-                input_tokens = token_usage.prompt_tokens
-                output_tokens = token_usage.completion_tokens
+                input_tokens = token_usage.get("prompt_tokens", None)
+                output_tokens = token_usage.get("completion_tokens", None)
```
daavoo, line 50: Did the type of `token_usage` changed in smolagents? I didn't see any CI failures

src/any_agent/frameworks/smolagents.py
```diff
@@ -1,25 +1,207 @@
...
+        )
+
+        self._anyllm_common_kwargs: dict[str, Any] = {
+            "model": model_id,
+            "api_key": api_key,
+            "api_base": api_base,
+            **kwargs,
+        }
+
+    def create_client(self) -> Any:
```
daavoo, line 72: We should try to use the API described in https://github.com/mozilla-ai/any-llm?tab=readme-ov-file#basic-usage where we can create an instance here and reuse it on each call

debug_script.py
daavoo: Let's clean it up from the PR

daavoo (changes requested):
@HareeshBahuleyan it seems that the integration tests are failing 

https://github.com/mozilla-ai/any-agent/actions/runs/18094817370/job/51483364282

I also don't see the changes that we discussed about updating the `_prepare_completion_kwargs` so thatmessages match the format expected by any-llm (regarding tool calls)

HareeshBahuleyan:
Hey @daavoo 
Regarding
>updating the _prepare_completion_kwargs so that messages match the format expected by any-llm (regarding tool calls)

I debugged and found the issue to be in this conversion being done by `smolagents`:
https://github.com/huggingface/smolagents/blob/f76dee172666d7dad178aed06b257c629967733b/src/smolagents/models.py#L237

### Issue and initial fix
So when using the original mapping mistral-small-latest fails, as it does not complete the task:
```
tool_role_conversions = {
    MessageRole.TOOL_CALL: MessageRole.ASSISTANT,
    MessageRole.TOOL_RESPONSE: MessageRole.USER,
}
```

What works was the following conversion:
```
{
    MessageRole.TOOL_CALL: MessageRole.USER,
    MessageRole.TOOL_RESPONSE: MessageRole.USER,
}
```
Please let me know if this is okay or we should find another alternative.

### Other mappings that don't work:
(and the errors thrown shown side-by-side)

```
tool_role_conversions = {} 
```
**Error**: Input tag 'tool-response' found using 'role' does not match any of the expected tags: <Roles.system: 'system'>, <Roles.user: 'user'>, <Roles.assistant: 'assistant'>, <Roles.tool: 'tool'>

---
```
{
    MessageRole.TOOL_CALL: MessageRole.USER,
    MessageRole.TOOL_RESPONSE: 'tool',
}
```
**Error**: Unexpected role 'tool' after role 'user'

---
```
{
    MessageRole.TOOL_CALL: 'tool',
    MessageRole.TOOL_RESPONSE: MessageRole.ASSISTANT,
}
```
**Error**: Expected last role User or Tool (or Assistant with prefix True) for serving but got assistant

---
```
{
    MessageRole.TOOL_CALL: 'tool',
    MessageRole.TOOL_RESPONSE: 'tool',
} 
```
**Error**: Unexpected role 'tool' after role 'user'

---
```
{
    MessageRole.TOOL_CALL: MessageRole.ASSISTANT,
    MessageRole.TOOL_RESPONSE: 'tool',
} 
```
**Error**: Not the same number of function calls and responses

Maybe I am also not going down the right path 😅 Please let me know if you have suggestions. Thanks!

HareeshBahuleyan:
Regarding the fix that works:
```
{
    MessageRole.TOOL_CALL: MessageRole.USER,
    MessageRole.TOOL_RESPONSE: MessageRole.USER,
}
```
**Because this was an issue encountered with our default LLM used in tests `mistral-small-latest`**,  instead of having it as a default everywhere, I tried to update only our tests:
https://github.com/mozilla-ai/any-agent/blob/bf2098b12ac93b80c649dd436fdc4d5d54e985bd/src/any_agent/testing/helpers.py#L46-L55

With that the integration tests pass:
https://github.com/mozilla-ai/any-agent/actions/runs/18129214754/job/51592175760

HareeshBahuleyan:
Ran the tests for the other models as well and it seems to work.

image

The issue with `grok` seems to be caused by something else (not due to the `messages` / `role`). The error is:
```
RuntimeError: Task <Task pending name='Task-9' coro=<run_async_in_sync.<locals>.run_in_thread.<locals>.run_with_cleanup() running at /Users/hareeshbahuleyan/Desktop/mzai_github/any-agent/.venv/lib/python3.11/site-packages/any_llm/utils/aio.py:47> cb=[_run_until_complete_cb() at /Users/hareeshbahuleyan/.local/share/uv/python/cpython-3.11.11-macos-aarch64-none/lib/python3.11/asyncio/base_events.py:181]> got Future <Task pending name='Task-10' coro=<InterceptedUnaryUnaryCall._invoke() running at /Users/hareeshbahuleyan/Desktop/mzai_github/any-agent/.venv/lib/python3.11/site-packages/grpc/aio/_interceptor.py:655> cb=[InterceptedCall._fire_or_add_pending_done_callbacks()]> attached to a different loop 
```

A similar issue is encountered when running debug_script.py with the error `Event loop is closed`

@daavoo I think it has to do with the `allow_running_loop` param from https://github.com/mozilla-ai/any-llm/pull/483 but I am not sure how to fix it. I was wondering if you have suggestions.

debug_script.py
HareeshBahuleyan: Done

src/any_agent/callbacks/span_generation/smolagents.py
HareeshBahuleyan: Was getting error `'dict' object has no attribute 'prompt_tokens'` hence added the change

src/any_agent/frameworks/smolagents.py
HareeshBahuleyan: Created and issue to follow up on this: https://github.com/mozilla-ai/any-agent/issues/824

HareeshBahuleyan:
@daavoo 
Integration tests passing: https://github.com/mozilla-ai/any-agent/actions/runs/18196938830/job/51806475077

So the 2 issues that remain are:
1. `xai` integration test failing with the error described in the [above comment](https://github.com/mozilla-ai/any-agent/pull/817#issuecomment-3357046588)

2. [Only with `openai` models] When running a sample debug script, the agent completes the task and provides the output, but finally throws the `Event loop is closed` error (but this does not happen with integration tests!). Can be reproduced by running:

```
import dotenv

from any_agent.tools import search_tavily

dotenv.load_dotenv()

agent = AnyAgent.create(
    "smolagents",
    AgentConfig(
        model_id="openai/gpt-4.1",
        instructions="You are a helpful assistant. Answer the question with available tools.",
        tools=[search_tavily],
        model_args={"tool_choice": "auto", "allow_running_loop": True},
    ),
)

trace = agent.run("Which agent framework is the best?")
```

HareeshBahuleyan:
Thanks @daavoo 

Update regarding the 2 issues mentioned above:

> 1. xai integration test failing

^An issue has been created to track this: https://github.com/mozilla-ai/any-agent/issues/827 


> 2. [Only with openai models] When running a sample debug script, the agent completes the task and provides the output, but finally throws the Event loop is closed error 

^This was apparently some issue with my env / cache. I tried the debug script from a fresh terminal and did not encounter it anymore.

PR #828, opened 2025-10-09
Migrate `openai` framework to use `any_llm` 

HareeshBahuleyan:
- Introduce `AnyllmModel` based on `LitellmModel` from openai agents SDK: https://github.com/openai/openai-agents-python/blob/main/src/agents/extensions/models/litellm_model.py#L64
- Updated tests to use appropriate mock depending on the llm library (litellm /  any_llm) used by the respective frameworks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
         exclude: \.ambr$
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.14.0"
+    rev: "v0.14.1"
     hooks:
       - id: ruff
         args:


HareeshBahuleyan:
Integration tests passing on CI: https://github.com/mozilla-ai/any-agent/actions/runs/18492592870

Status of running [OPENAI] framework integration tests locally:

- ✅ Mistral
- ✅ OpenAI
- ❌ Gemini (also fails when tested on `main`)
    - Reason: Gemini has no support for OpenAI json_schema for structured output; they use a different format: https://googleapis.github.io/python-genai/index.html#json-response-schema
- ❌ XAI (also fails when tested on `main`)
    - Reason: XAI SDK does not support OpenAI structured output format: https://docs.x.ai/docs/guides/structured-outputs
- ❌ HF Inference Endpoint
    - Reason: The `completion` fails with ClientResponseError / Unprocessable Entity when the request payload contains **both** `tools` and `response_format`!
        - However, if either one of them are popped out from the dict, it works. Could it be an [[model-specific issue](https://www.reddit.com/r/LocalLLaMA/comments/1h2y7ys/why_can_you_not_use_structured_output_and_tool/)](https://www.reddit.com/r/LocalLLaMA/comments/1h2y7ys/why_can_you_not_use_structured_output_and_tool/) with the hosted Qwen-3 model?
    - When tested with tinyagent (instead of OpenAI), the payload prepared by tinyagent contains either *only* `tool` or *only* `response_format` but not both.

@daavoo I guess Gemini and XAI not supporting OpenAI structured output format is something that we cannot fix. 
Hence I think we based on the code review, we could proceed with the PR.

src/any_agent/frameworks/openai.py
```diff
@@ -1,30 +1,379 @@
...
+                tools,
+                output_schema,
+                handoffs,
+                span_generation,
+                tracing,
+                stream=False,
+                prompt=prompt,
+            )
+
+            assert isinstance(response.choices[0], any_llm.types.completion.Choice)
```
daavoo, line 123: Why this assertion here?

src/any_agent/frameworks/openai.py
```diff
@@ -1,30 +1,379 @@
...
+    from agents.items import TResponseInputItem, TResponseStreamEvent
+    from agents.models.interface import ModelTracing
+    from agents.tracing.span_data import GenerationSpanData
+    from agents.tracing.spans import Span
+    from openai.types.chat import ChatCompletionToolParam
+    from pydantic import BaseModel
+
+
+class Converter(BaseConverter):
+    """Same converter as agents.models.chatcmpl_converter.Converter but with strict mode enabled."""
```
daavoo, line 49: I don't understand why we need this strict mode enabled, could you clarify?

daavoo:
> @daavoo I guess Gemini and XAI not supporting OpenAI structured output format is something that we cannot fix.

@HareeshBahuleyan I don't think we can fix it in `any-agent` but I don't see why we couldn't make this work with changes in `any-llm`

src/any_agent/frameworks/openai.py
HareeshBahuleyan: Without strict mode, when running the OPENAI test_agent integration test, I get:
```
`write_file` is not strict. Only `strict` function tools can be auto-parsed
```

src/any_agent/frameworks/openai.py
HareeshBahuleyan: Probably it is not needed, I was using the same steps as in the original LitellmModel class:
https://github.com/openai/openai-agents-python/blob/main/src/agents/extensions/models/litellm_model.py#L112

HareeshBahuleyan:
> but I don't see why we couldn't make this work with changes in any-llm

@daavoo Do you mean implementing a helper function in any-llm to convert openAI JSON schema format to Pydantic?

daavoo:
> > but I don't see why we couldn't make this work with changes in any-llm
> 
> @daavoo Do you mean implementing a helper function in any-llm to convert openAI JSON schema format to Pydantic?

no, I mean that any-llm should support receiving as `response_format` any of the argument types that openai supports and take care of converting them to the appropriate format for each provider under the hood (in this case there should be some conversion for gemini and xai)

HareeshBahuleyan:
Sounds reasonable, created issues in any-llm to track this: https://github.com/mozilla-ai/any-llm/issues/541 and https://github.com/mozilla-ai/any-llm/issues/542

src/any_agent/frameworks/openai.py
```diff
@@ -1,30 +1,379 @@
...
-        Runner,
+    import any_llm
+    from agents import Agent, FunctionTool, Model, ModelSettings, Runner, Tool, Usage
+    from agents.exceptions import UserError
+    from agents.items import ModelResponse
+    from agents.models.chatcmpl_converter import Converter as BaseConverter
+    from agents.models.chatcmpl_stream_handler import ChatCmplStreamHandler
+    from agents.models.fake_id import FAKE_RESPONSES_ID
+    from agents.tracing import generation_span
+    from agents.util._json import _to_dump_compatible
```
daavoo, line 21: This requires bumping `"openai-agents>=0.3.0, <1"` in pyproject.toml

src/any_agent/frameworks/openai.py
```diff
@@ -1,30 +1,379 @@
...
+        if model_settings.metadata:
+            extra_kwargs["metadata"] = copy(model_settings.metadata)
+        if model_settings.extra_body and isinstance(model_settings.extra_body, dict):
+            extra_kwargs.update(model_settings.extra_body)
+
+        # Add kwargs from model_settings.extra_args, filtering out None values
+        if model_settings.extra_args:
+            extra_kwargs.update(model_settings.extra_args)
+
+        ret = await any_llm.acompletion(
```
daavoo: Can we try using the class-based API?
Since this is async code, I don't think we will find the issues found with smolagents

daavoo (changes requested):
I run the tests with the debugger and implementation looks good.

We weed to update `pyproject.toml` and would be great to try the class-based AnyLLM API.

src/any_agent/frameworks/openai.py
HareeshBahuleyan: Yeah that works without issues, the change is in this commit https://github.com/mozilla-ai/any-agent/pull/828/commits/c06f1cdbeb68bcb9cb9e6fbe894240436d2c9c5c

HareeshBahuleyan:
Rerun of integration tests after class-based AnyLLM definition:
https://github.com/mozilla-ai/any-agent/actions/runs/18588782499

PR #840, opened 2025-10-20
feat(langchain): switch from litellm to any_llm

njbrake:
This PR removes the litellm dependency from langchain (which was a vendored langchain_litellm package) and uses any_llm. 

I went with putting the ChatAnyLLM model into the same file as the agent implementation, but I'm not against putting in a separate file if someone has opinions.

### DIFF
--- a/src/any_agent/frameworks/langchain.py
+++ b/src/any_agent/frameworks/langchain.py
@@ -80,7 +80,8 @@ class ChatAnyLLM(BaseChatModel):
             )
             return generate_from_stream(stream_iter)
 
-        message_dicts, params = self._create_message_dicts(messages, stop)
+        message_dicts = [_convert_message_to_dict(m) for m in messages]
+        params = self._create_params(stop)
         params = {**params, **kwargs}
         response = completion(messages=message_dicts, **params)  # type: ignore[arg-type]
         if not isinstance(response, ChatCompletion):
@@ -115,9 +116,7 @@ class ChatAnyLLM(BaseChatModel):
         }
         return ChatResult(generations=generations, llm_output=llm_output)
 
-    def _create_message_dicts(
-        self, messages: list[BaseMessage], stop: list[str] | None = None
-    ) -> tuple[list[dict[str, Any]], dict[str, Any]]:
+    def _create_params(self, stop: list[str] | None = None) -> dict[str, Any]:
         params = {
             "api_key": self.api_key,
             "api_base": self.api_base,
@@ -129,8 +128,7 @@ class ChatAnyLLM(BaseChatModel):
                 error_message = "`stop` found in both the input and default params."
                 raise ValueError(error_message)
             params["stop"] = stop
-        message_dicts = [_convert_message_to_dict(m) for m in messages]
-        return message_dicts, params
+        return params
 
     def _stream(
         self,
@@ -139,7 +137,8 @@ class ChatAnyLLM(BaseChatModel):
         run_manager: CallbackManagerForLLMRun | None = None,
         **kwargs: Any,
     ) -> Iterator[ChatGenerationChunk]:
-        message_dicts, params = self._create_message_dicts(messages, stop)
+        message_dicts = [_convert_message_to_dict(m) for m in messages]
+        params = self._create_params(stop)
         params = {**params, **kwargs, "stream": True}
 
         default_chunk_class: type[BaseMessageChunk] = AIMessageChunk
@@ -168,7 +167,8 @@ class ChatAnyLLM(BaseChatModel):
         run_manager: AsyncCallbackManagerForLLMRun | None = None,
         **kwargs: Any,
     ) -> AsyncIterator[ChatGenerationChunk]:
-        message_dicts, params = self._create_message_dicts(messages, stop)
+        message_dicts = [_convert_message_to_dict(m) for m in messages]
+        params = self._create_params(stop)
         params = {**params, **kwargs, "stream": True}
 
         default_chunk_class: type[BaseMessageChunk] = AIMessageChunk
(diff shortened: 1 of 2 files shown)


njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/18659442435

src/any_agent/frameworks/langchain.py
```diff
@@ -1,29 +1,373 @@
...
+
+    def _stream(
+        self,
+        messages: list[BaseMessage],
+        stop: list[str] | None = None,
+        run_manager: CallbackManagerForLLMRun | None = None,
+        **kwargs: Any,
+    ) -> Iterator[ChatGenerationChunk]:
+        message_dicts, params = self._create_message_dicts(messages)
+        params = {**params, **kwargs, "stream": True}
```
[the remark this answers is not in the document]
njbrake: Resolved

src/any_agent/frameworks/langchain.py
```diff
@@ -110,18 +454,20 @@ async def _run_async(self, prompt: str, **kwargs: Any) -> str | BaseModel:
                 structured_output_message,
             ]
 
-            # Use response schema if supported by the model
-            if supports_response_schema(model=self.config.model_id):
-                completion_params["response_format"] = self.config.output_type
```
njbrake, line 115: WIth this behavior change, we would now enforce that if you are asking for a response schema, the provider/model has to support returning a response schema.

njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/18669080798

src/any_agent/frameworks/langchain.py
```diff
@@ -1,29 +1,246 @@
...
+        llm_output = {
+            "token_usage": token_usage,
+            "model": self.model,
+        }
+        return ChatResult(generations=generations, llm_output=llm_output)
+
+    def _create_message_dicts(
+        self, messages: list[BaseMessage], stop: list[str] | None = None
+    ) -> tuple[list[dict[str, Any]], dict[str, Any]]:
+        params = {
```
daavoo, line 120: nit: I am a little confused why `params` is handled inside this `_create_message_dicts`, seem 2 completely different things.

src/any_agent/frameworks/langchain.py
```diff
@@ -47,7 +264,7 @@ def _get_model(self, agent_config: AgentConfig) -> "LanguageModelLike":
                 model=agent_config.model_id,
                 api_key=agent_config.api_key,
                 api_base=agent_config.api_base,
```
HareeshBahuleyan, line 267: Would this work when running the integration test with `huggingface/tgi`
I noticed that it sets the api_base in the model args:
https://github.com/mozilla-ai/any-agent/blob/3dad3f6c0e4b69b1878d61408306db384734f659/tests/integration/frameworks/test_agent.py#L159
and agent_config.api_base might be None. 
If thats the case, we might need something like:
```
agent_config.api_base or cast(
            "str | None", model_args.get("api_base")
        )
```

src/any_agent/frameworks/langchain.py
```diff
@@ -1,29 +1,246 @@
...
+        if stream:
+            stream_iter = self._stream(
+                messages, stop=stop, run_manager=run_manager, **kwargs
+            )
+            return generate_from_stream(stream_iter)
+
+        message_dicts, params = self._create_message_dicts(messages, stop)
+        params = {**params, **kwargs}
+        response = completion(messages=message_dicts, **params)  # type: ignore[arg-type]
+        if not isinstance(response, ChatCompletion):
```
daavoo, line 87: How could this Exception happen if we are just calling any_llm?

src/any_agent/frameworks/langchain.py
```diff
@@ -1,29 +1,246 @@
...
+    ) -> ChatResult:
+        if stream:
+            stream_iter = self._stream(
+                messages, stop=stop, run_manager=run_manager, **kwargs
+            )
+            return generate_from_stream(stream_iter)
+
+        message_dicts, params = self._create_message_dicts(messages, stop)
+        params = {**params, **kwargs}
+        response = completion(messages=message_dicts, **params)  # type: ignore[arg-type]
```
HareeshBahuleyan, line 86: Should we use `AnyLLM.create()` and reuse the same instance across completion calls?

daavoo (approved):
Went through the code with the debugger from the integration test.

src/any_agent/frameworks/langchain.py
njbrake: It's a typing quirk because technically completion returns either ChatCompletion or an iterator of chunks. This if statement makes the type checker happy by confirming that it's not returning an iterator

src/any_agent/frameworks/langchain.py
njbrake: Good question and good find! I think you're on to something here: we've got some quirks with agent_config.api_base vs model_args.api_base (and the other api client related settings) I went ahead and created a new github issue to track this: it's an existing issue/question and not related to this change set I don't think.

src/any_agent/frameworks/langchain.py
njbrake: Ooo good question. I tried this, but langchain enforces that the class has to be serializable, which the anyllm client is not. 😢 . If I add the client as an attribute of the class I get

```
                if hasattr(self.__class__, item):
>                   return super().__getattribute__(item)  # Raises AttributeError if appropriate
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E                   AttributeError: 'any_agent.frameworks.langchain.ChatAnyLLM' object has no attribute '__pydantic_fields_set__'. Did you mean: '__pydantic_fields__'?

.venv/lib/python3.13/site-packages/pydantic/main.py:1023: AttributeError
```

src/any_agent/frameworks/langchain.py
njbrake: Yeah you're right. It was a piece of the logical flow copied over from the langchain_litellm package. I'll separate them to make it a lil cleaner

njbrake:
Integration tests after PR review updates: https://github.com/mozilla-ai/any-agent/actions/runs/18679175077

src/any_agent/frameworks/langchain.py
daavoo: Could we do `cast(ChatCompletion...` instead?

PR #841, opened 2025-10-20
feat(llama_index): 769 migrate llama index to use any llm

njbrake:
Cursor/Claude helped me out a bunch with the initial migration. I went through and cleaned it up a bit and tried my best to make it clear which parts of the code are "vendored" utils code from the llama_index litellm implementation (pieces that don't use litellm at all but are used to convert llama_index objects to and from the openai spec).

Integration tests and unit tests are passing so we should have some baseline confidence that this is good enough to merge.


njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/18668633524

src/any_agent/frameworks/llama_index.py
```diff
@@ -1,35 +1,501 @@
...
+        llm_completion_callback,
+    )
+    from llama_index.core.llms.function_calling import FunctionCallingLLM
+    from llama_index.core.tools import ToolSelection
+    from llama_index.core.tools.types import BaseTool
 
     DEFAULT_AGENT_TYPE = FunctionAgent
-    DEFAULT_MODEL_TYPE = LiteLLM
+
+    class AnyLLMWrapper(FunctionCallingLLM):
```
daavoo, line 67: Why is the class definition inside the `try:` block?

src/any_agent/vendor/llama_index_utils.py
```diff
@@ -0,0 +1,223 @@
...
+# to have litellm installed (it's imported at the top of the file), so we copy the code here to avoid the dependency.
+from collections.abc import Sequence
+from typing import Any, cast
+
+from llama_index.core.base.llms.types import ChatMessage, ChatResponse
+
+
+def to_openailike_message_dict(message: ChatMessage) -> dict[str, Any]:
+    """Convert a ChatMessage to an OpenAI-like message dict."""
+    from llama_index.core.base.llms.types import (
```
daavoo: Why are these imports here if we are already importing from the same module outside?

src/any_agent/frameworks/llama_index.py
```diff
@@ -1,35 +1,501 @@
...
+
+        def _chat(self, messages: Sequence[ChatMessage], **kwargs: Any) -> ChatResponse:
+            if not self._client:
+                msg = "Client not initialized."
+                raise ValueError(msg)
+
+            message_dicts = to_openai_message_dicts(messages)
+            all_kwargs = self._get_all_kwargs(**kwargs)
+            if "max_tokens" in all_kwargs and all_kwargs["max_tokens"] is None:
+                all_kwargs.pop("max_tokens")
```
daavoo: Not sure I understand why this is needed

daavoo (approved):
Went through the code with the debugger from the integration test and everything kind of make sense (leaving aside all the framework-specific quirks).

They seem to use astream_chat so not sure if/why all the other APIs are needed

njbrake:
> They seem to use astream_chat so not sure if/why all the other APIs are needed

Kept the other APIs to make contributions back to the llama_index repo easier down the road, we'll see if that turns out to save or cost us time 😅

src/any_agent/frameworks/llama_index.py
njbrake: Since the FunctionCallingLLM is a llama_index class, the class definition depends on the llama_index package existing, so if this was outside the try/catch it would throw name errors and would bypass our llama_index_available logic. Which is now making me think that we may need to go back and touch this up for the other providers, not sure we did this consistently in the conversion process.

src/any_agent/frameworks/llama_index.py
njbrake: Thank you, good catch. Carry over from litellm wrapper logic

njbrake:
Integration tests after PR updates https://github.com/mozilla-ai/any-agent/actions/runs/18679692206

PR #847, opened 2025-10-21
feat: removal of litellm requirement

njbrake:
Now that all frameworks have transitioned to any-llm, we can remove litellm dependence.


njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/18683508383

PR #852, opened 2025-10-21
feat(callbacks): Callbacks can edit messages

njbrake:
Feature: before_llm_call can edit the messages.

The pros:
* This feature works on all frameworks, I feel good about the integration test I added sufficiently exercising the code so that we will know if any underlying logic changes for the situation we are testing, the test will fail and we'll know one of our assumptions is broken

The cons:
* Because we have to inject ourselves into specific parts of the agent loop, we're not pretty tightly tied to the agent being used inside the framework. If the user decides to use a different agent model from the default_agent_model in each framework, our callbacks will probably not work for them.

I don't see a good solution for the con here: callbacks to modify message behavior wildly varies from framework to framework and is heavily implementation specific. I lean towards merging this PR and fielding any bug reports if they come in, but I want to be clear that I'm not confident that this is going to be a robust long term solution, but I also don't see any other reasonable way to do this across all the agent frameworks.

### DIFF
--- a/src/any_agent/callbacks/span_generation/agno.py
+++ b/src/any_agent/callbacks/span_generation/agno.py
@@ -39,7 +39,9 @@ class _AgnoSpanGeneration(_SpanGeneration):
                             output = [
                                 {
                                     "tool.name": getattr(
-                                        getattr(tool, "function", None), "name", "No name"
+                                        getattr(tool, "function", None),
+                                        "name",
+                                        "No name",
                                     ),
                                     "tool.args": getattr(
                                         getattr(tool, "function", None),
@@ -63,7 +65,9 @@ class _AgnoSpanGeneration(_SpanGeneration):
                 if tool_calls := getattr(assistant_message, "tool_calls", None):
                     output = [
                         {
-                            "tool.name": tool.get("function", {}).get("name", "No name"),
+                            "tool.name": tool.get("function", {}).get(
+                                "name", "No name"
+                            ),
                             "tool.args": tool.get("function", {}).get(
                                 "arguments", "No args"
                             ),
--- a/src/any_agent/callbacks/wrappers/langchain.py
+++ b/src/any_agent/callbacks/wrappers/langchain.py
@@ -188,14 +188,11 @@ class _LangChainWrapper:
 
                 result = await self._original_agenerate(messages_list, *args, **kwargs)
 
-                try:
-                    context = self.callback_context[
-                        get_current_span().get_span_context().trace_id
-                    ]
-                    context.framework_state._message_getter = None
-                    context.framework_state._message_setter = None
-                except Exception:
-                    pass
+                context = self.callback_context[
+                    get_current_span().get_span_context().trace_id
+                ]
+                context.framework_state._message_getter = None
+                context.framework_state._message_setter = None
 
                 return result
 


daavoo:
> @daavoo before I get too deep into this PR, would you be able to give a quick look to make sure it aligns with what you were thinking too? I'm 90% sure it does, but want to check in before I go through the trouble of implementing it for all the frameworks

You're cooking, looks good to me

njbrake:
Smolagents implementation pending improvement based on any responses from the HF team on https://github.com/huggingface/smolagents/issues/1834

src/any_agent/frameworks/smolagents.py
```diff
@@ -66,6 +66,7 @@ def __init__(
             "model": model_id,
             "api_key": api_key,
             "api_base": api_base,
+            "allow_running_loop": True,  # Because smolagents uses sync api
```
njbrake, line 69: This was a bug in our code that was never found because we weren't testing callbacks in our integration tests

src/any_agent/callbacks/wrappers/smolagents.py
```diff
@@ -20,30 +20,102 @@ def __init__(self) -> None:
...
+                            break
+
+            context.framework_state._message_getter = get_messages
+            context.framework_state._message_setter = set_messages
+
+            # Only invoke callbacks on the first LLM call, not on retry attempts
+            # Retries can be detected by checking if there are error messages in the history
+            is_retry = any(
+                "Error:" in str(msg.content) and "Now let's retry" in str(msg.content)
+                for msg in messages
```
njbrake, line 95: The smolagents implementation is certainly quirky. They append error messages as tool responses and have retry hooks a little differently from the other libraries. Putting this here which Claude helped me generate (with a bunch of guidance):

The Complete Flow with Message Modification in Smolagents librar

  First Call (Step 1):

  1. _run_stream() line 576: calls _step_stream()
  2. _step_stream() line 1259: calls write_memory_to_messages()
  3. write_memory_to_messages() line 764: calls TaskStep.to_messages()
  4. TaskStep.to_messages() line 192: returns "New task:\nSay goodbye"
  5. Wrapper modifies: Changes last message to "Say hello and goodbye" ✓
  6. _step_stream() line 1284: calls model.generate() with modified messages
  7. LLM tries to call final_answer twice → raises AgentExecutionError
  8. _run_stream() line 597: stores error in action_step.error
  9. _run_stream() line 600: appends action_step (with error) to memory

  Second Call (Step 2 - Retry):

  1. _run_stream() line 576: calls _step_stream() AGAIN
  2. _step_stream() line 1259: calls write_memory_to_messages() AGAIN
  3. write_memory_to_messages() rebuilds from memory:
    - TaskStep.to_messages() → "New task:\nSay goodbye" (original task!)
    - ActionStep.to_messages() → Error message: "Error:\n...Now let's retry..."
  4. Messages are rebuilt from memory → modification is LOST ✗
  5. Wrapper modifies the last message (which is now the error message, not the task!)
  6. LLM receives original unmodified task + error message

  ---
  Why This Happens

  The design assumes messages are stateless and reproducible from memory structures. Smolagents never expects the messages parameter to be modified at runtime. The architecture is:

  Memory Structures (TaskStep, ActionStep) → to_messages() → Fresh Message List

  This happens before every LLM call, so any modifications to the message list itself are discarded.

njbrake: because of this, we don't want to redo the callbacks each retry, we want to only make that callback interference on the entry, not the retries (since that would be messing with the internal smolagents logic).

njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/18853291870

njbrake:
Integration tests: https://github.com/mozilla-ai/any-agent/actions/runs/19055632789

PR #859, opened 2025-10-28
Adding pre-commit check to remove jupyter output

angpt:
Using `nbstripout` with pre-commit to remove local output from Jupyter cookbooks. Also cleaned up existing cookbooks.


njbrake:
> Thoughts @njbrake ?

I have no strong opinions here. I think in some cases the jupyter output is useful for those that don't want to run the notebook, but then if they don't want to run the notebook, maybe they should be looking at a docs page instead of a cookbook, idk. I'm comfortable with whatever direction @angpt thinks is best 🚀

angpt:
Thanks for bringing this up @daavoo. Here's why I think the pre-commit approach + explicit output sections works better:

- Security: Local output can leak sensitive info (paths, keys in errors, environment details) that's hard to catch in review
- Maintenance: Output creates noisy diffs and can become outdated
- Intentional examples: Explicit "Expected Output" sections (like in the [any-llm cookbook](https://github.com/mozilla-ai/any-llm/blob/main/docs/cookbooks/any_llm_getting_started.ipynb)) let us show exactly what matters without the risks

This way users can still see what to expect, but we control what's shared.

daavoo:
> Thanks for bringing this up @daavoo. Here's why I think the pre-commit approach + explicit output sections works better:
> 
>     * Security: Local output can leak sensitive info (paths, keys in errors, environment details) that's hard to catch in review
> 
>     * Maintenance: Output creates noisy diffs and can become outdated
> 
>     * Intentional examples: Explicit "Expected Output" sections (like in the [any-llm cookbook](https://github.com/mozilla-ai/any-llm/blob/main/docs/cookbooks/any_llm_getting_started.ipynb)) let us show exactly what matters without the risks
> 
> 
> This way users can still see what to expect, but we control what's shared.

Makes sense to me @angpt 🙏 

Not a blocker for this, but at least in followup would be nice to then add explicit outputs to these cookbooks (not sure if there is an easy way to get the trace output to render nice in markdown cell🤔 )

